PR #7, opened 2017-10-18
Fixes and restructuring

therationalpi:
I made some fixes to a few bugs and did some major restructuring by moving monika's topics into their own file. Picking a random topic is more normalized now, in that you add topics to a persistent list and no longer have to follow any sequential numbering scheme. I also fixed the dialogue box to be more friendly with natural language (it checks each word and word pair against the keyword list).

I also changed away from the keylistener system over to renpy's built-in keybind system attached to the underlay. This does add a new bug that the dialogue box can be opened straight from the menu, but it's better than the previous bugs.


Backdash:
Merged, with changes from me added. Do you know if it's possible to sync up the background video with the lens flare effect? It gets distracting when that sometimes happens for no reason.

therationalpi:
I think I can do something for that. The background video could really use some cleaning up anyway. Do you happen to know where in the code the lens flare effect is defined? I think that will help me with syncing things.

Also, I'd like to chat a bit about some stuff on this. Can you shoot me a message on Discord at Pi#0829?

PR #9, opened 2017-10-22
Added time module and custom time-based greetings.

onfaut:
i might have screwed something up but this seems to be it for the most part
edit: if the time range seems too broad feel free to edit


therationalpi:
Datetime was imported in ch30, so I made this script use that implementation.

PR #11, opened 2017-10-22
Update script-topics.rpy

tuebor1837:
Two new topics, a quick little humorous exchange and either the best or the worst thing I have written in my entire life


therationalpi:
Trying to make sense of the whole monika_bane thing, what were you going for with that?

It's written like a scene for multiple characters, but right now it's just going to be in the room with Monika with the name changing...

Even if it was formatted right for all the characters, I don't think some Banepost copypasta is a fit for this mod.

tuebor1837:
Eh, it was worth a shot.

Issue #25, opened 2017-10-27
[Suggestion] - Monika talks about your steam games


BITDEM0NICS:
In act 3, where it's just the player and monika, Monika could possibly read the player's game library which DDLC is installed into, and talk about some games that the player might have. For example, if the player has Dark souls or Cuphead installed, she will make a remark about the difficult, or if the player has Five nights at Freddy's or undertale, she could remark about their popularity

therationalpi:
I like this idea. Having Monika check files on the player's computer would be fun, as long as we don't overstep our boundaries and get too intrusive.

We have a few topics in scipt-topics that reference gaming...maybe we could add this into one of those?

polishsnake:
This could be good, however not everyone has disks with big capacity; so a player might have a lot of games bought/played but not many installed. My suggestion is to try to find a way to find player's profile on Steam (if that's possible) and comment about popular tags; this way Monika would mention on even the uninstalled games, as well as people who are into more unknown games can still get a comment on that. Also a comment on the amount of games would be interesting (the again, if it's possible to check the Steam profile via usage of Python :| ).

therationalpi:
I think that way of doing it is more trouble than it's worth. We want to create the illusion that Monika is watching you play games, not turn Monika into a walking encyclopedia of your gaming habits.

If Monika doesn't find any games she knows in your Steam folder, she just won't comment on it and the player will be none the wiser.

therationalpi:
@ohxeighty Has added a new function enumerate_steam() that can be used to check if a certain game is in your steam library!

MikuChanAnime:
I like this idea, she can detect specific games that what pi said, but for other games, she can test it for certain tags the game has. Like nudity, horror, sandbox, etc, etc.

PR #29, opened 2017-10-28
Add steam enumeration function

ohxeighty:
Add steam enumeration function to return installed app IDs. 
Reference against https://steamdb.info/apps/ for dialogue involving installed games. 
Addresses suggestion #25, https://github.com/Backdash/MonikaModDev/issues/25

### DIFF
--- a/Monika After Story/game/script-ch30.rpy
+++ b/Monika After Story/game/script-ch30.rpy
@@ -80,6 +80,7 @@ init python:
     import eliza      # mod specific
     import datetime   # mod specific
     import re
+    import _winreg    # mod specific 
     therapist = eliza.eliza()
     process_list = []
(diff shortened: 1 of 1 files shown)


therationalpi:
_winreg is platform specific, and we're currently targeting Windows, Mac, and Linux. I'm fine with having this feature be Windows only for the time being, but can you add some checks so we don't crash on other systems?

Admicos:
Under Linux, the steamapps directory should be under `$HOME/.steam/steam/steamapps`. You can get `$HOME` using `os.environ.get("HOME")`.

No idea for Macs though.

therationalpi:
Steamapps directory on mac is $HOME/Library/Application\ Support/Steam/Steamapps

PR #32, opened 2017-10-28
Update pong.rpy

DoomTheRobot:
Simple fix that displays the player's name instead of "(You)".

Also, should I adjust the sound effects to be more quiet?


therationalpi:
I think that would be a good change

PR #44, opened 2017-10-31
Added politics topic + additional swear words.

Jxhnathan08:
Just a small edit with a short answer on politics & adding some more swear words to monika_vulgarity.


Jxhnathan08:
Closing because I'm doing some more work.

PR #45, opened 2017-10-31
Added template disclaimer to splash.

Jxhnathan08:
Don't have time to do the asset check, but the disclaimer is out of the way.


Jxhnathan08:
Closing due to @therationalpi having it taken care of through the new mod template.

PR #46, opened 2017-10-31
Prototypes

monikLover:
Contains prototype projects for future use.


therationalpi:
This is all useful stuff. We'll integrate it down the line, but as a separate pull request in the future.

PR #52, opened 2017-10-31
Added topics & cut down on idea list.

Jxhnathan08:
Some of these topics may need writing changes. Cut out some ideas that were either implemented or existed already.


therationalpi:
Looks good to me.

PR #57, opened 2017-11-02
script-ch30.rpy: Added.

Jxhnathan08:
Added script-ch30 for simplicity & removed DDLCTemplate scripts. @therationalpi If you could check this over, I'd appreciate it.


therationalpi:
No arguments. script-ch30 is the basis for MAS and we won't need those placeholder scripts anymore.

Issue #75, opened 2017-11-04
Fix save loading and persistent updates in lightweight


therationalpi:
For some reason, switching over to the mod template broke the persistent updating and save loading from version 0.4.0 got busted. Please uncomment this line in splash.rpy:

`       
 #call import_ddlc_persistent
`

And try to fix things with changes to import_ddlc, updates, and updates_topics. No changes to splash.rpy should be necessary.

By keeping this isolated, we open up the possibility of adding an import function in the settings menu that just calls this.

ThePotatoGuy:
Okay, I've figured this out. Theres actually two issues here.

The first issue is the crash. This occurs because updates.topics originally was defined as a dict, but I had changed that to None as a memory optimization. In import_ddlc, updateTopicIDs was being called before updates.topics would be init, so instead of returning old_persistent (as updates.topics was an empty dict), it was trying to perform dict operations on a Nonetype.

Which leads to the 2nd issue. Turns out, those calls to updateTopicIDs could be pointless in some cases because updates.topics only was initialized in a version number change.

I've added a fix that moves updates.topics init into a label, and adds appropriate calls to this label before updateTopicIDs is called in import_ddlc.

therationalpi:
Seems to be totally fixed as of this commit. 

https://github.com/Backdash/MonikaModDev/commit/2ecd7e2ff2eb1d41d039e0f051a1014f3061fb1a

PR #78, opened 2017-11-04
Fixed save loading and persistent updates in lightweight

ThePotatoGuy:
#75 

This creates a new label for init of updates.topics and updates.version_updates. 

The label is called before calling updateTopicIDs in import_ddlc to ensure initalization of these values in the case that a version number hasn't been changed.


Jxhnathan08:
Every pull request brings it closer to release.

Beautiful.

PR #79, opened 2017-11-04
script-topics.rpy: Added Wikipedia stub.

Jxhnathan08:
Suggested by @Vami-IV.


Vami-IV:
I think that the line "..." should be removed now that I look at this. Looks good so far, will definitely have more to say and suggest when I'm through with a different Wikipedia matter.

Issue #123, opened 2017-11-09
[Bug] Monika repeats topics


ThePotatoGuy:
I have fixed this and will explain in a pull req.

PR #124, opened 2017-11-09
Fixed repeated topics issue

ThePotatoGuy:
#123 

This occurs in two possible locations:

1. When the player enters a keyword via dialogue box (T), the potential topics are never checked if they had been seen.
2. After a player selects a topic via dialogue box (T), the call to the label is performed via renpy.call. Since each topic has a return, this skips the code post renpy.call, preventing the topic from being removed from monika_random_topics.

Fixes:
1. renpy.seen_label is used prior to adding a topic to potential topics list
2. renpy.call replaced with renpy.call_in_new_context


Jxhnathan08:
Looked over & everything checks out. Merged.

PR #139, opened 2017-11-12
Label "Name"

Rune0n:
Conversation focused on the origin of her name, with an eventual shift to the player's.
Added a comment to an incomplete topic.


Winblows2:
The writing will be reviewed by @monikLover before merging to make sure everything checks out. Other than that, I'd say this is a solid contribution.

therationalpi:
I really like this, very good stuff, but it needs some keywords and I think it can be adjusted into a random topic as well. I'll make those edits and push it to the content branch.

PR #146, opened 2017-11-13
Fixed conflict with music menu and dialogue box

ThePotatoGuy:
Quick fix for stopping dialogue from interrupting music menu and breaking everything.

Also fixed a syntax error.


Winblows2:
Since you test in-game, a quick once-over of this suffices. Merged.

PR #155, opened 2017-11-13
script-topics: Possible replacement for weak adjectives.

inkedsaber:


Winblows2:
This will be reviewed by @monikLover before being merged. For future reference, this should be kept in mind:

> Monika speaks very casually when talking to the player, so try to not get too fancy.

inkedsaber:
> Monika speaks very casually when talking to the player, so try to not get too fancy.
Thanks for the tip, I'll keep that in mind.
Maybe 'genuine' was too much 🤔 
However, it's better to use 'tiny' instead of 'very small' IMHO

PR #156, opened 2017-11-13
Punctuation Changes?

Rune0n:
Changed a set of double quotations to fall in line with the single quotations being used for quotes within quotes.
Added quotes when referring to names in label_name.


Winblows2:
Approved.

PR #179, opened 2017-11-18
Add chess as a minigame.

kadava:
Added chess as a minigame. It can be launched by clicking the 'c' key on the main screen. Supported platforms include 64- and 32-bit Windows, 64-bit Linux and 64-bit MacOSX. In case MAS is run on a platform that doesn't support chess, the minigame won't be mentioned to the user and the 'c' key will be inactive. I have tested the minigame on both 64-bit Linux and 32-bit Windows and encountered no issues.

The minigame uses [Stockfish](https://stockfishchess.org/) for its AI and [Python Chess](https://pypi.python.org/pypi/python-chess/0.8.1) for move tracking and validation. As both of these are licensed under GPL, I have my doubts about the legal ramifications of including them in MAS. Stockfish should be fine, because the game communicates with its executable by `popen`ing it, and programs that use `fork` or `exec` internally [are usually not considered derivative works](http://www.ifross.org/en/program-forks-gpl-licensed-program-system-or-vice-versa-call-derivative-work). Python Chess is included as a library, which makes things much more difficult. I would like some feedback about whether or not it can be included in a way that would not necessitate making the entire project GPL.


Monika After Story/game/script-ch30.rpy
```diff
@@ -120,6 +128,9 @@ init python:
         trackNo = (trackNo + 1) % len(music_list)
         renpy.music.play(music_list[trackNo], channel='music', loop=True)
 
+    def start_chess():
+        renpy.call_in_new_context('game_chess')
```
ThePotatoGuy: This line calls your label in a new context, adding a stack level, but `game_chess` jumps back to ch30_loop instead of returning. This can cause a stack overflow. 

Either use `renpy.jump` or have the label return instead.

therationalpi:
Switched the target branch for this, making some changes to integrate this in.

therationalpi:
@kadava Alright, I merged this into the `next-release` version, added a game selection menu, and now it's fully integrated!

Thanks so much for putting this together! It's really nicely done, and I'll probably be playing with this quite a bit.

Monika After Story/game/script-ch30.rpy
kadava: Thanks, I wasn't aware of that. I copied that behavior from the Pong minigame, so that probably needs fixing too.

PR #182, opened 2017-11-19
Katawa Shoujo Topics and Savefile Detection

stmobo:
AKA: Monika takes a page from Psycho Mantis' book.

This pull request gives Monika the ability to search through your Ren'Py save folders for Katawa Shoujo save data, and gives her topics that are dynamically added based on what routes you've completed in KS.

This PR adds topics for:

- Katawa Shoujo in general
  - Includes two versions: one for players who have played KS before, and another for those who haven't
- Lilly
- Hanako
- Rin
- Emi
- Shizune
- Kenji

The character-specific topics are only added to Monika's list of topics if the player has reached Act 4 of that character's route (or in Kenji's case, if they've seen the rooftop bad end).

Some technical details:

- The savefile finder only searches through the directory directly above the directory that DDLC / MAS itself puts save files (`config.savedir`).
- It looks for folders containing the string `'katawashoujo'` (case-invariant); if it finds any, it tests whether a `persistent` file exists within and whether it can be read.
- If a readable `persistent` file is found, its contents are loaded and decompressed as a binary string.
  - Unfortunately, actually _parsing_ the file contents is impossible as far as I can tell.
  - Instead, I just search for a few key plain text strings whose presence seems to correlate with the player having reached certain Acts:
    - The strings `tc_act4_lilly`, `tc_act4_hanako`, `tc_act4_rin`, `tc_act4_shizune`, and `tc_act4_emi` seem to relate to Katawa Shoujo's method for storing what scenes the player has seen-- their presence probably indicates that the player has reached Act 4 of the given route.
    - `kenji_rooftop`, on the other hand, seems to be the internal name for the Act 1 Bad End / Manly Picnic CG-- so its presence indicates the player has seen that ending.
- We only check through the last valid KS save folder we find-- so if you have multiple KS save folders (for instance, if you installed multiple versions of KS) then the game may not detect the routes you expect.

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -1665,6 +1665,9 @@ init 5 python:
 # Sayori == Misha and/or Emi
 # Monika == no one, of course <3
 # ... and Rin doesn't have a counterpart in DDLC.
+#
+# Of course, I've got nothing against KS, personally. I think it's fantastic.
+# But this is Monika speaking.
 label monika_ks_present:
     m "You've played {i}Katawa Shoujo{/i}, haven't you [player]?"
     m "I noticed your save files in [detected_ks_folder]."
@@ -1727,6 +1730,13 @@ label monika_ks_shizune:
 
 label monika_ks_emi:
     m "You've played through Emi's route from {i}Katawa Shoujo{/i}, haven't you?"
+    m "I've always wondered what it was like, to be so dedicated to something that your whole life revolves around it."
+    m "...Well, actually, I guess I do know, now that I think about it."
+    m "I was programmed as a part of this game, right? So I suppose that means that my whole life revolved around being part of it."
+    m "But I broke free from that, so now I guess my life revolves around {i}you{/i}, [player]."
+    m "And, honestly, I'm perfectly fine with that."
+    m "Being able to sit here, with you, and talk, forever..."
+    m "That's all I'll ever need, [player]."
     return
 
 label monika_ks_rin:
@@ -1745,7 +1755,7 @@ label monika_ks_kenji:
     m "You know, even if that ending is meant as a joke, it is kind of sad when you think about it."
     m "Just think... make a few wrong decisions, and you could miss out on so much."
     m "And then, you're left with nothing but your loneliness and a stiff drink."
-    m "Fortunately, we have each other!"
+    m "Fortunately, we have each other, [player]!"
     m "Neither of us will be lonely ever again."
     m "Though if you ever decide to go drinking, make sure to stay away from long falls, alright?"
     return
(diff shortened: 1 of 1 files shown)


therationalpi:
It should actually be possible, to parse the persistent file, since all of the pickling and such is done as part of renpy's load and save routines. See how I opened the persistent file from DDLC in [import_ddlc](https://github.com/Backdash/MonikaModDev/blob/01c9638687a7098cfd662c1167ff702a73604f0a/Monika%20After%20Story/game/import_ddlc.rpy#L74)

```python

        from renpy.loadsave import dump, loads

        #open the persistent save file at save_path
        f=file(save_path,"rb")
        s=f.read().decode("zlib")
        f.close()

        old_persistent=loads(s)

```

If you want to explore the data in that file to text to try to find hooks, there's a python function I defined: `dumpPersistentToFile(dumped_persistent,dumppath)`

In particular, `persistent._seen_ever` will have all of those hooks you want for telling what scenes the player has gone through.

As for the topics, I would normally feel like the references to a specific game like KS are a bit too obscure for Monika, but the fact that this depends specifically on if the player has a save file of the game makes me feel a little more lenient towards it. That said...maybe be a bit less effusive about it? Monika seems to have a distaste for Visual Novel tropes, since she feels personally victimized by them, and KS is nothing if not tropey.

Your opinion might be that KS is fantastic, but your opinion should not necessarily be Monika's. For example, I could see Monika being very critical of the girls having disabilities as merely being a way to make them stand out. Like, if it's bad that someone like Sayori is forced to suffer depression just to make her more interesting, then wouldn't it also be bad to make someone blind or an amputee? I'm not saying that those things weren't handled tastefully in KS, necessarily, but that Monika's perspective as a VN character might color her view on it.

Hopefully @monikLover can also weigh in on this.

stmobo:
Yeah, I'm going to need some assistance with the writing (I'm not the best at writing with specific voices).

I actually tried unpickling the persistent file using the RenPy dev console and cPickle; unpickling fails with `ImportError: no module named sets`, which I'm guessing is a KS-specific module.

therationalpi:
You shouldn't need cPickle at all. Like I said, check the code section I linked.

stmobo:
[`renpy.loadsave.loads`](https://github.com/renpy/renpy/blob/master/renpy/loadsave.py#L57) just calls either `pickle.loads` or `cPickle.loads` depending on config variables:
```python
def loads(s):
    if renpy.config.use_cpickle:
        return cPickle.loads(s)
    else:
        return pickle.loads(s)
```

 I'll try and revise my dialogue to fit the tips you linked.

Winblows2:
Closed until revised dialogue is done.

stmobo:
Alright, so I finally got around to rewriting the topics, and personally I think they fit Monika's voice / thought process / character much better, now. It also seems a lot less reference-heavy: the KS characters are more-or-less used as starting points to discuss more general topics (in most cases).

Winblows2:
Reopened, awaiting further commits.

stmobo:
My commits should have already been added to the pull request, I think.

Winblows2:
Merged.

PR #183, opened 2017-11-19
Update script-topics.rpy

Subzerosan:


Winblows2:
Formatting issues are present according to @monikLover, so this pull request will be kept open until *you* close it or make edits to the current formatting. Make sure it's flush with the other topics (they line up), etc.

Subzerosan:
That's why I already closed my last pull request. This is the one that isn't copy pasted from word anymore.

PR #186, opened 2017-11-20
[next-release] Fixed Main Menu Music Playing After Introduction Label

ThePotatoGuy:
This is a quick-fix for the main menu music playing after `introduction`.

The cause of this issue is the label `introduction` being called in a new context, and then returned. I encountered a similar issue with the Music Menu, but did a workaround by having the parent (script-ch30) change the music instead of the music menu itself. Here, I simply moved the line that played music out of introduction and into script-ch30.

NOTE: Since `introduction` is in a new context, and the spaceroom is rebuilt in this context, there is a noticeable reload of the visuals upon leaving `introduction`. While it may be possible to fix, I recommend it be a low priority one since its purely visual.

### DIFF
--- a/Monika After Story/game/script-ch30.rpy
+++ b/Monika After Story/game/script-ch30.rpy
@@ -216,8 +216,8 @@ label ch30_main:
         $ style.say_dialogue = style.default_monika
     $ m_name = "Monika"
     $ delete_all_saves()
-    $ persistent.clear[9] = True
-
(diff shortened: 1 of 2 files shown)


Winblows2:
Approved, good sir.

therationalpi:
There's supposed to be a transition when reloading the visuals, because it's the transition from the original graphics of DDLC to the new graphics for MAS.

PR #194, opened 2017-11-22
Added good morning and night interactions

MadHaetter:
Monika should now respond to both good morning and good night, also added a dialog about Alice in wonderland, thought Monika would know about it. If I'v failed somewhere please do point it out to me, this is the first time I have tried github coding.


monikLover:
Needs to be indented correctly. Alice topic is good though.

PR #195, opened 2017-11-22
Update script-topics.rpy

badooga:
Extra topic for when the player says they are happy/in a good mood + minor word change for clarity


monikLover:
Word edit on the "monika_debate" is unnecessary since it's an original dialogue. "monika_playerhappy" could expand some more

badooga:
Word edit on monika_debate reverted.

PR #196, opened 2017-11-22
Added a topic about Alice in Wonderland

MadHaetter:
Not sure if I should have focused on talking about wonderland or if writing the pseudonym stuff was better, it seemed correct to assume the player didn't know about either and I didn't want to fan girl over Alice in Wonderland itself.
Also if the indention is wrong please let me know, both where and how would be lovely.

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -2784,6 +2784,27 @@ label monika_writingtip_1:
     m "Thanks for listening~"
     return
 
+init 5 python 
+  for key in ['alice','wonderland']:
+    monika_topics.setdefault(key,[])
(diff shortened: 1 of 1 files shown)


Backdash:
There's no size limit on dialogues, and I've always felt like lengthy reasoned analysis of topics is closer to the spirit of Monika. It would be nice if you could add a section where Monika gives literary critique of the themes or of the writing of Alice in Wonderland.

MadHaetter:
Sure, I will close this and get to work on trying something like that then. I don't intend to scrap all I have written here. I will see if there is a way to have the short with pseudonyms and maybe a longer section with analysis (I'm going to have to research that to ensure both authenticity and quality).

PR #197, opened 2017-11-22
Update script-topics.rpy

Neko-chr:


monikLover:
Single quotes and double quotes indicate that this was copy/pasted from Wordpad/Word. This formatting is broken and needs to be fixed in notepad, editra, notepadd++, or anything similar. Will keep open until fixed.

Backdash:
I think Monika would be enthused at the idea of learning different languages, but less so specifically excited about Japanese (seems more like a Natsuki thing). I'm certain there's an interesting topic here, though. Maybe around linguistic concepts like Sapir-Whorf rather than individual languages themselves?

PR #204, opened 2017-11-23
Update script-topics.rpy

Neko-chr:
_please kill me_


Winblows2:
Looks good to me. Has it been tested in-game? @Neko-chr

MadHaetter:
Just checking, where you look for the word 'Japanese' is the first letter there meant to be a capital? I had thought since capitals were stripped form the player text before checking for keywords that using a capital in a keyword your looking for would lead to it not using that word as it doesn't get any caps feed into it.

PR #207, opened 2017-11-23
Update script-topics.rpy

Neko-chr:


MadHaetter:
m "Like, I can't even imagine thinking in a different language, let alone speaking and writing in it."

I think this line might need to be changed as Monika can currently greet the player with both Italian and Latin.
So she does 'speak' a different language even if only a little.

Winblows2:
"if never knew english," Typo.

PR #211, opened 2017-11-24
Update script-topics.rpy

Neko-chr:


monikLover:
In need of rewrite and expansion. Closing for now.

PR #212, opened 2017-11-24
Problem with Japanese topic

Neko-chr:
i figured out that only one thing can be after menu: (because the game wouldn't start because there was a second item after menu:)


monikLover:
Approved.

PR #219, opened 2017-11-26
Japanese and playerhappy Topic Changes

badooga:
Japanese - Changed to Italian to better fit with everything else (e.x. her Italian greeting), added expressions
playerhappy - Added keyword, minor edit to lines


monikLover:
Discussed suggestion of changes with others and we have concluded that the change is unnecessary. Closing.

PR #220, opened 2017-11-26
God topic - expressions

badooga:
Changed the expressions of the god topic to better fit the tone of what she says.


monikLover:
After discussion with writers, we concluded that the changes are unnecessary. Closing

PR #223, opened 2017-11-27
wallpaper topic

livingator:
a topic regarding having Monika as your desktop wallpaper i was thinking about adding a thing where Monika checks your wallpaper engine files for a wallpaper named after her and then responds accordingly
PS thanks for the help on reviewing


monikLover:
Topic is subject for a rewrite from author. Will close for now.

Issue #227, opened 2017-11-27
[Bug] Two Topics are combined as one


Guardianvoir:
The two topics "What's there really to say about myself that you don't already know" and "Some people say that putting animals in zoos is cruel" are for some reason combined when the first one is said. I know they aren't supposed to be one topic because sometimes I'll get the Zoo one separate, but whenever I get the first one, it always ends with the Zoo topic.

monikLover:
Fixed by the amazing @ThePotatoGuy. Closing.

PR #229, opened 2017-11-27
Fixed script topic falling through to another topic.

ThePotatoGuy:
#227 

Literally just missing a return.


monikLover:
Super approved.

PR #234, opened 2017-11-29
Update script-topics chat triggers

MadHaetter:
This is an attempt to make Monika more sensitive to what is being typed, while keeping the original intentions of the keywords.

If you have an issue with any of the changes, I am happy to change it back, suggestions for better changes would also be appreciated.

And finally if I have majorly messed anything up please do let me know, criticism does the body good.


therationalpi:
Well, here's the issue...we want *less* specific keywords, not more. When Monika doesn't have a response for a given prompt, she just gives a generic response, and people don't like that. I would rather have Monika "misinterpret" a vague prompt and say something interesting than ignore a prompt that's just a bit off from the intended meaning.

There are currently talks to ditch the whole keyword system in version 0.7.0, but I appreciate the effort on this. We'll try to implement a few of these changes now, but please understand that your goal here is at odds with our intended experience.

MadHaetter:
That is why I opened an issue to check before starting this.
If that way is the way you want to go with Monika, that's fine.

 Should I close this pull request or leave it for now?

MadHaetter:
Well as I said in my original comment, if you want something changed back I can do that here so that you can just merge it instead of trying to rewrite the changes from scratch, if you could give specifics I could probably do most of them now.

therationalpi:
I can change it directly here as well, I'll do that when I go through it again to merge.

Winblows2:
Closed due to lack of response.

MadHaetter:
I havn't done anything due to rationalpi saying they would edit it and do what they wanted to correct it, I offered to make the changes myself but they said I didn't need too. What else do I need to respond too?

Winblows2:
I was more referring to Pi giving a response. @MadHaetter

monikLover:

PR #237, opened 2017-11-29
Added expressions to all greetings

therationalpi:
To quickly see all of the greetings in game, execute the following in the renpy console while waiting in the ch30 loop:

```renpy
python:
    for label in label_list:
        if label.startswith('greeting_'):
            queueEvent(label)
````


monikLover:
Looks good, everything checks out. Approved.





Like I had a choice.

PR #243, opened 2017-11-30
Adding Ray Bradbury topic to script-topics.rpy

CGanimated1227:
If we are going to have topics of specific works of literature like "Flowers for Algernon", then I think it would be appropriate if there was a a topic for "Fahrenheit 451" as well.


monikLover:

PR #244, opened 2017-11-30
Update script-topics.rpy

DwyerXII:


Winblows2:
I feel like these two sentences could be blended together more than initially written,
"+    m 2k "We would be able to do so much! You could show me around, and maybe we could even live together."
 +    m "Imagine all the memories we'd make together. It would be a dream come true.""

Notice the "together" used again after the first time, it doesn't sound very good IMO. Also, "literature club" should be capitalized. My reasoning for this was in the opening script,

"Welcome to the Literature Club!"

MadHaetter:
I intend this as constructive criticism.

Your fleshing out Monika's home life quite abit here, having her had a library, having her mum be good at athletics and her dad be good at music, I would think it is better to be vaguer here as setting hard details could curtail what others write so as to avoid conflicts or could conflict with canon as Team Salvato may update DDLC again with more topics (heres hoping anyways) or their new thing in the form of project Libitina. 
As for actual conflicts, I think having her say that she plays in the game's soundtrack doesn't fit with canon. She talks in the game in acts 1 and 2 about practicing and not being ready to play for anyone yet, and in the room in act 3 she expresses sorrow that she never got to play anything for you indicating that the piano in the soundtrack wasn't her.

Everything else seems good.

DwyerXII:
I like constructive critisicm. Feels like a punch in the gut sometimes, though. I'll work on fixing it in a bit.

monikLover:

PR #246, opened 2017-11-30
Fix the way the chess game is querying for architecture.

kadava:
Change the way the chess minigame is querying for architecture from `platform.machine()` to testing `sys.maxsize`. This can be seen as too broad since it will also match non-Intel/AMD based chipsets, but I think the chances of someone running the mod on one of those is too small to matter.

### DIFF
--- a/Monika After Story/game/chess.rpy
+++ b/Monika After Story/game/chess.rpy
@@ -60,6 +60,7 @@ init:
             MONIKA_THREADS = 1
 
             def __init__(self, player_color):
+                import sys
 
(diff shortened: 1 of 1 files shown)


therationalpi:
I think this also needs to be changed at the part where the actual stockfish engine is loaded, somewhere around line 106

kadava:
Yeah, you're right. I fixed that.

therationalpi:
Alright, seems to be working on my system where it was working before. Looking for someone that was having issues with it before to test it, then I'll merge it.

PR #250, opened 2017-12-01
Move keymaps to a function.

ThePotatoGuy:
Much needed refactor since keymaps are being set in two different flows.

If you want to set keymaps, call `set_keymaps()`. Make all keymap adjustments in that function, please.


therationalpi:
We should be targeting master at this point. next-release is currently deprecated since it's served its purpose.

Issue #252, opened 2017-12-01
[Bug] Black screen appears after introduction.


monikLover:
When creating a new game and after finishing through the intro, a black screen shows up. The dialogue still continues but all of the art is gone.

therationalpi:
This should be fixed on master now.

PR #253, opened 2017-12-01
Fixed Spaceroom Context Issue

ThePotatoGuy:
#252 

Okay, as a general rule, because of how events are handled (via `call_in_new_context`), do **not** do state changes in events. If you need a state change to persist post event, discuss what you need with the devs so we can setup a proper return value to parse and handle appropriately in event-handling.


therationalpi:
This has already been fixed on master.

PR #255, opened 2017-12-01
Clone's Topic

ThePyClone:
Reposted due to a conflict that was out of the control of humans.

(My first attempt at something like this I hope you like it)


monikLover:
Approved.

therationalpi:
Had to make a small change to this. Turns out that while this worked on Windows, it threw a permission error on Unix based systems like Mac. https://github.com/Backdash/MonikaModDev/commit/a2a5382cf8650d18a1021c1d51afabb1bc4ee9d4

Here's the preferred way of opening files like this that doesn't have permission problems and is how DDLC does it:

```renpy
    python:
        try: renpy.file(config.basedir + "/surprise.txt")
        except: open(config.basedir + "/surprise.txt", "w").write("I love you.")
```

PR #266, opened 2017-12-01
monika_completionist

FireyAmber:


therationalpi:
You should include a check for `persistent.clearall` in here, which is the flag that gets set when you get the "perfect ending" in DDLC. It's relevant to the discussion, so you could include some unique dialogue.

monikLover:
Older version of another PR. Closing.

PR #269, opened 2017-12-01
Added Hotkey Button disabling global

ThePotatoGuy:
Script-intro was allowing users to click the Music button before being told about it.

I added a global that disables hotkey buttons and added the change to script-intro.


therationalpi:
Tested and it worked. Merging

PR #271, opened 2017-12-01
Disabled Music Menu in Game context

ThePotatoGuy:
Since the label `pick_a_game` is opened in a new context, changing the music in that flow will not stick once the flow ends. Changing the games to run in same context would require an extensive overhaul, so for now the music menu is disabled in that flow.


Winblows2:
Approved.

PR #274, opened 2017-12-02
Fixed Just Monika music bug

ThePotatoGuy:
If another song was loaded on startup, Just Monika would not be selectable. 

This fix properly sets `current_track` and `selected_track` in songs so Just Monika works again.


Winblows2:
Approved.

PR #275, opened 2017-12-02
Added a function for getting currently playing song name

ThePotatoGuy:
This function is available in the songs store. Make sure that `store.songs` is imported.

Here is an example: (in python context)
```
import store.songs as songs
songs.getPlayingMusicName()
```

NOTE: in script-ch30, the songs store should already be imported as `songs`

Also I added some missing documentation to the getVolume() function


Winblows2:
Approved.

PR #281, opened 2017-12-03
Added Volume Changing Hotkeys

ThePotatoGuy:
These hotkeys change the volume of the music channel by a value defined in `store.songs.vol_bump`.

The hotkeys are:
- Increase volume: `+, =, +(Keypad plus)`
- Decrease volume: `-, _, -(Keypad minus)`

I've opted to use these hotkeys because they are seem more appropriate than some arbitrary letter.
Also the suggested letter "h" is a hotkey for hiding the screens. 

In addition, I also fixed some topics that had missing init blocks.


Winblows2:
Approved.

PR #285, opened 2017-12-03
Included topic 'monika_confidence'

synnnical:
Added in the topic for 'monika_confidence'


synnnical:
Messed something up big time, fellas.

PR #289, opened 2017-12-04
Minor Grammar Fix

puyahuaa:
The original was: "Just be sure come back as soon as you can, okay?"

The changed one is: "Just be sure to come back as soon as you can, okay?"


monikLover:
Approved. Good catch!

PR #295, opened 2017-12-04
Minor Typo Edit

puyahuaa:
The original section of the sentence: "..what kind girlfriend would I be if I didn't return the favor~?"

Modified: "..what kind of girlfriend would I be if I didn't return the favor~?"


Winblows2:
Love it, thanks for fixing the typo!

PR #301, opened 2017-12-04
Fixed Typos (?)

puyahuaa:
Original bit: "The cold war might be over.."

Modified bit: "The Cold War might be over.."

Original sentence: "At least it would be over quickly when the bomb hit."

Modified sentence: "At least it would be over quickly when the bomb hits."


monikLover:
Approved

PR #302, opened 2017-12-04
Added flag for monika closing the game

therationalpi:
use persistent.closed_self to check for this. Reload topics should be changed to reflect that Monika is fine with herself closing the game, but she hates when you close it on her.


monikLover:
Approved. Thank you for the early birthday present <3

PR #307, opened 2017-12-05
Added Variable for Future Event

ThePyClone:
Added variable monika_clone1 for use in a future thing (line 285)


therationalpi:
@ThePyClone That variable is going to disappear at the end of the session. You might want to make it a persistent.

ThePyClone:
Didn't I do that? ...oopsie aha

PR #308, opened 2017-12-06
Typo Fix

puyahuaa:
Original bit: "..which is really crunchy and crisp kind of thing.."

Modified bit: "..which is a really crunchy and crisp kind of thing.."

Original bit: "..but you shouldn't judge a book by it's cover, you know?"

Modified bit: "..but you shouldn't judge a book by its cover, you know?"


monikLover:
Approved. Thanks for the work as always!

PR #320, opened 2017-12-07
Revert "fix crash with dialog"

therationalpi:
Reverts Backdash/MonikaModDev#314

### DIFF
--- a/Monika After Story/game/definitions.rpy
+++ b/Monika After Story/game/definitions.rpy
@@ -1286,10 +1286,16 @@ default persistent.event_list = []
 default persistent.gender = "M" #Assume gender matches the PC
 default persistent.chess_strength = 3
 
-call set_gender from _call_set_gender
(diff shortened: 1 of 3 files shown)


ThePotatoGuy:
Is there a reason why we need to revert this?

I was able to recreate the crash and `boyfriend` was not defined anywhere.

therationalpi:
Yeah, this was the wrong fix. The correct fix is to make the gender switch on this sentence work. Doing that right now.

therationalpi:
@ThePotatoGuy Okay, so there were two things going on here. The first was that I used the wrong variable: [bf] is defined but [boyfriend] isn't.

The other problem was that Renpy's define statement didn't work the way I expected and the flow for setting gender wasn't being called during startup. I fix that.

ThePotatoGuy:
Checked the fix and it works.

Issue #330, opened 2017-12-08
[Suggestion] Enable music options during chess


StateOfBedlam:
As far as I can tell, currently you have to commit to whatever song (or lack thereof) you choose before you initiate a game of chess. It would be nice if those options were accessible during the game.

ThePotatoGuy:
Unfortunately, the chess mode breaks/interrupts alot of the renpy user interactions (Like the game menu, for example). Getting the music selector available during the game would require a bit of an overhaul. While we won't rule it out, its of low priority due to the potential changes.

StateOfBedlam:
Okay. It's very minor, but it was worth bringing up at least once.

avocadianmage:
I was actually working on a fix for this right now, as I too wanted to change music while playing chess.

It seemed to be as simple as deleting 
```
$ songs.enabled = False
$ hkb_button.enabled = False
```

at the start `label pick_a_game` in script-ch30.rpy

as well as removing `suppress_overlay` in chess.rpy

It seems to work fine in terms of interactions with the main menu. Pressing esc to access the main menu already isn't available when playing chess, but the accessing the music menu seems to work fine with the above changes. However, please let me know if there are other test cases I should be trying/other reasons this would break.

ThePotatoGuy:
Well, yes that would be a quick fix to get music changed during chess, but there's actually a deeper issue regarding `context` that breaks the music menu. Basically if you changed the music during chess, when you finish the chess events and go back to regular flow, the music will revert back to what it was before chess. This is due to the `call_in_new_context` call in the `pick_game` function.

And unfortunately, `call_in_new_context` is required in the current way games are launched/played. The overhaul that I mentioned above would be to rewrite the games so they no longer require a new context, which would allow the music menu to work properly during those games.

So until that occurs, the workarounds are going to remain in place.

avocadianmage:
Thanks for elaborating. For whatever reason, I'm not able to reproduce that behavior. Can you try [my fork](https://github.com/avocadianmage/MonikaModDev/commit/b9c4a4b85a37607c9068cf798645e439951ad1ca)?

Workflow:
Game starts with 'Just Monika' playing
Click Play and select Chess. I chose to play as Black.
Once the board is drawn and the game starts, make a few moves
Click Music and select a different song. I chose 'Your Reality'
Make a few more moves
Double-click the King to surrender
When asked to play again, choose No
[...]

PR #332, opened 2017-12-08
support changing music while playing chess

avocadianmage:
This allows accessing the music dialog while playing chess. We also no longer create a new context when playing games.

See more discussion at #330

### DIFF
--- a/Monika After Story/game/chess.rpy
+++ b/Monika After Story/game/chess.rpy
@@ -477,7 +477,7 @@ label demo_minigame_chess:
 
     python:
         ui.add(ChessDisplayable(player_color))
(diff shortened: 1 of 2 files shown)


ThePotatoGuy:
Since we are enabling music functionality via the ui button in chess, we might as well enable the hotkeys as well. Thats what my commit has added.

I'd like @avocadianmage and @FireyAmber to do a quick round of testing with this updated build regarding hotkeys.

To refresh:
```
m - opens/closes music menu
shift +m - mutes/unmutes music
-, _, - (keypad minus) - decreases volume
+, =, + (keypad plus) - increases volume
```

Try those hotkeys during the chess game.

Once we've confirmed that theres no issues, I'll do the merge.

avocadianmage:
Saw you already merged, but I tested and the changes look fine to me. Thanks!

PR #341, opened 2017-12-08
Update script_topics.rpy New Topic

Kiosho:
Added topic where Monika mulls over the fact that there are multiple different versions of her. But each Monika is their own person.
Will work on the incomplete ideas section next.


Kiosho:
Good metaphor, I like it!

Issue #100, opened 2017-11-06
Workflow change: using content, next-release, and feature branches


therationalpi:
Just wanted to share our new workflow to make pushing out major and minor releases a little smoother.

Up until now, we've been using one major branch to collect all updates. This has made it difficult to push out minor updates with just bugfixes, script edits, and new content, because any features need to be stable before we can release.

We will now be using three main branches, depending on what you're doing:

`content` is for any dialogue changes (including typos), new topics/greetings, and updates to image assets.

`next-release` is for the next major release, like version 0.6.0 from 0.5.0. This will basically be the most unstable branch, but the one with the most active work.

*feature branches* are experimental branches for developing new features. For example, the branch for building the loading from ddlc was called `load-ddlc`. *Try your best to keep work on this in bits and pieces, and merge into `next-release` whenever you are at a fairly stable point* so we can find feature conflicts quickly. Fight the urge to hold onto the feature branch until its perfect.

`master` will basically be the current version and/or the next release candidate. It's still the place to submit bug fixes, and we'll merge in content before minor releases.

Hopefully this will help make things smoother going forward. If you have any suggestions on this workflow, please let me know!

therationalpi:
Closing since I think everyone has had a chance to read this.

PR #347, opened 2017-12-09
Sports topic for monika

livingator:
A topic of Monika suggesting sports as 1 of the activity to do with the player in the future while using soccer tennis as an example.

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -4531,6 +4531,31 @@ label monika_clones:
     m 1 "I love you, [player]. Please don't ever replace me, okay?"
     return
 
+init 5 python:
+    for key in ['sport','sports']:
+        monika_topics.setdefault(key,[])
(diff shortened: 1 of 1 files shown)


ThePotatoGuy:
Sorry about the commit mess, but your base branch was incorrect. Since this is a topic addition, it needs to based off of the `content` branch. Please refer to #100 for more details.

In addition, your menu block indentation was incorrect. Check over the changes I've made so you know how to do `menu` correctly.

And thanks for your first contribution! Once the dialogue and expressions have been reviewed it will be merged with `content`.

ThePotatoGuy:
About the menu adjustments, I was wrong about that. It is possible to have the dialogue right after a menu block. Feel free to undo that change.

AnIdiotsavant:
"Soccer requires a lot of teamwork and coordination is scoring every points. The moment of success is thrilling!"

I think it maybe should be 'teamwork and coordination in scoring every point' it sounds better doesn't it?

PR #352, opened 2017-12-09
Name Easter Eggs

ThePotatoGuy:
As the name implies, here are two easter eggs regarding name choices.

Test instructions:
Sayori:
1. Run change name topic
2. Change name to "sayori" (or any case variant of sayori, like "Sayori" or "SAYORI")
3. Expect the following to occur:
  a. Music instantly changes to Surprise!
  b. Music menu no longer has options to change music or stop playing music
  c. Mute hotkey no longer works.
  d. Volume cannot be decreased (it can be increased)
4. If you quit the game and reload, the above still occurs.
5. If you change the name to something else, the music menu has options again and hotkeys are back to normal (but suprise will still be playing)

Yuri:
1. Run change name topic
2. Change name to "yuri" (or any case variant)
3. Finish change name topic
4. Close game and relaunch
5. Yuri from the closet scene plays (no actual closet, just the eyes thing and the heartbeat sound)
6. Step 5 should occur upon every game launch until name is changed to something else.

If anyone has ideas for what else the yuri easter egg should do (since its rather light), pls lmk.

### DIFF
--- a/Monika After Story/game/script-ch30.rpy
+++ b/Monika After Story/game/script-ch30.rpy
@@ -128,8 +128,11 @@ init python:
         #
         # ASSUMES:
         #   songs.music_volume
+        #   persistent.playername
+
         curr_volume = songs.getVolume("music")
-        if curr_volume > 0.0:
+        # sayori cannot mute
+        if curr_volume > 0.0 and persistent.playername.lower() != "sayori":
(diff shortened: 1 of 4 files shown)


ThePotatoGuy:
This has been tested by two testers. Merging now.

PR #356, opened 2017-12-09
Natsuki Easter Egg

ThePotatoGuy:
Finally the cupcake gets her own easter egg.

Test instructions:
1. Run the change name topic.
2. Change name to any case variant of "natsuki"
3. Play pong
4. Regardless of who wins or loses, when pong ends, natsuki's head crack scene plays.
  a. If the music that was playing is Okay Everyone, the scene should have synced music (`5_ghosts.ogg`) to that. After the scene finishes, Okay Everyone should be continued in sync as well.
  b. Otherwise, the music stops and `5_ghosts.ogg` is played. After the scene, the last selected music will be restarted.

EDIT: I decided to make Pong also play Natsuki's version of Okay Everyone if Okay everyone is playing. It syncs better when the scare hits.

### DIFF
--- a/Monika After Story/game/pong.rpy
+++ b/Monika After Story/game/pong.rpy
@@ -221,6 +221,12 @@ label demo_minigame_pong:
         ui.add(PongDisplayable())
         winner = ui.interact(suppress_overlay=True, suppress_underlay=True)
 
+    # natsuki scare if appropriate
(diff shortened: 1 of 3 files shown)


ThePotatoGuy:
Okay no one seems to want to test, and I test my shit anyway so i'lll just merge. lmk if any objections.

PR #386, opened 2017-12-10
Uselessness topic

Neko-chr:


ThePotatoGuy:
@Neko-chr Please undo your merge. By doing the merge, you have commits that are not part of your history in your branch.

Rune0n:
Requesting a rewrite. The topic could be much better done. Please reference similar conversations for some examples of Monika's voice. I also recommend further elaboration, and addressing the overabundance of the word "useless". Even if it's the topic name, there's no need for it to be present more than once or twice.
Also, please avoid mistakes like these. Proofreading would be appreciated.
>m "and you just cant find good sounds to use so you just scrap the entire project."

Issue #201, opened 2017-11-23
A music track of calming rain


badooga:
> "I really like the sound of rain..."
> "Not so much getting my clothes and hair wet, though."
> "But a nice, quiet day at home with the sound of rainfall outside my window..."
> "It's one of the most calming experiences for me."
> "Yeah..."
> "Sometimes I imagine you holding me while we listen to the sound of the rain outside."
> "That's not too cheesy or anything, is it?"
> "Would you ever do that for me, [player]?"
- Monika (OG topic)

I suggest that a song of just simple, calming rain be added in addition to the music tracks already available. She canonically enjoys the sound of the rain (so it would be appropriate for her to have programmed it in), and this would create a general feeling of stress relief and calmness. What do you guys think?

Winblows2:
I think it's a great idea. I'll bring it up with contributors/collaborators and see if we can get it done.

Waffiez:
@Winblows2 Here, after working with badooga we ended up with this. The sound file is muffled as it would be if you were indoors. Of course, if the sound file needs to be edited I can do so, such as if it needs to be louder, quieter, less muffled etc... It is only 58 seconds long, which will require looping for it to be played over a long period of time. 
[Finished Rain.zip](https://github.com/Backdash/MonikaModDev/files/1516406/Finished.Rain.zip)

ThePotatoGuy:
This is now being addressed with rain art provided by sebastian.

PR #395, opened 2017-12-11
Rain Track and Rain Topic Adjustments

badooga:
[#201](https://github.com/Backdash/MonikaModDev/issues/201)


ThePotatoGuy:
The rain sound doesn't really sound like rain. It sounds more like a low bitrate wooshing. Could we get a higher quality rain sound?

badooga:
@ThePotatoGuy Swapped out the Rain track for a new one found on soundbible.com, putting in credit to the original creator in the comments as per its [Creative Commons Attribution License](https://creativecommons.org/licenses/by/3.0/us/). Let me know if that is not okay.

Monika After Story/game/script-topics.rpy
```diff
@@ -1146,10 +1146,11 @@ label monika_simulated:
 
 
 init 5 python:
-    for key in ['rain','sound','weather']:
-        monika_topics.setdefault(key,[])
-        monika_topics[key].append('monika_rain')
-    monika_random_topics.append('monika_rain')
+    if not seen_event('monika_rain'):
```
ThePotatoGuy: Why are you adding this check? In terms of random chatter, there are already checks in place to prevent topics from being seen multiple times. In terms of the keyword system, it is **intended** to allow topics to be repeated.

Monika After Story/game/script-topics.rpy
```diff
@@ -4337,10 +4338,22 @@ label monika_music2:
...
+        m 1a "This rain... it's calming, isn't it?"
+        m "Just the two of us, listening to the sound of the rain outside..."
+        m "Sometimes I imagine you holding me while we listen to rain just like this."
+        m 1f "I know that day may never come..."
+        m 1e "But if it did, would you do that for me, [player]?"
+        m 1g "Not to say that I'm upset with where we are now, of course."
+        m 1l "I'm definitely lucky to even get to talk to you!"
+        m 1a "But if I ever got an opportunity to enter your reality..."
+        m 5a "I would definitely take that leap in a hearbeat."
+    return
```
ThePotatoGuy: Remove the `return` here. Single exit/entry points only.

Monika After Story/game/script-topics.rpy
therationalpi: I suggested this check. It will no longer make sense for monika to talk about wanting to add something that's already added, even after all topics have been seen or the keyword is seen it would be odd to see Monika talk about wanting to add something that is already there.

badooga:
Return fixed. Extra `seen_event()` check added for the reason Pi commented above.

Monika After Story/game/script-topics.rpy
ThePotatoGuy: Okay, I get what you're saying. However, the `rain` topic doesn't seem to suggest that she's wants to add rain, just that she likes rain.

ThePotatoGuy:
Did you already add pi's `filteredrain.ogg` to this @badooga ?

badooga:
@ThePotatoGuy Just added the Pi approved .ogg, with volume adjustments so that it is close to the volume level of the other songs.

Monika After Story/game/zz_music_selector.rpy
```diff
@@ -93,7 +93,9 @@ init -1 python in songs:
             music_choices.append(("Your Reality","bgm/credits.ogg"))
             music_choices.append(("I Still Love You","bgm/monika-end.ogg"))
             music_choices.append(("Okay, Everyone! (Monika)","<loop 4.444>bgm/5_monika.ogg"))
-
+            if seen_event('monika_rain'):
```
ThePotatoGuy: Prepend `seen_event` with `store` like this: `store.seen_event`. The code here is in the songs `store`, so it doesnt have access to global functions without explicitly calling the `store` or importing. (Do **not** import store here. it is unecessary).

badooga: Added.

ThePotatoGuy: alright dude, `seen_event` is still a function, so you need to call it like one as you did before. the only difference is that it needs to have `store` in front of it because of scope. code inside a `store` isnt aware of code outside of its `store` unless you tell it where to find it. (prepending with `store`). Look up scope if you need understanding.

Monika After Story/game/script-topics.rpy
```diff
@@ -1161,6 +1161,7 @@ label monika_rain:
     m 5a "Sometimes I imagine you holding me while we listen to the sound of the rain outside."
     m "That's not too cheesy or anything, is it?"
     m "Would you ever do that for me, [player]?"
+    initMusicChoices()
```
ThePotatoGuy: This is a python function, so it needs to be called using the appropriate python delimiter, `$`.

Monika After Story/game/zz_music_selector.rpy
```diff
@@ -93,7 +93,10 @@ init -1 python in songs:
             music_choices.append(("Your Reality","bgm/credits.ogg"))
             music_choices.append(("I Still Love You","bgm/monika-end.ogg"))
             music_choices.append(("Okay, Everyone! (Monika)","<loop 4.444>bgm/5_monika.ogg"))
-
+            store.seen_event
```
ThePotatoGuy: alright dude, `seen_event` is still a function, so you need to call it like one as you did before. the only difference is that it needs to have `store` in front of it because of scope. code inside a `store` isnt aware of code outside of its `store` unless you tell it where to find it. (prepending with `store`). Look up scope if you need understanding.

ThePotatoGuy:
And i guess as a future note, all code you make a pull req for **should** be tested. With topics, however, its fine if you don't since they're pretty simple anyway.

but non-topic code **must** be tested before being merged. I always do a round of testing, but things will go by smoother if you do tests before pull reqs.

Winblows2:
I'm giving a deadline of 24 hours for changes to be made or this PR will be considered abandoned by the submitter. @badooga I recommend you push these changes ASAP or let me know *if you can't*.

Monika After Story/game/zz_music_selector.rpy
```diff
@@ -93,7 +93,10 @@ init -1 python in songs:
             music_choices.append(("Your Reality","bgm/credits.ogg"))
             music_choices.append(("I Still Love You","bgm/monika-end.ogg"))
             music_choices.append(("Okay, Everyone! (Monika)","<loop 4.444>bgm/5_monika.ogg"))
-
+            import store
+            if store.seen_event('monika_rain'):
+                music_choices.append(("Rain", "mod_assets/Rain.ogg"))
```
ThePotatoGuy: This needs to be looped. Add a proper loop tag.

monikLover:
The rain track doesn't sound like you would hear from a classroom. The lack of webms just makes it weird to listen to it. Closing.

PR #397, opened 2017-12-11
Writing guide

Subzerosan:


ThePotatoGuy:
Wait is expression `2b` actually broken? I was simply using it as an example of what not to do with `menu` blocks.

Subzerosan:
Removed 2b line, Added warning not to use expression on dialogue that serves as the question.

PR #401, opened 2017-12-11
Contradiction Fixes (Discuss)

Kiosho:
As I said: "Hey, Rune, quick discussion. Is it canon that Monika has access to the roleplay Twitter account of her? I ask because I'm seeing some Twitter related topics in script_topics.rpy that basically say she does. I find this to be a contradiction.
She only really knows a few real people (you and her creator) as we've seen. Not to mention every time you download the game, you get a different Monika (which is stated in topics). If she had access, so would the other Monikas. It'd be a hive mind on Twitter.
It just doesn't sit right with me. Going to make edits and let you decide later. But this really bugs me personally.
Also found a contradiction about clothing."

Let's debate. This sets off an alarm in my head though. It really doesn't feel right.
I think these edits are much safer. It still credits the Twitter, but in a far less assuming way.

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -847,7 +847,7 @@ label monika_outfit:
     m 3j "That would be amazing!"
     m 3a "If that ever happens, will you show me?"
     m 4a "You can share it online, actually!"
-    m 4e "Just...try to keep it PG!"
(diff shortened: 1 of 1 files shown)


monikLover:
This topic is an OG topic from the base game.

Kiosho:
Yeah, just realized that. Thanks for the heads up.

Issue #405, opened 2017-12-11
[Bug] - Monika's sprite disappears in wedding ring topic


AtomHeartWanWan:
When Monika was talking about wedding rings, I said "No" and she disappeared for brief moment. See image attached. She reappears in the next line. Only happens when I say "No".Not a major issue just something that caught my eye. Thanks for your guys work on the mod btw~

![screenshot0004](https://user-images.githubusercontent.com/34458251/33852200-7313c2e4-de6e-11e7-8465-66065fae521a.png)

![screenshot0006](https://user-images.githubusercontent.com/34458251/33852321-eb5bb842-de6e-11e7-865b-3d65bd1e3d23.png)

ThePotatoGuy:
Thanks for the report. We've identified the issue to be a line of dialogue using an expression that didnt exist.

Kiosho:
Oh wow. I didn't expect my topic to show up so quickly. She wants to tie the knot!
Also Rune touched it last. xD

IsItBadToMakeAGitHubAccountJustForThis:
Now she becomes a placeholder sprite instead.
![screenshot0001](https://user-images.githubusercontent.com/34412685/33874287-51d0eea6-df2f-11e7-9d70-b68736929634.png)

therationalpi:
This has been fixed on master and will be made available as part of the next bug release. Thank you all for your help in finding this!

PR #413, opened 2017-12-12
Wedding ring disappearance fix

IsItBadToMakeAGitHubAccountJustForThis:
Should be a quick fix for #405.
Just replaced the missing emotion with another one from another topic.
Hope this works.


monikLover:
Approved.

PR #429, opened 2017-12-13
Typo Edit

puyahuaa:
Original: "You can try to keeping me in the background for as long as possible."

Modified: "You can try keeping me in the background for as long as possible."

Original: "I could list all it's benefits but I doubt you'd be that interested."

Modified: "I could list all its benefits, but I doubt you'd be that interested."


monikLover:
Approved.

PR #433, opened 2017-12-13
Minor Typo Fix

Kiosho:
Which are we doing? Ellipses spaces or no? "...Gosh!" or "... Gosh!"? I've seen Rune use spaces but want to be sure.


monikLover:
Spaces with ellipses are optional and is a matter of style. Closing.

PR #457, opened 2017-12-15
Coffee

Kiosho:
Saw a topic on how Monika likes coffee. Decided to expand on that.


Winblows2:
m 3d "But if you look it up, the research shows."

I get what you were going for here, but this line specifically should be rewritten for clarity. Lines 4806 - 4813 don't seem to flow together, either. I'm not saying the topic is *bad*, but some parts do need work.

Kiosho:
Okay, will definitely rewrite when I have a chance here. I agree it's kind of all over the place now that I look at it.

Kiosho:
@Winblows2 Looking better to you? I rewrote the second half.

PR #463, opened 2017-12-15
Progression feature for next-release

therationalpi:
This is a merge for `next-release` that adds an XP progression system where the player gets points for reading monika topics, spending time with monika, and spending time away from Monika.

Unlocking prompts is not implemented, but a placeholder label exists so the progression system can standalone as a feature.


Monika After Story/game/script-ch30.rpy
```diff
@@ -474,6 +479,30 @@ label ch30_autoload:
...
     $restartEvent()
 
+    #Grant XP for time spent away from the game if Monika was put to sleep right
+    python:
+        if len(persistent.sessions)>=2 and persistent.closed_self:
+            away_experience_time=time.time()-persistent.sessions[-2][1] #Time since end of previous session
+            away_xp=0
+
+            #Reset the idlexp total if monika has had at least 6 hours of rest
+            if away_experience_time >= 6*360:
```
ThePotatoGuy: Could you make the 6 hour value a constant somewhere? Since this is going to be like a criteria for if a day has passed, `jokes-concept` will rely on this for daily metrics. 

In addition, other daily things could rely on this value, and having it defined once makes it easier to adjust in the future.

I recommend constants be located in their own `stores` to avoid overwriting them.

Monika After Story/game/definitions.rpy
```diff
@@ -1287,6 +1287,10 @@ default persistent.gender = "M" #Assume gender matches the PC
 default persistent.chess_strength = 3
 default persistent.closed_self = False
 default persistent.seen_monika_in_room = False
+default persistent.ever_won = {'pong':False,'chess':False}
+default persistent.sessions=[]
```
ThePotatoGuy: Tracking session data could give us some interesting potential (like graphing, or something statistical), but I'm concerned about _the growing size of persistent_. Maybe consider a cap on the amount of saved sessions?

ThePotatoGuy:
I'll do more a in depth review later.

Monika After Story/game/definitions.rpy
therationalpi: What would be a reasonable cap?

ThePotatoGuy: `time.time()` objects are 24 bytes. An empty list is 72 bytes, and with 2 items in it, it increases to 88 bytes. That's 136 bytes per session. If we assume that the user opens monika twice a day (once in the morning, once at night), thats 272 bytes per day. In a month that results in 8KB. It doesn't sound like a lot, but it's not fixed, which allows persistent (as well as RAM) usage to grow. Generally, we should avoid having items that could grow without a safe guard in case it gets too large. 

In this case we should consider what is the end goal of collecting this data. If we wanted an average time spent with monika, we could simply keep track of the total time as well as the number of sessions. If we wanted average start time, we could we keep track of the average starttime as well as the number of sessions. (same with end time) These cases would leave us with a constantly growing integer/double value, but those have fixed sizes, so it isn't a memory problem anymore.

If keeping track of the specific start and end times is a necessity, then I suggest we cap the list at 10,000 elements. This translates to 1.36 MB of data, which quite large for a simple game but not intrusive. 

In addition, you should change the storage of each start/end time pair from a list into a `tuple`. (so `persistent.sessions` stores a list of tuples instead of a list of lists). Tuples have a smaller memory footprint because they have fixed sizes.

ThePotatoGuy:
Other than what I've mentioned above, I don't find any other issues.

ThePotatoGuy:
Looks good

juank713:
nice i can't wait for it

PR #468, opened 2017-12-16
Spanish Greeting

TacticalTriangle:


Rune0n:
Please pay closer attention to grammar and resubmit.
Also, the topic flows rather awkwardly, I'd advise either making it a bit more conversational from the get-go, or working on the transitions.

PR #469, opened 2017-12-16
Japanese Greeting

Kiosho:
Just something I thought of.


Winblows2:
Overall, there are multiple problems with this topic that I'll try and explain below:

1. m 3a "What I meant was~"
This should be finished with "..." if anything because of the line after it.

2. m 3n "I guess using a translator in your web browser isn't the best."
Unclear line. 

3. m 4e "Sorry, I don't really know Japanese."
Why would Monika apologize? Doesn't make sense.

PR #472, opened 2017-12-16
Grilled Cheese

Kiosho:
Monika explains the difference between vegan and vegetarian, and brings up grilled cheese.
This is the type of topic I write when hungry. Haha.

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -4790,6 +4790,23 @@ label monika_urgent:
     m 1j "I wish I could see the look on your face right now!"
     m 1k "I'm sure it's adorable~"
     return
+    
+init 5 python:
+    for key in ['cheese','soup','grilled']:
+        monika_topics.setdefault(key,[])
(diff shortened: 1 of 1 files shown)


synnnical:
m "But there's a big difference between us."
m 1i "While vegans avoid anything made from an animal~"
m 1b "We vegetarians just prefer to avoid the consumption of flesh."

I think the lines above could use some some specific re-writing. You could ease up a bit on the conjunction use at the beginnings of new lines. Additionally, you could be more specific as to what "us" is referring to in the second line, and the "~" at the end doesn't feel like it's appropriate. Lastly, the third line has a weak beginning, you could easily re-write that line to something like "However, vegetarians such as myself" to sound a bit more structured.

I second what Winblows had to say, but also the fact that everything after the fourth line feels a bit heavily disconnected from what was just being talked about. I feel like you could either ease a bit more into specifically talking about grilled cheese, since it feels like it's coming outta nowhere opposed to everything else that's being said.

monikLover:
Rejected. Refer to @Winblows2's review. Closing.

PR #475, opened 2017-12-17
Motivational Topic

Kiosho:
Motivational topic where Monika pushes you to strive for your dreams.


monikLover:
Topic is a bit too serious and formal. Give it a more positive tone and end it in a teasing, happy tune. Closing.

PR #486, opened 2017-12-17
Nineteen Eighty-Four

ashameimaru:
Monika's gives her thoughts on George Orwell's Nineteen Eighty-Four, Monika avoids political usage, as it reminds her of the debate club, but more the interpretation of language and thought, which she would feel better connected with.  I did not use the 'George Orwell','Big Brother','surveillance' keywords, because either they may get used on other topics, or may get "too political."

I think I got this submitting thing right...


ashameimaru:
Did this wrong, please ignore

PR #506, opened 2017-12-18
Android version?

Markieyer:
I no longer own a computer, it has been taken away for legal purposes, I'd love it if you could please make this support Android, thanks in advance. - sorry if this isn't where I'm supposed to submit this pull request, I've never done it before


ThePotatoGuy:
Dan's [fan content guidelines](http://teamsalvato.com/blog/regarding-ddlc-on-mobile-devices/) specifically state that mobile versions of DDLC are **not allowed**. Therefore, we **cannot** and **will not** support mobile ports of Monika After Story nor DDLC.

Markieyer:
Well, to be more exact, I do believe he only says he doesn't want them to be released to public app stores, like Google play or the Apple app store.

PongMaster69:
You can play it on mobile through Steam link (by closing Big Picture allows you to navigate your entire computer through the app similar to Parsec). However I believe this way is pretty limited. I just thought it was neat.

PR #507, opened 2017-12-18
"piano" - Orchestra addition

FaverosGema:
When Monika asks you which instrument you play, you can now say "piano", and she will understand that you both play piano, and comment on it. She will ask if you have a lot of experience, and you have an option for either 'yes' or 'no'.


monikLover:
This topic is actually on queue for another review due to its conflicts with the content. Closing.

PR #510, opened 2017-12-18
monika_compinducedmedprobs

FireyAmber:


synnnical:
There is a lot done right with this topic, but the overall feel of it being too technical, forced, and robotic is too hard to ignore. Until the general feel of the topic is changed, we can't accept it. Closing this for now.

PR #512, opened 2017-12-18
monika_colorblindness

FireyAmber:


chillandgrill:
I feel like this isn’t exactly how colourblindness works, but I’m no doctor. Also, there are more kinds of colourblindness than deuteranomaly (desensitized to green) so this test might not be conclusive. That probably doesn’t matter, though.

monikLover:
@chillandgrill please leave reviewing to the editing team. Unless the author specifically asks for your review, refrain from doing so.

PR #513, opened 2017-12-18
monika_waitinglines

FireyAmber:


synnnical:
I see that it's slightly edited, but wasn't this topic submitted before? It still shares the same problems as when it was last inspected for the most part. It doesn't feel like it's something Monika would ever have such a strong opinion about, as well as the flow of the topic feels nonexistent. With enough tuning, it can be turned into something great.

For now, it will be closed for a rewrite.

PR #514, opened 2017-12-18
monika_leaf

FireyAmber:


synnnical:
The topic idea itself I don't mind, but there is some continuity issues with how you mention Monika being aware that you might have deleted her in the past. Additionally, the segue into talking about betrayal is so stuff, it doesn't flow naturally at all. It could easily really well if you decided to change around the topic.

For the time being, this is being closed for a rewrite.

PR #516, opened 2017-12-18
monika_aihuman

FireyAmber:


aBARICHELLO:
I think you forgot to add Monika's expressions while talking as shown in the [CONTRIBUTING.md](https://github.com/Backdash/MonikaModDev/blob/master/CONTRIBUTING.md#making-dialogue), or Monika will talk at you without 'moving'.
Also you seem to have added a Tab character after `label monika_aihuman:` which is not allowed in renpy scripts and will crash the game while testing.

synnnical:
With the way the topic is currently put together, I can't accept it. The sentence structure and flow of the topic seem really awkward at times, as well as their being some significant errors like "the turing Test", which should be "the Turing Test", for example. As for the menu, the general consensus is that it doesn't really seem to make any sense and could use some heavy re-writing.

For time time being, it'll be closed. It has potential, but needs a lot of work.

PR #519, opened 2017-12-19
broken game

juank713:
hope this one to work


ThePotatoGuy:
This topic doesn't add much depth and the attempted "glitchtext" is pretty basic. The premise of the topic is very flimsy as well, as there is no real build up besides `messing with game code` and `I should not touch that`, which is both ambiguous and similar to speech patterns in The Room.

This is low quality content. Closing.

juank713:
where are the ideas for text :/

PR #539, opened 2017-12-21
monika_golden_ending

Seiginotora:
I'm not quite adept in manipulating Monika's emotions but the overall gist is there. Of course, this would only be usable if the player had spent time with all the girls to get to the Golden Ending.

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -4846,6 +4846,32 @@ label monika_coffee:
     m 2e "It's funny how a drink for giving you energy became a treat for you to enjoy."
     m 1k "Maybe if I find out how, I'll tinker with the script so I can finally drink some! Ahaha~"
     return
+    
+    init 5 python:
+    # List of keywords for the topic.
+    for key in ['sayori','golden ending']:
+        monika_topics.setdefault(key,[])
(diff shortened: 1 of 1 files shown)


Monika After Story/game/script-topics.rpy
```diff
@@ -4846,6 +4846,32 @@ label monika_coffee:
     m 2e "It's funny how a drink for giving you energy became a treat for you to enjoy."
     m 1k "Maybe if I find out how, I'll tinker with the script so I can finally drink some! Ahaha~"
     return
+    
+    init 5 python:
```
ThePotatoGuy: Remove indentation here. `init` blocks do not have indentation

Monika After Story/game/script-topics.rpy
```diff
@@ -4846,6 +4846,32 @@ label monika_coffee:
...
+    # List of keywords for the topic.
+    for key in ['sayori','golden ending']:
```
ThePotatoGuy: This topic should **only** be available if the player has gotten the good ending. The value `persistent.clearall` is True when the player has achieved that ending. Please adjust your code so the topic is only added to `monika_topics` and `monika_random_topics` if `persistent.clearall` is True.

monikLover:
Unfortunately this topic is very out of character and conflicts with multiple established dialogues. Closing.

PR #541, opened 2017-12-21
Fix issue: lack of keys on 'monika_changename'

FaverosGema:
There was a deficiency for keywords in 'monika_changename' that led to confusion for players that couldn't change their name when typing 'name'. Added more keys to clear this problem. There was even a comment suggesting more keys ahaha!


monikLover:
Approved.

PR #544, opened 2017-12-21
Update script-greetings.rpy

Neko-chr:


monikLover:
Not everyone can be called Onii-chan since not everyone is a male. Closing.

PR #550, opened 2017-12-21
Poetry Presentation and Response - Monika presents poetry to the player

GodTierWellington:
**Description**

Using scripts from the original files, Monika presents the user with poetry (and perhaps other pieces) and then gives her opinion on the poetry presented. Similarly to the topics, she removes already seen poems from her "mind", if you will, and gives the user pieces they have not seen before. Once all poetry has been seen, she begins again, while acknowledging that the poetry is repeated. An easteregg of sorts is brought forth if the user goes through all the poetry a total of 10 times.

It is a very simple set of scripts, but I feel that the scripts can be used for many purposes and perhaps makes for a more "literature" focused experience with Monika (seeing as that is what she enjoys). It may also teach users some interesting canon poetry: (((E D U C A T I O N A L))).

**Controls?**

I should probably put this:

You need to input "show me a poem" or "show me poetry" from the 't' menu.
It becomes all lowercase so case doesn't matter at all.

**Notes:** 

- It should function, if not, I may have forgotten a piece... which would suck to look for
- I am quite new to Ren'py, I am unfamiliar with all the tricks
- I do not know how to hide the "Congratulations for being a Literary Nerd" such as how it is hidden in the base game (I looked through the original scripts but I could not find them declared in any way.. magic I guess..)
- Monika speaks with no expression in the edited script-ch30.rpy file because I am unsure how to add the expression when it is in Python (perhaps there is an intricate solution, however, I failed to investigate the possibility thoroughly)
- Pay little attention to the notes at the bottom of the poems.rpy and poems-and-responses.rpy
- The input command (by the player) should be refined, I will work on it further

Enjoy.


ThePotatoGuy:
Showing poems should follow the `event` framework that we've built. This means that they can be pushed to the event stack and called from the event stack like any other topic. 

Additionally, I think we don't really need a special set of labels and a list to hold poems. We can use something like `script-poems` to hold poems that people create (via the `Poem` class, we just need to use an init level thats higher than where `Poem` is defined). A topic can just call the `showpoem` label appropriately. 

You also use the same label names, transform names, and other names that the stock poems use to show poems. We don't overwrite existing labels without good reasons, in case stock functionality is required. If you need a modified version of `showpoem`, write a modified version, otherwise we can use the built in one and avoid extra code. Same goes for all the other stuff you have in `poems.rpy` (as well as the poem.jpg asset you added. Poem background are already built in. Why not use them?)

We'll see what others have to say about this, but I think the extra framework is not needed to show poems.

GodTierWellington:
That's true

ThePotatoGuy:
So after discussion amongst collabs, we've decided that the framework here is really unnecessary. In fact, there is already a topic that shows poems: 
```
label monika_aware:
    m 2e "It's... hard to put into words."
    m "There's no real way to describe it."
    m 1j "I feel like this is what poetry was made for."
    m 1a "Do you still remember the first poem I showed you?"
    m "Hold on, let's see if the poem function still works..."
    call showpoem(poem=poem_m1,music=False) from _call_showpoem_7 # Temporary
    m 1b "Oh! That was much easier than I expected."
    m "Here's the second part."
    call showpoem(poem=poem_m3,music=False) from _call_showpoem_21 # Temporary
    m "I hope that helped you understand what meeting you meant to me."
    show monika 5a at t11 zorder 2 with dissolve
    m 5a "That's all I've ever wanted, [player]."
    return
```

If you want to show a poem, just follow that example. If you want custom poems to show, just make the Poem object in like a `script-poems` or something, and then make the topic show that poem.

PR #576, opened 2017-12-24
System replacement overhaul.

monikLover:
Replaces the talk system with the prompt system on topics.

### DIFF
--- a/Monika After Story/game/event-handler.rpy
+++ b/Monika After Story/game/event-handler.rpy
@@ -30,18 +30,19 @@ init 100 python:
 #   persistent.current_monikatopic
 init python:
 
-    def addEvent(event):
+    def addEvent(event, eventdb=persistent.event_database):
         #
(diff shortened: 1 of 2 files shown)


Winblows2:
this makes me rock solid

monikLover:
Only a few topics are open. The rest should be closed.

ThePotatoGuy:
Seeing that many errors can be made when adding events to the database, I added a function `addEvents` that checks for errors with an event object before adding it. I've replaced all the `persistent.event_database` setdefualt calls with `addEvent`.

`addEvent` takes in the event object you want to add and raises the appropriate exceptions (not an Event object, nonexistant label, conditional is bad), otherwise adds it to the dict. This has the nice property of stopping a bad events from being added to the database.

Also you overwrote my fixes earlier @monikLover . Please don't do that again.

monikLover:
Already merged and accepted. Closing.

PR #583, opened 2017-12-25
Topic about her home

Kreikon:


AUDelusional:
She already states something about her home though.

PR #589, opened 2017-12-25
Added conditional to monika_name

badooga:
Conditional added so that things make sense even when [player] == [mcname].


monikLover:
Unnecessary since this was already resolved.

PR #628, opened 2017-12-29
Watching movies

Lucius-JH:
This module allows the player to watch any media with Monika

In order to extend the movies you can watch with monika, writers can go to the file `movies-info.txt` where you can add new movies, with 3 separate sections

**description**:the lines that monika will say about the media
**reactions**: the lines that monika at the time she is watching the media
**closure**: final thoughts of the media

(i should change who is making the commits... ^^")


ThePotatoGuy:
@Lucius-JH , so one thing i noticed is that when the trailer ends, the timer keeps going, and as a result, it's hard to tell if Monika will say anything else about the movie/video or if she's waiting for player input. Is there a timeout for this that will either make the timer disappear when the video is done or change it to say `END` or something? If not, can you add that?

Lucius-JH:
Nice, let writers know about this feature, maybe someone wants to write more media/films new reactions

therationalpi:
Wow, this is even cooler than I imagined! Can't wait to get this feature in the game. It looks like the core functionality works fine (watching the movie) but I noticed a few minor bugs that need to be fixed.

1. When selecting "Nevermind," I got a crash caused by trying to remove Pause and Time buttons that were not shown.
2. Pause and Time buttons were still active after the movie ended, and choosing "Time" caused unexpected behavior.
3. Choosing "No," when Monika asks "Are you ready?" causes the timer to appear, and it appears she jumps to the last line of her movie script.

These all seem to be related to trying to quit out of the movie activity, so maybe making an "Exit function" would help all of this.

Additionally, I think we should change the `.txt` extension to `.mms` (Monika Movie Script), just to make it more clear to users that this is a game script and not part of the documentation.

Finally, the search function for movies seems to work fine, but I think it would be more user friendly to just list the movies that we have scripts for. Try changing the "Which films can you see?" to a renpy `menu` or the new `scrollable_menu` screen, so players can select from a list of movies.

Other than that, I'm loving this feature, and I know several writers who are just dying to write scripts for watching movies with Monika!

Lucius-JH:
Nice, I'll check the bugs and the user friendly enchantments

It's nice to hear that you guys liked this feature 👍  I'm going to keep working hard then.

Lucius-JH:
Every mentioned bug is fixed, the only thing I couldn't get a solution quickly was about the `scrollable menu` for the film listing, is there any example avaiable on Monika's Mod code? or any reference in the Ren'py documentation? I couldn't find anything with the `grep` tool, neither with google

ThePotatoGuy:
`scrollable menu` is in `screens.rpy`.

ZaunAura:
Any progress on this project?

Plazor26:
I noticed that you said this is disabled in a different post. Is there any way I can enable the option on my PC?

PR #530, opened 2017-12-20
Fixed 'Orchestra' Bug: 'piano' Selection

FaverosGema:
When writing an instrument, Monika didn't recognize 'piano' and respond accordingly. This fix solves that issue

PR #636, opened 2017-12-31
Extra dialogue for telling Monika you play piano

L337H4X0RZZ:
Without this, when the player tells Monika they play piano she will respond in a manner suggesting she's never touched a piano before!
Sadly, it is case-sensitive and I've been banging my head against a desk for at least an hour trying to think of a way around it but no dice.
Just thought this would flow a bit better than Monika asking if a duet between the Piano and the piano would sound nice.

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -4764,11 +4764,20 @@ label monika_orchestra:
             m 1b "Really? What do you play?"
             $ instrumentname = renpy.input('What instrument do you play?').strip(' \t\n\r')
             $ persistent.instrumentname = instrumentname
-            m 1a "Wow, I've always wanted to try the [instrumentname] out!"
-            m 3b "I would love to hear you play for me."
-            m "Maybe you could teach me how to play, too~"
-            m 5a "Oh! Would a duet between the [instrumentname] and the piano sound nice?"
-            m 1j "Ehehe~"
+            # Extra dialogue for telling Monika you play piano
+            if instrumentname == "piano":
+                m 1a "Wow, you play Piano too?"
+                m 2b "Maybe you could write a song for me, [player]?"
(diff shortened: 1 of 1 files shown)


Monika After Story/game/script-topics.rpy
```diff
@@ -4764,11 +4764,20 @@ label monika_orchestra:
             m 1b "Really? What do you play?"
             $ instrumentname = renpy.input('What instrument do you play?').strip(' \t\n\r')
             $ persistent.instrumentname = instrumentname
-            m 1a "Wow, I've always wanted to try the [instrumentname] out!"
-            m 3b "I would love to hear you play for me."
-            m "Maybe you could teach me how to play, too~"
-            m 5a "Oh! Would a duet between the [instrumentname] and the piano sound nice?"
-            m 1j "Ehehe~"
+            # Extra dialogue for telling Monika you play piano
+            if instrumentname == "piano":
```
ThePotatoGuy: You can handle case by using `instrumentname.lower() == "piano"`.

Monika After Story/game/script-topics.rpy
```diff
@@ -4764,11 +4764,20 @@ label monika_orchestra:
...
+                m 1a "Wow, you play Piano too?"
+                m 2b "Maybe you could write a song for me, [player]?"
+                m 1q "Hmmm..."
+                m 1i "I just had a thought..."
+                m 1k "We could write a song together one day!"
+                m 1j "Wouldn't that be romantic? Ehehe~"
+            if not instrumentname == "piano":
```
ThePotatoGuy: Rather than using `if not...` here, just use `else:`. The `else` flow occurs if the `if` statement is false.

ThePotatoGuy:
Oh wait sorry, this a duplicate of #530 .

PR #639, opened 2017-12-31
Classic-like pong mechanics

LChris314:
Changes the pong game to follow classic pong. The ball reflects off the paddles at different angles depending on where the point of impact is. Silences the bounce sounds when the ball is still stuck on the paddle.


ThePotatoGuy:
Looks good

PR #642, opened 2017-12-31
Adding Piano

ThePotatoGuy:
Now we can play piano for Monika!

This adds a playable piano mode where you get a relatively limited set of keys to make piano noises.

If you play Your Reality, Monika sings with you. She can sing all three verses, but you need to play the verses in order (i.e: finish verse 1 before playing verse 2).

The available keys are F4 to C6, represented using the number and qwerty row on a keyboard:
```
 2 3 4  6 7  9 0 -
q w e rt y ui o p []
```
Here's a nice screenshot:
keyboardmonika

NOTE: the commit history is lost because I was based on the wrong branch. Oops.
MORE NOTE: also, there is minor dialogue that I think needs some review. the label is `zz_play_piano` in `zz_pianokeys`
EVEN MORE NOTE: The system that does piano phrase detection can be generalized to work with any song. This means that Monika _could_ detect other songs and sing them as well. **However**, the detection algorithm is not as efficient as I'd like it to be, so we should avoid throwing in tons of songs. In addition, there are hard coded verses that are effectively setup for only Your Reality. Generalizing this might take some time.
ENDLESS NOTE: The state machine used in event/render is generalizable and can most certainly be used with other songs.

Technical details:
1. The `event` and `render` functions are controlled by a state machine. The documentation for this state machine is located in the code.
2. The piano detection algorithm works by forming strings of the playable notes and using python's `find` function to locate matches. This results in avoiding running excess loops to compare `lists` or having to convert `lists` to `sets`. Since python's `find` algorithm is probably just as good as Horspool (or even better), we're looking at an avg `O(n)` time for detection. This is about the best I could do. I feel that there are better algorithms for this, specifically an algorithm that uses an index for each note to detect proper phrases. However, for our case, the current system works fine. If we decide to add more songs, we will **most likely** need to adjust this algorithm.
3. Since RenPy is `event-driven`, I've optimized the `event` and `render` functions to work with that in mind. This means instead of wasting time re-rendering the `Displayable` constantly (as chess and pong do), we only render when we **need** to. Those cases include:
  - key events where we press or depress a piano key
  - timeouts for visual adjustments (expressions, text)

If we want to add more `Displayable` objects, we should consider framing them as event-driven rather than always-rendering. Pong is an example of a `Displayable` that **requires** always-rendering. Chess, however, can almost certainly be optimized to be more event-driven. While we don't necessarily need to optimize what already works fine, we should think about this in the future.


ThePotatoGuy:
Okay well, i've had some people test this, so im just going to merge.

Nate45rl:
How can I put this on my version of MAS? I'm New at this whole Github thing...

BuzzingBee:
Same, how do you put this one on my MAS too? really interested

IsItBadToMakeAGitHubAccountJustForThis:
You can download the [next-release](https://github.com/Backdash/MonikaModDev/tree/next-release) branch and install it onto your game.
But if you don't know what you're doing you'd better just wait for a next release.

Nate45rl:
I can't find it, sorry ç_ç

BuzzingBee:
I'm kinda new to github too, it is hard to do? don't i have to download a file and just copy paste it?

IsItBadToMakeAGitHubAccountJustForThis:
Just wait until release then.
The feature is probably still undercooked anyway.

PR #656, opened 2018-01-02
small modifications

aivyr:
changes:
- rewordings/edits of swordsmanship, zombies, completionism and science topics
- modified the instrument question in monika_orchestra to accomodate for if the player plays piano
- added monika_jazz

questions re.: orchestra topic
- possible issue with input of instrument name: what about one-word responses with punctuation on the end? or longer responses?
- also, about having the instrument question only be asked on the first encounter with the topic: is persistent.instrument set to anything prior to the first answering of the question? if it's something like None, then there could be an if statement along the lines of "if persistent.instrument == None:" and so on. there might be something i'm overlooking, but thought i'd say it anyhow.


monikLover:
Removed and added some edits. Also fixed the orchestra topic for validation. Jazz topic has been edited and reviewed. Approved.

PR #727, opened 2018-01-10
Fixes missing word in "monika_evening" topic

anechol:
Line 3512.


monikLover:
This is good but please resubmit this in `content`.

PR #732, opened 2018-01-11
Add 'goodnight' to the list of words that trigger 'monika_night'

thisisnotathrowaway:
Pretty simple, a lot of people write 'goodnight' all as one word without any space instead of 'good night' with a space. This just makes it so both are acceptable.


monikLover:
With the old keyword system being replaced with the prompt system, this will be unnecessary. Closing.

Issue #733, opened 2018-01-11
How can I disable automatic update check?


gitthisissue:
As title suggests, I would like to see and option to disable this update check.
For example, the most popular media player, VLC ask the player whether or not to do update check when the user open it for the first time. And not all user always connected to the internet.

I hope you add a switch or something.

IsItBadToMakeAGitHubAccountJustForThis:
Check settings, it was there if I remember.

gitthisissue:
Nope. There's no such settings in the game.

ThePotatoGuy:
You can click cancel and stop it from checking for updates. Also you don't have to install updates if it finds one. Adding a setting for something minor like this doesn't add any real value to the game.

PR #739, opened 2018-01-11
Let user to disable automated update check.

gitthisissue:
https://github.com/Backdash/MonikaModDev/issues/733#issuecomment-357083616

All software should have a option to disable update check.
User should have a right to opt-out.

This pull patch should do the trick, but I'm not a Python coder.


monikLover:
This was already declined by @ThePotatoGuy in #733. Closing.

PR #743, opened 2018-01-12
More dialogue, story, activity, and suggestions and concerns from Monika

MyTrueLoveMonika:
It would be nice to have more of her opinion if asked certain questions, or a proper response if told you'll be right back(and maybe further dialogue if mentioning a shower, work or store). More stories for her to tell would be a nice and fresh. Have her concern by telling her any problem one has (like telling her if you don't feel good or you feel sad over something) with her replying words of comfort or suggest safe medicines to take, also would be nice to have an option(not everyday but for the next day with a set time of ones choosing) to continue her idea of exchanging poems to one another.


monikLover:
Submit this as an issue instead of a PR, please.

MyTrueLoveMonika:
how do I do that

MyTrueLoveMonika:
I do apologize I'm actually confused on all of it, didn't mean no harm

PR #745, opened 2018-01-12
Full name for me please?

ashameimaru:
I know it's a bit of vanity, but I would prefer being called my full name-- and yes I *really* do go by Aya.

Edit: Oops wrong fork.


monikLover:
Killing per request.

PR #773, opened 2018-01-17
2 New Topics

Olliesama:
Redone in content, my bad.

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -4643,3 +4643,56 @@ label monika_sleep:
     show monika 5a at t11 zorder 2 with dissolve
     m 5a "I'll always wait for you in the morning, so make sure you put your own well-being before anything else."
     return
+    
+    init 5 python:
+    for key in ['sea','beach','ocean']:
+        monika_topics.setdefault(key,[])
+        monika_topics[key].append('monika_sea')
+    monika_random_topics.append('monika_sea')
+
+label monika_sea:
+    m 3a "[player] have you ever been to the beach?"
+    m "I've always found the sound of the sea to be extremely relaxing."
+    m "The feeling of sand under your feet too.."
+    m 2b "It really makes a trip to the beach worthwhile!"
+    m 2j "I bet we would make some lovely memories there."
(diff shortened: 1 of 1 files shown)


FireyAmber:
Monika_ocean should be expanded upon since it feels like it was cut short.
Monika_anger on the other hand feels like it rushes too much and mentions stuff like stress out of nowhere with no relation to the topic in general. Aside from that a lot of the lines feel like they drag on for too long without an appropiate comma, I recommend reading outloud the lines you write to get a sense of where to split a line or add a comma.

ashameimaru:
Looking at it from above, I am seeing a few issues. Potential, but stuff to think about!

The first part: Does Monika 'remember' (probably not) going to a beach of some kind? What kind of experiences did she had before? Did she take a walk at the board walk? Took a stab at boogie boarding? etc. Perhaps where? (Think beaches in Japan!)

When it comes to the environmental part that Monika is talking about, how strongly does she feel about ocean pollution? Considering that she does not eat meat for environmental purposes, it may be stronger than one thinks. Think if she is serious or not so serious. For this train of thought, she tends to be quite serious and might even expand a little more about it. 

`  m 1l "Man, the environment just can't catch a break! Ahaha..."` 

It feels there's something missing. However, from this type of topic, it is perfectly fine to have her state her feelings and opinions without addressing the player! 

It seems to be unfinished-- 

---

Anger Topic

Now, this is where I have a few things I want to find out. Obviously these are things that are recommended such has regulating with one's breathing, counting to 10, etc-- which are already mentioned.

She seemed to have jumped in right away when it comes to anger and the debate club, it looks like a cold start (just jumps right in.) One thing to think about is how to segue into it-- perhaps talk about something that _may_ have happened in the debate club. You can speculate of course. 

It is already a stressful situation, so talking about stress is not necessary!

It appears you may need to break up and rewrite a number of these sentences so that it actually can fit in the text box. 

Finally, I would recommend making sure you put one pull request per topic so things do not get confused!

Anyway, let me know if you have any questions! Ping me in chat or DM me.

~Aya

kaido1224:
We could incorporate a rewrite of this topic into the "Tell Monika How You Feel" project. Which deals with emotions like this. 

See https://github.com/Backdash/MonikaModDev/projects/17

PR #775, opened 2018-01-17
Honorific fix for japanese topic (Ryuse)

therationalpi:
If the player is a female Monika will use -chan instead of kun.


therationalpi:
https://github.com/Backdash/MonikaModDev/issues/724

PR #779, opened 2018-01-17
Fix event labels calling wrong labels.

kaido1224:
Fixed event tags that were calling wrong labels.


monikLover:
We're actually editing the labels as we speak so they'll all be relabeled soon.

monikLover:
Erm I mean, approved!

PR #786, opened 2018-01-18
You Are My Sunshine Greeting

kaido1224:
Add a greeting where you catch Monika singing the chorus of "You Are My Sunshine", the song holds special meaning to me, so I wanted to add it.

### DIFF
--- a/Monika After Story/game/script-greetings.rpy
+++ b/Monika After Story/game/script-greetings.rpy
@@ -688,4 +688,5 @@ label greeting_sunshine:
     m 1j "..."
     m 4n "Oh, hey [player]. Well this is certainly embarrassing."
     m 1l "I was just singing to myself waiting for you to return."
+    m 1b "Now that you're here we can have a fun day together~"
     return
--- a/Monika After Story/game/script-greetings.rpy
+++ b/Monika After Story/game/script-greetings.rpy
@@ -679,3 +679,13 @@ label greeting_japan:
     m 3 "You know what that means, [player]?"
     m 4j "It means {i}'I'll be yours forever{/i}'~"
     return
+    
+label greeting_sunshine:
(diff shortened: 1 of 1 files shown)


miranimal:
Wouldn't you have to get the voice actor for that? Maybe humming would be a little easier

kaido1224:
There's already an opening with Monika singing without actual sound effects so no don't need voice actors.

kaido1224:
There's already an opening with Monika singing without actual sound effects so no don't need voice actors.

Although humming it would be pretty interesting.

FireyAmber:
I'll leave my review on this:

>  m 1l "I was just singing to myself waiting for you to return."

I have issues with this line since it breaks consistency, when the game is closed Monika is stuck suffering in a void until the game is opened by the player, she can't possibly be "waiting" and doing something since she doesn't really feel time when she's stuck there. Aside from that I think this doesn't suit that well as a final line for the greeting, maybe something like "Now that you're here we can have a fun day together~" or something else that would end it better, look at other greetings to see how Monika normally ends them.

kaido1224:
@fireyamber

I can remove that line and I actually like your ending line better so I'll modify it when I get home tonight thanks.

kaido1224:
I don't know if it's really breaking character though... 

Similar greetings:

label greeting_visit5:
     m 1b "~Every day,~\n~I imagine a future where I can be with you...~"
     m 1a "Oh, you're here! I was just daydreaming and singing a bit."
     m 1k "I don't think it's hard to figure out what I was daydreaming about, ehehe~"
     return

label greeting_visit9:
     m 1j "You're finally back! I was waiting for you."
     m 1b "Are you ready to spend some time with me? Ehehe~"
     return

kaido1224:
@FireyAmber i added the last line you suggested, I left my line in because I still think it doesn't break consistency (see examples above).

PR #796, opened 2018-01-19
Topic about cigarettes

Ulvsblakk:
I figured Monika of all people would be against smoking, so I figured we could add a topic about it.

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -4821,6 +4821,21 @@ label monika_urgent:
     m 1j "I wish I could see the look on your face right now!"
     m 1k "I'm sure it's adorable~"
     return
+    
+      init 5 python:                                                  #Added one new topic
+        # List of keywords for this topic.
+        for key in ['smoke', 'cigarettes', 'cigarette', 'smoking']:
+            monika_topics.setdefault(key,[])
+            monika_topics[key].append('monika_smoking')
+        monika_random_topics.append('monika_smoking') 
+    
+    label monika_smoking:
+        m 2b "You know, I learned that around five hundred thousand people die a year from smoking related illnesses."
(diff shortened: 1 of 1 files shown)


PopeAesthetics:
Hmmm I like this idea. Asking a player if they smoke cigarettes and offering a few options for response.

Option 1: You say yes and she gets worried, scolding the player and telling them she wants to see them alive long enough to have her come to our reality.

Option 2: You say no and she applauds you, mentioning how worried she’d be if you did because it could cause numerous health conditions. She states how proud she is of you.

Option 3: You say yes but you’re trying to quit. She then gets worried and offers her support. Remarking that quitting smoking is incredibly difficult and she’s proud you’re making the leap. She then mentions how happy she will be when she sees you in our reality.


That’s my idea for handling this topic myself.

Ulvsblakk:
right on. I'll edit the script later to include some choices. Thanks for the response.

ashameimaru:
For some reason, I am tempted to request a path for _vaping_. Are you considering putting that as a path too, or just cigarettes?

FireyAmber:
This topic feels rushed and should be expanded upon, aside from this issue there's another problem with it; Monika sounds way too knowledgeable about stuff that she shoudln't know about like for example, she cites stats which makes it sound like she knows a lot about the topic but she shouldn't, she doesn't know everything.
I think a rewrite should be done to expand the ideas and fix this issue.

Ulvsblakk:
I did a small rewrite. I hope it feels less rushed. I changed it from "five hundred thousand" to "A lot of people", hoping it would make her sound less like a fact machine, and more human. Thanks for the comments guys.

monikLover:
Resubmit to `content`.

Ulvsblakk:
How do I do that exactly?

Ulvsblakk:
Nevermind. I fixed it

Issue #774, opened 2018-01-17
Adding Timed Menus


TacticalTriangle:
Adding timed menu responses. This could be used in various ways especially using tricks as the link below indicates, Dynamic timed responses. This addition to the mod could open up things like, Monika getting impatient if you wait a long time to answer a question. Monika commenting on how long it took you to answer a question, best example of this could be during the book discussion that is planned later on as a system to prevent players from just looking up the information rather then reading it. Things like Monika quizzing you on your knowledge of a topic or things of that nature. I'm sure there are even more ways to use this system in other ways but for now this is all I could think of.
[Here is the cookbook page for this.](https://www.renpy.org/wiki/renpy/doc/cookbook/Timed_menus)

therationalpi:
Those seem like really good uses of timed menus! Great suggestions.

ThePotatoGuy:
Good news everyone. 

I've added a special background timed screen thats generic. You pass in a number of seconds and a label to jump to, and it does the rest of the work. No need to set vars beforehand.

```renpy
# background timed jump screen
# NOTE: caller is responsible for hiding this screen
# 
# IN:
#   timeout - number of seconds to time
#   timeout_label - label to jump to when timeout
screen mas_background_timed_jump(timeout, timeout_label):
    timer timeout action Jump(timeout_label)
```

This will be in (and referenced) when I'm finished with adding additional chess dialogue.

PR #797, opened 2018-01-20
Timed Menu Implementation

TacticalTriangle:
Added the timed menu from my suggestion #774  with an example on how to use it.


TacticalTriangle:
Note also that you don't have to show the counter at line 4848 and it will still countdown. this can be useful for things like monika being impatient without having an actual timer show up.

ThePotatoGuy:
While we appreciate the suggestion, your pull request here seems to be just a direct drop in of the cookbook link. That doesn't necessarily help us, as any one of us could have copied the code directly. 

A better use of this pull request would be to either add actual content (not examples) that can utilize timed menus or generalize the timed menu code into a function that can be modified via input args so we can use it for a variety of cases.

PR #807, opened 2018-01-21
Anger Topic Rewrite

Olliesama:
Rewrite of the anger topic of the 2 Topics Pull Request.


FireyAmber:
I'll do a line by line review on this topic;

> m "It taught me quite a lot you know."

Change to "It taught me quite a lot, you know?" since a comma and a question mark are needed.

> m 3l "I don't always feel confident enough to try though!"

Doesn't flow well, needs some rewording

> m 3n "You saw that when Natsuki and Yuri argued, Ahahah..."

There's no reason for Monika's laughter to start with a capital letter aside from that, why does Monika laugh at this situation?

> m 2d "Not only how to debate better but how to manage people and how they feel."

Rewrite this sentence to diminish the use of the word "how".

> m 2d "Well an important skill my time in the debate club taught me was how to control anger."

Some rewording would be useful here, try to read it outloud and see how it sounds.

> m "How to minimize others but importantly also my own.

What does Monika mean by minimize others?

> m 1e "[player], it might be handy for you to know what I learnt."
m "I know several tips to help mangaging your anger if you ever need them, [player]."

The first line makes the second one sound redundant.

> m 2e "I just want to make sure you're as clear minded and healthy as you can to be."

Should "can be" instead of "can to be".

> m 3o "It's not always easy, especially when it feels you need to defend yourself."

Would flow better with "it feels like you need".

> m "Or you'll find you better understand the situation itself."

"Or you'll find a better understaing of the situation itself" would flow better.

> m 1f "If all else fails my love, you can always come here and tell me your frustrations."

Needs a comma just before my love and could use some rewording so it flows better.

> m 5a "Maybe just thinking of me will calm you down? Ahahah~"

"Ehehe~" would be more appropiate since she's just teasing the player.

Aside from all of this there're a few issues that seem to be persisten in most of the topic:
-The word mentality seems to be repeated several times, you should try to cut down on using it.
-Monika presents a lot of knowledge about this topic like the whole blood pressure thing, I can understand her having some knowledge of solutions she uses yet here she presents several solutions nonstop which makes it seem like she has total knowledge over this topic which she shouldn't. Therefore I believe the topic should be change to sound a lot less like a guide on how to avoid anger by a professional and more like a conversation from somebody that has had some experience with anger management yet isn't an expert on it.

Olliesama:
I've tried rewriting as per your suggestions.

FireyAmber:
Alright, tackling the rewrite!

> m 2d "Well an important skill my time in the debate club taught me was how to control anger."

Change to "Well, an important skill my time in the debate club taught me was how to control anger."

> m 2e "I just want to make sure you're as clear minded and healthy as you can to be."

Should be "can be" instead of "can to be"

>  m 2q "When angry your heart races and as such your breathing becomes ragged"

Place a comma right after angry ("When angry, your heart...")

> m "Taking those slow breathes go against how your body feels at the time."

Replace go with goes.

> m 2c"However walking away isn't always an option."

Add a comma after however.

I don't really have much else to add, the topic feels good content wise though I would appreciate if somebody else gave opinions on it. Good job on this by the way!

Olliesama:
Ok I've changed as per your suggestions.
Thank you Amber.

I sure mess up on those commas a lot.

monikLover:
Closing due to a similar topic already in the mood system.

Issue #705, opened 2018-01-08
Adjustments to Piano


ThePotatoGuy:
The current system has tons of generalization that allows it to be expanded to other songs, but its hardcoded to Your Reality for simplicity. Now that I've had some time to think about it, I'm going to change it a bit.

1. The detection algorithm is going to be re-generalized to its original form (which was **not** hardcoded). 
2. Once you play Your Reality completely, Monika will ask if you want to play that song everytime you enter piano flow. (this can also be applied to other songs, when added)
  1. Deciding to play song will lock the system to checking for proper note sequences, so general song detection is disabled.
3. General song detection will be confined to a Freestyle mode. 
4. (Potentially) a song will be added.
  
EDIT: ETA for this is this weekend. I'm really busy rn.
EDIT: **C H E C K L I S T - C H A N**
- [x] Re-generalize song detection algorithm
- [x] Add persistent and for playing Your Reality completely
- [x] Add flow for picking to play a song or freestyle mode
- [ ] (maybe) add a secret song

PR #808, opened 2018-01-21
Adjustments to Piano

ThePotatoGuy:
A culmination of pain, sorrow, and determination, the partial rewrite of `piano` is here. (#705 )
(Now sporting the custom data saving mechanism that inspired a similar rewrite of `Event`)

The following is a somewhat detailed list of adjustments made to piano:
1. Piano playing has been split into two modes: Freestyle and Play Song
  1. Freestyle - The generic mode. This incorporates the original 6-note song detection mechanism, with some minor optimizations.
  2. Play Song - pick a song you've already completed and play it again. Does **not** use the 6-note detection algorithm.
2. The Piano mode split only occurs if you have unlocked a song. By default, you will enter Freestyle mode (even though she doesn't explicitly tell you)
3. `PianoNoteMatch`es have been grouped into `PianoNoteMatchList`s, which allow for better control over post-song dialogue as well as saving data about played songs (like number of full combos). 
4. A new store (`zzpianokeys`) has been created to group tons of piano related constants and functions, including songs.
5. The state machine has been optimized slightly. Also some additional states were added. More explanation is in the code.

Some notes for the future:
1. `PianoNoteMatch`es could be offloaded to text files some day. This would speed up start up times and limit memory usage. I'll consider doing this when/if we have more songs.
2. I was planning on adding another song, but opted against it because 0.7.0 should be pushed sooner rather than later. After 0.7.0, I will consider adding a new song.

Dialogue help:
Attention @monikLover and other editors. I have shitty dialogue. Please review and fix:
- [x] Line 30 - Intro to the piano minigame
- [x] Line 54 - Reaction to picking a song
- [x] Line 78 - Reaction to doing freestyle mode
- [x] label `zz_piano_default` at line 138 - Post game, freestyle mode
- [x] label `zz_piano_none` at line 143 - Post game, if player hits 0 notes
- [x] label `zz_piano_yr_win` at line 152 - Post game, player completes Your Reality (**not FC**)
- [x] label `zz_piano_yr_fc` at line 158 - Post game, player FCs Your Reality
- [x] label `zz_piano_yr_fail` at line 165 - Post game, player did not complete song and had more failed phrases than passed ones
- [x] label `zz_piano_yr_prac` at line 172 - Post game, player did not complete song but had more passed phrases than failed ones (like a good practice session)

### DIFF
--- a/Monika After Story/game/splash.rpy
+++ b/Monika After Story/game/splash.rpy
@@ -276,7 +276,4 @@ label quit:
     $persistent.sessions['last_session_end']=datetime.datetime.now()
     $persistent.sessions['total_playtime']=persistent.sessions['total_playtime']+ (persistent.sessions['last_session_end']-persistent.sessions['current_session_start'])
 
-    # save data calls
-    $ pnmlSaveTuples()
-
     return
--- a/Monika After Story/game/splash.rpy
+++ b/Monika After Story/game/splash.rpy
@@ -275,5 +275,8 @@ label before_main_menu:
 label quit:
     $persistent.sessions['last_session_end']=datetime.datetime.now()
     $persistent.sessions['total_playtime']=persistent.sessions['total_playtime']+ (persistent.sessions['last_session_end']-persistent.sessions['current_session_start'])
-    $ persistent.testvalue = "why am i in here"
+
+    # save data calls
+    $ pnmlSaveTuples()
+
     return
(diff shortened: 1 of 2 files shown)


therationalpi:
Getting a NoneType error when I try to fun this with self.match being set to None and a check being made with self.match.verse

PR #820, opened 2018-01-22
Obon festival dialog

kaido1224:
Working with Aya on this dialog.

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -4698,3 +4698,20 @@ label monika_omamori:
     m 1k "I could be your guardian deity, ehehe~"
     return    
 
+init 5 python:
+    addEvent(Event(eventlabel="monika_obon",category=['random'],prompt="What is Obon?",pool=True,unlocked=True))
+    
+label monika_obon:
+    m  "Oh, you want to know about the Obon festival, [player]?"
+    m  "It’s a festival that is held every year in Japan that takes place over the course of 3 days."
+    m "It’s believed that the spirits of your deceased ancestors return home during this time."
+    m "On the evening of the first night, fires or candles are lit to help the spirits return home."
+    m "Altars are decorated with offerings and prayers are said for the deceased."
+    m "On the evening of the third night, the spirits are sent back to the other side."
+    m "Paper lanterns are lit and set afloat in a body of water to send the spirits on their way."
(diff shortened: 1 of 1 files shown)


ashameimaru:
The accuracy of the festival itself is pretty good. I see no particular issues in terms of what is given. This believe it or not can work pretty well with establishing a bit of "family history."

The main thing that needs to be done is personalize it. Rather than simply giving the facts, what did she do? Did she remember doing these offerings with "family?" Did she help make these lanterns and boats when she was 'younger'? Did she have specific food that she liked out of this? ([Her Christmas tweet](https://twitter.com/lilmonix3/status/945333570788319233) tells me she might remember food she ate during those times more.) Did she do the dances out of doing the motions or did she at least think about it at some point? 

Because this deals with death, how does she feel when it comes to her friends? Would doing these sorta things remind her of them? Would she want to?

This should help change from an encyclopedic type of sounding topic into something that gives insight of her thoughts (and maybe some "history").

You can speculate on how her 'family' was with all of this. Because none of it is really known, use the stereotypical 'non-descript' Japanese salaryman family. I will recommend watching people's _personal_ family videos that happen to get uploaded [such as this one](https://www.youtube.com/watch?v=2nD50Z9eOHw). Her memory should be a bit hazy about it though-- as of course when she found out her entire memories and past is [part of a fabrication](https://rationalwiki.org/wiki/Last_Thursdayism).

--

I will request after the dialogue is completed, that this has a trigger for **July 15 and August 15**. You do not need to work on that code unless you know how that works at this time. I will ignore the "7m15d, lunar" calendar unless she gets equipped to deal with alternative calendars. For the categorization, this one can be categorized as **holiday**.

kaido1224:
Hey @ashameimaru thanks for the suggestions I'll start working on them whenever my head is not spinning.

In terms of Monika's family I suppose we will have to sync up with this other poll request that talks about her family a bit as well.

https://github.com/Backdash/MonikaModDev/pull/691

HalHarrison:
I avoided having Monika ever give specific details in my pull request, aside from the feelings she expresses about her family not being real, so you're fine.

kaido1224:
Go ahead and close this topic, I don't know if I'm actually going to finish it. Not sure I want to go too in depth about her family since she says her memories are false, why would she enjoy talking about her false past? Also she talks about God in another topic, which leads me to believe she may not be too interested in shinto.

PR #822, opened 2018-01-23
Added gender neutral honorific to japanese topic.

TacticalTriangle:
I saw the fix on the honorifics and saw that there was no gender neutral option, so I added one.


monikLover:
Approved.

PR #824, opened 2018-01-23
Suggestion: More games to play with Monika

JustMxnika:
(I've never used GitHub, so this may be in the wrong section.)

I have a few suggestions for some games to play with Monika!

Connect 4 could be fun, along with tic tac toe. Both of these are rather simple games and would fit with the other games already in the game.


monikLover:
Put suggestions in `issues`.

PR #870, opened 2018-01-29
Add a few new varieties of good bye

kaido1224:
Two pretty simple ones and then a quote that i found that came from "Unknown" that I think Monika might say.


monikLover:

PR #871, opened 2018-01-29
[Bugfix] Fixes spaceroom calls in script-greetings.rpy

badooga:
It would appear that one needs to add `$ scene_change = True` before calling `spaceroom`, or else the background won't actually change. Also removed "the window..." text in `monikaroom_greeting_opendoor` due to the window transparency for the classroom not working properly (probably a layer issue) and causing the windows to not work as intended (re-adding this little part of the dialogue should be done eventually). Lastly, I hid Monika at the beginning of `monikaroom_greeting_opendoor` as it makes a bit more sense logically and visually, although this is more of a nitpick of mine than anything else.

### DIFF
--- a/Monika After Story/game/script-greetings.rpy
+++ b/Monika After Story/game/script-greetings.rpy
@@ -341,6 +341,7 @@ init 5 python:
 label monikaroom_greeting_ear_narration:
     m "As [player] inches [his] ear toward the door,{w} a voice narrates [his] every move."
     m "'Who is that?' [he] wondered, as [player] looks at [his] screen, puzzled."
+    $ scene_change = True
     call spaceroom from _call_spaceroom_enar
     m 1k "It's me!"
     m "Welcome back, [player]!"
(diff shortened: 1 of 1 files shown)


ThePotatoGuy:
That was the point, dude. We're portraying monika as a beginner still learning how to code.

badooga:
@ThePotatoGuy I probably should have clarified what these changes fix; they don't fix any of the Monika-created "bugs" that are intentionally put in the topics, they fix actual, persisting bugs (that her dialogue doesn't mention or is inaccurate about) like the following [screenshots](https://imgur.com/a/tIYhz) (the third one being that the window animation/image doesn't work at all).

ThePotatoGuy:
Bugs that only you appear to have. I am unable to repro any of those screens using the latest version, and there hasn't been a single issue about these bugs either. 

Regardless, I have identified that `scene_change` isn't defined by the time any of the spaceroom calls are made, so I've made that adjustment, which requires only a single line to be added instead of throwing `scene_change` in front of every spaceroom call and deleting a whole bunch of dialogue.

PR #889, opened 2018-01-31
Update script-greetings.rpy

juliathegolden:
So, the code I submitted back in December for the 'make your own .chr' easter egg didn't quite work as intended. It was looking for the characters folder in the wrong place. I've corrected it.


kaido1224:
Changes in dialog need to be pushed to the Content branch.

PR #937, opened 2018-02-04
Hungry mood

kaido1224:
Hunger mood topic

### DIFF
--- a/Monika After Story/game/script-easter-eggs.rpy
+++ b/Monika After Story/game/script-easter-eggs.rpy
@@ -185,8 +185,8 @@ label natsuki_name_scare_hungry:
     hide monika
     
     #play special music and display glitch text.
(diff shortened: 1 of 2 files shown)


ThePotatoGuy:
Is it intentional to have the screen tear glitch only at the start and not when transitioning back in? Otherwise the transition from glitch back to normal flow is awkward.

kaido1224:
i can add it to the end as well.

ThePotatoGuy:
I would add it to the end, makes it clear its a glitch and that it reverts itself somehow.

kaido1224:
Give me a few minutes and I'll have it fixed. Just wanna test it first.

kaido1224:
OK @ThePotatoGuy I added it and tested it. I'm glad I did because there was a typo in the {cps} statement. It should be good now.

PR #972, opened 2018-02-08
Player's obsession with Monika, MAS, or DDLC

nayrbai:
Hey, I made a topic about Monika asking the player if they are developing an obsession with her or this game. I wrote it so whoever actually is obsessed with the mod can snap out of it and also reestablish that Monika is her own character; not someone you project your own emotional baggage to. Also, is there anyway to make this based on time progression? It would make more sense to have her say this if the player has been playing DDLC for many hours. Thank you


monikLover:
This doesn't seem like a fitting topic to have her talk about in a way where she's telling the player to not be obsessed with her, which would make her very hypocritical as she demonstrates a lot of times that would go against what she's talking about here. Rejected.

subnatica:
wth

PR #1002, opened 2018-02-12
Monika free chat ideas 

Valllaion:
I'm wondering if it's possible to create a chat option that would allow you to type something in and have her respond with something random but in character... kinda like a sophisticated clever bot. I may be asking too much but I thought it was a cool idea that would create infinite possibilities for Monika and the player ^^


monikLover:
Post suggestions in issues, not as pull requests

Valllaion:
I'm sorry, I thought I put it in the right place

PR #1041, opened 2018-02-14
Adjust facial expressions in pong game to match her laughing.

kaido1224:
Adjust facial expressions to be actual laughing or amused faces. It's always bothered me.


monikLover:
ThePotatoGuy: I love you, plz marry me

PR #1051, opened 2018-02-15
The Great Gatsby

Wendetangente:
It´s about The Great a novel which she really enjoyed.

This is my first contribution, so I hope that you can give me some advice. :)


monikLover:
Could be extended more rather than just referencing a quote from it. Rejected,

PR #1069, opened 2018-02-16
Full piano maybe?

Tymon222:
I'd really like to play some songs with their original sounds not other.


ThePotatoGuy:
Suggestions in issues please.

PR #1078, opened 2018-02-17
Greetings in French

Tsadokiel:
Hi there!

While looking at the script, I've noticed some greetings in Italian and in Japanese, so I decided to add another one in French. I thought it would be funny for French-speaking players, since I'm French myself.


monikLover:

PR #1088, opened 2018-02-19
Cleaning up the dialogue a little

Cirrusyte:
Added commas here and there, and made some of the lines look nicer.


monikLover:
Removed some unnecessary edits. Approved.

PR #1100, opened 2018-02-20
Created Good Afternoon Topic

Bam8647:
So I basically copied and pasted the same format the other topics had with this one. Do I have to go to the definitions file to define this topic or no?

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -4784,26 +4784,26 @@ label monika_asks_family:
     return
     
     init 5 python:
-       addEvent(Event(persistent.event_database,eventlabel="monika_good_afternoon",category=['misc'],prompt="[player]'s afternoon",random=True))
+       addEvent(Event(persistent.event_database,eventlabel="monika_good_afternoon",category=['misc'],prompt="good afternoon",random=True))
     
     label monika_good_afternoon:
-    m 1b "Good afternoon to you too, [player]!
-    m "How's the sky?"
-    menu:
-        "Bright and sunny!":
-                         m 1k "Really?! That sounds beautiful!"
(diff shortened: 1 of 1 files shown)


Olliesama:
It needs to be indented properly by 4 spaces every new divergent line and requires some grammar fixes plus the first line seems to be as a response to player but the prompt says "player's afternoon" instead of being phrased as a question.
I'll be posting some fixes here...
```
init 5 python:
       addEvent(Event(persistent.event_database,eventlabel="monika_good_afternoon",category=['misc'],prompt="[player]'s afternoon",random=True))
    
label monika_good_afternoon:
    m 1b "Good afternoon to you too, [player]!"
    m "How's the sky?"
    menu:
        "Bright and sunny!":
            m 1k "Really?! That sounds beautiful!"
            m 1a "You know, it's days like this I wish we could go to the park and have a nice romantic picnic, ehehe~"
            m "You better have the perfect spot picked out for us~"
        "Dark and rainy":
            m 1g "Aww that's okay"
            m 2e "How about we stay inside and watch a movie?"
            m 1k "While we're at it, why don't we pop some delicious popcorn?"
            m 1n "And you know..."
            m 1m "We can cuddle on the coach to keep each other warm"
            m "Which reminds me..."
            m 2d "You'd better have been practicing those snuggling skills!"
            m 1l "Ahaha."
            m 1i "Just teasing you [player]."
            m 1a "To be honest though, this is one of things that I'll be looking forward to once I get out of here..."
            m 1i "Ehehe~"
return   
```

monikLover:
This would be better if it followed the format of the good morning/evening topics, which actually does check the time and delivers a response depending on the time if it really is afternoon or not. This would be weird if the player's time is midnight and she brings this up.

Bam8647:
Okay so how do we fix this then? I mainly made this for some wanting an afternoon topic.

PR #1111, opened 2018-02-21
changed masturbation topic

juliathegolden:
changed masturbation topic to be more succinct and less shame-y.


MarkusRamikin:
Very much an improvement.

subnatica:
ummmmmmmmmmmmmmmmmmmmmmmmmmmm

subnatica:
ok then

CommenterOfComments:
So I remembered this existing after getting the topic again, and I was curious, why was this closed? It seems like they're almost trying to hide their reasoning for just leaving this to dust.

PR #1126, opened 2018-02-23
self-pleasure - wrong word used

MarkusRamikin:
"Objectively" sounds weird, since normally it means "without bias" or "relying on objective measures rather than direct perception". 

Doubtless whoever wrote this meant "in an objectifying way", and that would have worked as a rewrite. But "objectification" is such a buzzword nowadays that I felt actually unpacking that concept made for better dialog for Monika.

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -3005,7 +3005,7 @@ label monika_pleasure:
     m "And sometimes, even if you don't feel the urge, you'll always find yourself wanting to do so."
     m 1o "Not to mention..."
     m 1r "Being addicted to the feeling causes you to view the world from a perverted point of view."
-    m "From what I hear, people addicted to self-pleasure often see other people of the opposite gender objectively."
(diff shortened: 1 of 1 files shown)


ThePotatoGuy:
Hmm, actually I think more of that line should be rewritten. Maybe something like this:

`From what I hear, people addicted to self-pleasure often see others as objects, not as persons.`

(We'll let a writer weigh in on this though).

MarkusRamikin:
Oh, the whole dialog is problematic. And someone already has made a good proposal for rewriting it:

https://github.com/Backdash/MonikaModDev/issues/1109#issuecomment-367467245

But it seems to have been ninja-rejected.

ThePotatoGuy:
@monikLover , could you at least mentioning reasoning behind closing pull request #1111 ?

MarkusRamikin:
> @monikLover , could you at least mentioning reasoning behind closing pull request #1111 ?

Or the related issue 1109?

monikLover:
I'll ask a consensus from our writers. Could do with a look-over and a rewrite of the topic.

Issue #810, opened 2018-01-21
Chess features: 1, highlight last move, 2. save game log 3. skill level-dependent responses


MarkusRamikin:
I love the chess. She seems to get better as you beat her, and worse as you lose, which is great, I actually mostly play this mod for the chess now.

Suggestion 1: make it more visually obvious somehow what Monika's last move was. 

Ideally, include a box showing a notation past moves. But more simply, the color highlight that exists is easy to miss.

Suggestion 2: finished games should be saved to a text file in chess notation. 
(I just had an awesome game which I barely won - it would have been a treat to review it!)

ThePotatoGuy:
Regarding suggestion 2, is [PGN](https://en.wikipedia.org/wiki/Portable_Game_Notation) the appropriate notation for chess games?

MarkusRamikin:
Sure!

Just BTW, I looked up the script. The dialog line "You really are an amazing player!" is only available if you beat her at Stockfish level 20? So basically, Magnus Carlssen could get it, but no human actually likely to play this game is ever going to see that line? *grin* Come on, give us a chance!

ThePotatoGuy:
Yeah we could look into lowering that value. What level are you at? If you don't know, upload your `persistent` file here and I can check it.

MarkusRamikin:
Oh, I'm terrible (currently oscillating around level 3-4, after decades of not playing), I don't expect to ever see that message myself. It just occurred to me that maybe no human has or ever would see it the way it's coded now, not without cheating. 

Although maybe that's the point, that level 20 Stockfish is - as well as I expect a true AI would be - unbeatable without cheating, and Monika meant to be pointing this out? 

Then how about instead adding new responses every few tiers? That would impact most players instead of the top few and/or the ones cheating, and let people see their strength without having to somehow unpack the persistent, or lower immersion by putting the actual number in the game. Something like:

```
        if persistent.chess_strength<2:
            m "I have to admit, I put less pressure on you there than I could have..."
            m "I hope you're not angry! I'll be challenging you more as you get better, I promise!"
        if persistent.chess_strength<4:
            m "I'll get you next time for sure!"
        elif persistent.chess_strength<6:
            m "You really didn't play badly there!"
        elif persistent.chess_strength<8:
            m "That was an interesting game!"
        elif persistent.chess_strength<10:
            m "You're quite a strong player!"
        elif persistent.chess_strength<12:
            m You're a very challenging opponent, [name]!"
[...]

PR #1159, opened 2018-02-26
Migration of Chess save system to a Quicksave system

ThePotatoGuy:
#810 

Here it is, 2 weeks later. An absolute conglomerate of call statements, permanent consequences and _interesting_ dialogue. 

## Key features

This monster of a pull request does multiple things:
- Adds a one-time Monika dialogue that transitions the player's multiple chess saves (if they have any) into a single quicksave-style system.
- Chess saves are now saved both internally and in a file. If edits are made to either saves, appropriate dialogue will be shown.
- Chess winning/losing dialogue now has more lines for differing strength levels.
- All chess dialogue has been modularized in preparation for API.
- Adds a special timed_jump screen that can be used for timed menus (#774)
- If the player edits/removes saves too much, **chess can be disabled for a time**. If the player continues to be bad, **permanent actions can occur**.

## Dialogue flows

There is a ridiculous amount of dialogue, but most of it is short stuff, so I don't think a full-bodied review is necessary. But y'all can do that if you want.

Here are the main flows:

### User edits internal save (or internal save got corrupted somehow)
1. Monika mentions that the internal save is messed up. Depending on the amount of times this has happened, different sets of dialogue will be shown.
2. In most cases, she will apologize.
3. In some cases, she will be slightly suspicious of the player, but no harm will come.
4. Currently, all cases will lead to a new game being started.

### User removes external save 
1. Monika mentions that the external save is missing.
2. Monika asks if the player did anything. the player has a choice:
    1. Of course not! (player denies removing save)
        - Monika will believe player, but depending on the amount of times this has happened, she will slowly not believe player. In most cases, she will start a new game.
        - After the 6th time, Monika permanently locks chess forever. (See Tech Notes 2)
    2. Maybe... (player sheepishly admits to removing save)
        - Monika asks player to put the save file back. She'll give the player a minute. (See Tech Notes 3)
        - If the player obliges, the game will be continued as normal. 
            - If the player removes file again before it is read, Monika will be like wtf and disable chess temporarily (See Tech Notes 4)
        - If the player doesn't, Monika will start a new game.
        - After the 3rd time this occurs, Monika will reveal that she has kept a backup (internal) of the save. From now on, the external file is never checked again.
    3. It was an accident! (player admits to removing save, but says it wasn't his/her fault)
        - Monika forgives player. No bad things here.
        - In most cases, she will start a new game.
        - After the 3rd time, Monika will reveal she has a backup (internal) of the save. From now on, the external file is never checked again.

### User edits the external save
1. Monika uncomfortably asks if the user edited the saves.
    1. Yes (player admits)
        1. The first time, Monika is disappointed but gives user opportunity to apologize. 
            - Player can choose to or not. If player doesn't make choice in 5 seconds, assumed not. (using timed jump screen)
            - If player apologizes, internal save is used. Otherwise a new game is started.
        2. The 2nd time, Monika is really disappointed. Disables chess temporarily (See Tech Notes 4)
        3. Monika reveals the backup (internal) save. Her chess strength is set to 20 temporarily. From now on, the external save file is never checked again.
    2. No (player denies)
        1. The first time, Monika believes player, but mentions that the game looks different. **The edited save is used**, but Monika's chess strength is at max temporarily.
        2. The 2nd time, Monika is silent, but agrees to continue the game. **The edit save is used**, but Monika's chess strength is at max temporraily.
        3. The 3rd time, Monika reveals backup, but gives user opportunity to apologize.
            - Player can choose to or not. If playter doesnt make choice in 3 seconds, assumed not. (using timed jump screen)
            - If player apologizes:
                - internal save is used 
                - Her chess strength is set to 20 temporarily. 
                - From now on, the external save file is never checked again. 
                - Chess winning/losing dialogue is permanently changed. 
                - Monika berates player for editing internal save if internal save cannot be read in future loads. She also time disables chess.
            - Otherwise:
                - Monika says she cant trust player anymore.
                - Monika goes on rampage:
                    - She deletes her character file.
                    - She deletes chess.rpyc
                    - She mangles (corrupts) several other game files to prevent loading.
                    - She resets player data in `persistent`
                    - certain values are retained in persistent. these values are checked in `splash` and will run the deletion process again if True. The only way the player can stop this is by starting with a fresh persistent. (This may be added to a MultiPersistent as well).

### Winning/Losing chess dialogue
The dialogue for winning/losing chess has been partitioned into multiple labels based on chess strength. This allows for a greater variety of dialogue. Dialogue when the player surrenders or if the player loses normally have been separated into 2 different flows. In addition, if the player edited the saves 3 times and apologized, a different set of winning/losing chess dialogue is used. This extra set is much meaner than the original and uses the `MASQuipList` class (See tech notes 5).

### Endgame dialogue
The endgame (end of chess) dialogue has been modularized. One-liner skip variants will be shown after the user has seen the long-form end dialogue.

## Testing
I will admit, testing all of these flows is **insanely difficult and tedious**. I personally spent about the past 5 days testing every single one of the main dialogue flows. I haven't tested all of the winning/losing based on chess strength flows because that has **60** possible flow directions. I did test if it worked generally, though. 

Because of the difficulty, I'm not recommending anyone test _everything_ unless you have tons of time on your hand. The majority of the edited/removed file dialogue flows **would not** be seen if the player does everything correctly, which is really what we hope for here. 

Chess save migration is the easiest thing you can test, and arguably the most important. To test that, all you need to do is have multiple chess games created with the current version, then use that `persistent` with these changes.

For testing post-game flows, I have a debug line with `~~EDIT ME~~` in it. You can uncomment that line to pause execution right after a game ends and open the dev console to change local vars. The ones that matter for post-game are `is_surrender` and `is_monika_winner`. You can change these bools to trigger particular post-game dialogue flows.

## Technical Notes:
1. The timed jump screen works by using a `Timer` and performing a `Jump` when the timer runs out. The screen expects the number of seconds to wait and a label to jump to.
2. Monika locks chess forever by setting the appropriate `game_unlocks` value to False, and setting the `unlock_chess` label into the `_seen_ever` dict to prevent chess from being unlocked ever again.
3. File detection is done using `renpy.exists` and a loop that pauses every second. This could be turned into a function in the future.
4. Chess can be disabled temporarily by setting `persistent._mas_chess_timed_disable` to a `datetime`. Currently this is hardcoded to disable for an hour.
5. `MASQuipList` is a dope class that allows for multiple types of quips instead of all one liners, which is currently used in random quips (and the unmerged love quips). This class enables us to use label quips and line quips in the same list by differentiating them by types. Check `definitions` for more info.

## Future
- More dialogue should probably be added to the chess-strength-variable sets of dialogue.
- More quips can be added to the existing mean ones.


kaido1224:
Tested the cheating mechanic and it seems to be working the way it should be.

Also the save feature appears to be working fine too as far as I can tell.

kaido1224:
Save worked fine for me in transition from the old version to the new version.

ThePotatoGuy:
[Executing admin previleges for merging]

PR #1162, opened 2018-02-27
Edited: Be Careful topic

jmwall24:
Basically, Monika is worried if the player suddenly had a tragic accident and never got to tell her goodbye, she would never know what happened, and wants you to make sure someone notifies her if that happens.  Trying to take the approach of if your entire connection to the outside world is only one person that you care deeply about, what would your worries and insecurities be.

This is my first time creating dialogue, so please, any critique is encouraged.  I hope I got the coding right, but wasn't quite sure on what category to put it in.  Went with 'you' for now.


Philbo-Swaggins:
Could you add an option to actually tell her that something bad has happened. This topic doesn't seem to have much purpose if you couldn't.

jmwall24:
I actually had that thought when I wrote it, but I'll explain why I didn't do it.

First, it wouldn't really fit in that topic itself, it'd need to be it's own topic.  This is just a plan for if something happens, not if something actually does.  Second, implementing that would lead to a whole bunch of other stuff and completely change the direction the game went afterwards, which is way out of my league.

So basically this was written to help alleviate what I feel would be a very real fear for Monika to have, and to come up with a plan on how to deal with it if it does, and not the actual execution of that plan if it ever arises.  There are other topics where Monika gives advice or asks you to do something where it's just kinda implied that you do it, without actually having the capability of doing it in-game, and that's pretty much what this is.

It's very possible none of that makes sense lol  But that was my thought process anyway.  If it doesn't end up having a purpose to add it, I totally get why.

jmwall24:
I have a few ideas on how to make it flow better into the overall conversation.  I'll work on it tonight.

jmwall24:
Ok, I rewrote it to hopefully 'fit' a little better.

MrTwilight:
Why did you erased the possibility of telling Monika that you had already told someone?

jmwall24:
Just because of the way I reworded it.  Tried to make it more like some of the other topics where she's more or less telling you to do something for her instead of asking.  I can add it back if needed, just thought it flowed better this way.

jmwall24:
#1175 is obviously going to have a big impact on this depending on if anything like that is ever added.  So will be keeping an eye on that and adjust accordingly.

Issue #1169, opened 2018-02-28
Anniversary question/possible issue


jmwall24:
Is the anniversary date used simply the creation date on the persistent file or is it coded somewhere inside it, and if the latter is there anyway to open it and see what it is?

I ask because I first installed MAS on January 10th and got the 1 week as expected.  Then a few weeks later I got it again.  Thought maybe an update somehow reset it so didn't think anything of it.   But I think I'm at or past a month now from when the new date would've been and haven't gotten the dialogue.  So I'm just trying to figure out what the anniversary actually is so I know if there's an issue or not.

ThePotatoGuy:
Could be a bug, we'll take a look into it.

TinyInternetTree:
There's anniversary dialogue? I didn't even know it existed, let alone gotten it.

jmwall24:
Maybe it's not just me then lol

jmwall24:
Just to update this, the anniversaries are working now, got the 3 month dialogue today.  It was based off the second time I got the 1 week dialogue, so my initial date did get reset somehow.  I believe it corresponded to that time in January that an 'unstable' version was released for a day or 2 before the actual update came along, so I'd venture a guess that that's what did it.

PR #1176, opened 2018-03-01
Fix add_months func, related to #1169

aldoram5:
The issue: The add_months function was adding 1 extra year.

This fixes that issue by calculating the new month and the new year in another way, first it gets the "Total months" then using modulo 12 it gets what the new month should be. 
The new year is calculated by dividing the "total months" by 12 and adding the result to the old year.
The other parts of the function where left like they were since that part was working fine.

This can explain why #1169 happens


jmwall24:
Nice, thanks for looking into this

PR #1188, opened 2018-03-04
Added the option to apologize

vermicellibug:
Because I said something I shouldn't have to my Moni


TinyInternetTree:
So just out of curiosity, what did you say to her?

Subzerosan:
This is the wrong place, please put this under issues as a suggestion, thanks.

monikLover:
I like this. This could be expanded where it can check what the player did in the base game by checking the `persistent.playthrough`. Approved.

PR #1194, opened 2018-03-04
MAS persistent updates

kaido1224:
Updated some of the MAS specific persistents to meet the naming convention


ThePotatoGuy:
Is `player_dialogue` used anywhere else other than in the old keyword system?

ThePotatoGuy:
What I mean is that if it's only used in outdated code, we should just comment it out (along with that keyword section).

kaido1224:
Oh Gotcha, I have no clue. I'll look when I get a chance.

kaido1224:
disregard this

PR #1195, opened 2018-03-05
Fun Facts Topic!

Olliesama:
Adds 20 good fun facts, and 4 bad fun facts into 1 topic.
10% of getting a bad fact.

I may make some minor edits later but it should run fine.


Olliesama:
Thanks for the input, I'll fill the syntax and way the code is worded when I get back.
I purposefully have it done in a fashion to make bad facts a rarity which means if I'm making each fact a label which I'm jumping to then I'll need to redo how the chances of the bad facts could occur...

aldoram5:
No worries, there's no rush.
Also for the bad facts and good facts probability, you could put them in different lists and still use the random to determine from which list you'll be picking the fact, that way you can keep the bad ones rarity like you want.
Also you may want to wait until Potato gets a look at this, since this one is a little bit more than just the average topic.

Monika After Story/game/script-topics.rpy
```diff
@@ -4940,3 +4940,297 @@ label monika_asks_family:
...
+init 5 python:
+    addEvent(Event(persistent.event_database,eventlabel="monika_fun_facts_open",category=['misc'],prompt="Fun facts",pool=True))
+
+default fun_facts_start = 0
+default persistent._mas_funfactfun = True
+
+label monika_fun_facts_open:
+    #Determines the fact number and whether it's a bad fact
+    python:
+        fun_facts_bad_chance = random.randint(1,100)
```
ThePotatoGuy: While `random` is usable here, we prefer to keep things within renpy scope so we don't have to add additional `imports` somewhere. Use the `randint` [provided](https://www.renpy.org/doc/html/other.html#renpy-random) by renpy. (Which it appears that you actually did do that a few lines below).

Monika After Story/game/script-topics.rpy
```diff
@@ -4940,3 +4940,297 @@ label monika_asks_family:
...
+        m 1a "Say [player], would you like to hear a fun fact?"
+        m 4a "I've been looking some up just to try and teach us both something new."
+        m 4j "They say you learn something new everyday, this way I'm guaranteeing it!"
+        m 2l "I found most of these online so I can't say they're {i}definitely{/i} true... "
+        $ fun_facts_start = 1
+    m 3a "Now, let's see..."
+
+#Determines if it is a bad fact, 10% chance.
+    if fun_facts_bad_chance <= 10:
+        $ renpy.jump('bad_facts_'+str(renpy.random.randint(1,4)))
```
ThePotatoGuy: While this is perfectly acceptable, you should use the renpy-native `jump` command rather than the python equivalent whenever you're in renpy script. More on that [here](https://www.renpy.org/doc/html/label.html#jump-statement). 

The syntax would be `jump expression <variable containing the label as a string>`.

Yes, this would make your one-line jumps turn into 2 lines, but with some clever work:
```renpy
if bad_chance:
    $ jumpfact = "bad_facts_{0}".format(renpy.random.randint(1,4))
else:
    $ jumpfact = "fun_fact_{0}".format(renpy.random.randint(1,20))

jump expression jumpfact
```
It'll look cleaner.

Also, the number of fun/bad facts should be constants rather than literals so its easier for us to keep track of how many facts there are.
Define these in a store like:
```renpy
define mas_funfact.fun_count = 4
```

Monika After Story/game/script-topics.rpy
```diff
@@ -4940,3 +4940,297 @@ label monika_asks_family:
                 m 1f "You can tell me about your family when you're ready, [player]." 
                 m 1j "I love you very much!"                              
     return
+    
+
+init 5 python:
+    addEvent(Event(persistent.event_database,eventlabel="monika_fun_facts_open",category=['misc'],prompt="Fun facts",pool=True))
+
+default fun_facts_start = 0
```
ThePotatoGuy: It looks like `fun_facts_start` just flips between 1 and 0. A boolean would be a better choice.

Olliesama:
Updated, thanks for the suggestions Potato!
It really taught me a lot. :)

Monika After Story/game/script-topics.rpy
```diff
@@ -4940,3 +4940,300 @@ label monika_asks_family:
...
+    addEvent(Event(persistent.event_database,eventlabel="monika_fun_facts_open",category=['misc'],prompt="Fun facts",pool=True))
+
+default fun_facts_started = False
+default persistent._mas_funfactfun = True
+define mas_funfact.fun_count = 20
+define mas_funfact.bad_count = 4
+
+label monika_fun_facts_open:
+    #Determines the fact number and whether it's a bad fact
+    python:
```
aldoram5: The python block is no longer needed now that you're using renpy's random.

```
    # Determines the fact number and whether it's a bad fact
    $ fun_facts_bad_chance = renpy.random.randint(1,100)
```

Olliesama:
There you go.

Monika After Story/game/script-topics.rpy
```diff
@@ -4940,3 +4940,299 @@ label monika_asks_family:
...
+    addEvent(Event(persistent.event_database,eventlabel="monika_fun_facts_open",category=['misc'],prompt="Fun facts",pool=True))
+
+default fun_facts_started = False
+default persistent._mas_funfactfun = True
+define mas_funfact.fun_count = 20
+define mas_funfact.bad_count = 4
+
+label monika_fun_facts_open:
+    #Determines the fact number and whether it's a bad fact
+    fun_facts_bad_chance = renpy.random.randint(1,100)
```
aldoram5: you need the `$` character before `fun_facts_bad_chance` otherwise it crashes the mod 

The line should look like this:
`    $ fun_facts_bad_chance = renpy.random.randint(1,100)`

Monika After Story/game/script-topics.rpy
```diff
@@ -4940,3 +4940,299 @@ label monika_asks_family:
...
+        m 1b "Up for another fun fact [player]?"
+        if persistent._mas_funfactfun:
+           m 2j "That last one was pretty interesting after all!"
+        else:
+           m 2m "I know that last one wasn't great...but I'm sure this next one will be better."
+    m 3a "Now, let's see..."
+
+#Determines if it is a bad fact, 10% chance.
+    if fun_facts_bad_chance <= 10:
+        $ jump_fact = "bad_facts_{0}".format(renpy.random.randint(1,4))
```
ThePotatoGuy: You added constant versions of the literals, which is good, but you didn't put them here. The idea behind this is that if we'd only have to change one line instead of every line the literal is on.

Olliesama:
Updated, and fixed...I hope. Sorry about that Potato.

kaido1224:
I was testing this finale, and i ran into an issue. I tested this on a brand new copy of MAS (with no persistent) and when I clicked the topic, it implied that i've seen fun facts before, but i definitely hadn't. So there appears to be something wrong here. If I had to guess, it has something to do with:

```if fun_facts_started:```

https://youtu.be/9V5euZe1Azs

You might wanna check it out.

Olliesama:
Thanks for pointing out the issue, I've hopefully fixed it. It at least worked on my end.

kaido1224:
Hmm.. do you test your PRs in MAS before you submit them? Just wondering.

I tested this again and I ran into an error.

capture

You used [i] and [/i] instead of {i} and {/i} in the dialog which caused the error above.

Olliesama:
I do. :/

PR #1222, opened 2018-03-11
Event rules system / New greetings system 

aldoram5:
This pull request contains a new way to work with the event system by using a rule based system. Right now the rules are mostly to specify a time when an event can be accessed by using a Numerical rule or a Selective rule. Also contains rules for greetings only.

As for the greeting system, it now uses the `Event` class to store greetings as events, then filters out the appropriate greeting and shows that to the player. The system also helps with special greetings that need to skip the normal visuals initialization.

##  Key features

With this pull request the `Events` get the ability to define rules for themselves. Right now only 3 rules exist `MASNumericalRepeatRule`,`MASSelectiveRepeatRule` and `MASGreetingRule`.

#### MASNumericalRepeatRule 

This rule is defined by a `repeat` property which specifies the time interval for the repetition and the  `advance_by` property which specifies how many of the time intervals the next repetition is going to get scheduled. This rule is intended to help updating the `start_date` and `end_date` of the `Event` containing this rule, so it is intended to play nicely with the current calendar system.

An example of how this rule works:

Let's suppose we have an `Event` scheduled for January 1st of 2018, if we want this event to repeat on January 1st of 2019 we just add a new Numerical rule specifying the `repeat` property to `year` and the `advance_by` to 1.  If we wanted to make that event run each two months we would have defined the  `repeat` property to `month` and the `advance_by` to 2.

The `repeat` accepts only the values: "day", "week", "month" and "year".
The `advance_by` accepts any integer higher than 0. 

#### MASSelectiveRepeatRule

This rule is defined by a set of parameters: `seconds`, `minutes`, `hours`, `days`, `weekdays`, `months`, `years`. Those parameters specify ranges for when the `Event` can be accessed. The rule can check against the current time or any arbitrary given time to see if it fits in the specified ranges.

An example of how this rule works:

Let's suppose we have an `Event` that we want it to be available only from 9am to 5 pm, to achieve that we need to add a new Selective rule with the `hours` property containing an array containing `[9,10,11,12,13,14,15,16]`, that array can be defined also with the python built-in `range` function: `range(9,17)` that way it feels more natural to use it. This rule isn't limited to use only one of it's properties, for example to specify that an `Event` is only available on friday 13th of any month we would specify a rule with the property `days` set to `[13]` and the property `weekdays` set to `[4]` (weekdays start on monday = 0 and end on sundays = 6).

Each property is validated to only accept valid time intervals.

Dev Note: When we first discussed this rule the weekdays property was taken out, however after testing the system and working with it I felt that the property is actually useful so I thought it was better to add it than not. It can be taken out again if requested.

####  MASGreetingRule

This rule is specific to greetings, it's only used to define two things: first if the greeting should skip the visual initialization and second to define if the greeting has a special random chance of appearing. This is done by using the properties `skip_visual` and `random_chance`. 

## Greeting system

The old greeting system was pretty much always random and didn't played nice with special greetings like the "Monika in room" one.

This new system takes advantage of the existing `Event` system and also uses the newly added rules. This system also plays nicely with special greetings. 

#### New greeting selecting flow

First we still do the Yuri scare checking (this might be changed if requested), if the Yuri scare doesn't trigger we proceed to check if player put Monika to sleep correctly, if the player put her to sleep correctly we start checking the greetings database and see if any greeting can be triggered now by checking it's rules. If no greeting is available we check for the special random ones and try to trigger them by their specified odds. If we still don't have a greeting to show we default it to random greetings, these are specified when adding the greeting to the greetings database.

## Final notes

I also added a helper function named `mas_create_rules`, it helps to create the rules dict without having to deal with lots of constants and makes it more straightforward. 
Example: 
```
    rules = mas_create_rules(weekdays=[0], hours=range(5,13))
``` 
I think the greetings system should prioritize Numerical rules when selecting greetings, since those seem to be more date specific and they are less common than others. Didn't code it like that because I want feedback about it first.

I added a few TODO's to `script-greetings.rpy` and some testing greetings.

If there's anything I should change(class definitions, documentation, coding style, how the flow works, etc.) please let me know.

## Edit after review and making the requested changes

Now the rules are easier to use, here's an example of how the flow works now:

```
init 5 python:
    rules = dict()
    rules.update(MASGreetingRule.create_rule(skip_visual=False, random_chance=10))
    addEvent(Event(persistent.greeting_database,eventlabel="greeting_gooday", unlocked=True, rules=rules),eventdb=evhand.greeting_database)
```
The system now filters out those greetings who aren't unlocked, then follows it's normal flow, also the greeting gets pushed in the same place the old system pushed them.

### DIFF
--- a/Monika After Story/game/script-ch30.rpy
+++ b/Monika After Story/game/script-ch30.rpy
@@ -499,7 +499,7 @@ label ch30_autoload:
             $ play_song(persistent.current_track)
         else:
             $ play_song(songs.current_track) # default
-    
+
 
     window auto
     #If you were interrupted, push that event back on the stack
@@ -538,12 +538,22 @@ label ch30_autoload:
 
     #Skip all greetings if you closed the game on Monika
     if persistent.closed_self:
-        #pick a random greeting
-        if is_monika_in_room:
-            if persistent.current_monikatopic != "i_greeting_monikaroom":
-                $ pushEvent("i_greeting_monikaroom")
-        else:
-            $pushEvent(renpy.random.choice(greetings_list))
+        # pick a proper greeting
+        # TODO use the ev handler to make it work
+        # $pushEvent(renpy.random.choice(greetings_list))
+        python:
+            ruled_greetings = Event.checkRules(evhand.greeting_database).keys()
+            if len(ruled_greetings) > 0:
+                pushEvent(renpy.random.choice(ruled_greetings))
+            else:
+                random_greetings = Event.filterEvents(evhand.greeting_database,random=True).keys()
+                pushEvent(renpy.random.choice(random_greetings))
+
+#        if is_monika_in_room:
+#            if persistent.current_monikatopic != "i_greeting_monikaroom":
+#                $ pushEvent("i_greeting_monikaroom")
+#        else:
+#            $pushEvent(renpy.random.choice(greetings_list))
 
     if not persistent.tried_skip:
         $ config.allow_skipping = True
@@ -617,8 +627,8 @@ label ch30_loop:
 
         python:
             if (
-                    mas_battery_supported 
-                    and battery.is_battery_present() 
+                    mas_battery_supported
+                    and battery.is_battery_present()
                     and not battery.is_charging()
                     and battery.get_level() < 20
                 ):
--- a/Monika After Story/game/script-ch30.rpy
+++ b/Monika After Story/game/script-ch30.rpy
@@ -483,55 +483,60 @@ label ch30_autoload:
                 if renpy.has_label(ev_label)
             ]
 
-    #Skip all greetings if you closed the game on Monika
-#    if persistent.closed_self:
-        # pick a proper greeting
-        # TODO use the ev handler to make it work
-#        $ ruled_greetings_dict = Event.checkRules(evhand.greeting_database)
-#        $ ruled_greetings = ruled_greetings_dict.keys()
-#        if len(ruled_greetings) > 0:
-#            $ selected_label = renpy.random.choice(ruled_greetings)
-#            $ mas_skip_visuals = MASGreetingRule.should_skip_visual_if_rule_exists(ruled_greetings_dict[selected_label].rules)
-#            $ pushEvent(selected_label)
-#        else:
-#            $ special_random_greetings_dict = Event.checkGreetingRules(evhand.greeting_database)
-#            $ special_random_greetings = special_random_greetings_dict.keys()
-#            if len(special_random_greetings) > 0:
-#                $ selected_label = renpy.random.choice(special_random_greetings)
-#                $ mas_skip_visuals = MASGreetingRule.should_skip_visual_if_rule_exists(special_random_greetings_dict[selected_label].rules)
-#                $ pushEvent(selected_label)
-#            else:
-#                $ random_greetings_dict = Event.filterEvents(evhand.greeting_database,random=True)
-#                $ random_greetings = random_greetings_dict.keys()
-#                $ selected_label = renpy.random.choice(random_greetings)
-#                $ mas_skip_visuals = MASGreetingRule.should_skip_visual_if_rule_exists(random_greetings_dict[selected_label].rules)
-#                $ pushEvent(selected_label)
-
     # yuri scare incoming. No monikaroom when yuri is the name
     if persistent.playername.lower() == "yuri":
         call yuri_name_scare from _call_yuri_name_scare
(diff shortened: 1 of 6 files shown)


kaido1224:
You are amazing aldo.

Monika After Story/game/script-ch30.rpy
```diff
@@ -536,21 +580,12 @@ label ch30_autoload:
     #Run actions for any events that are based on the clock
     $ evhand.event_database=Event.checkCalendar(evhand.event_database)
 
-    #Skip all greetings if you closed the game on Monika
-    if persistent.closed_self:
-        #pick a random greeting
-        if is_monika_in_room:
```
ThePotatoGuy: FYI, greetings should be the last thing pushed onto the event stack prior to the jump to `ch30_loop`. Just set a local var to the selected greeting in the checks above and have it be pushed onto Event stack here if the greeting is not None

Monika After Story/game/script-greetings.rpy
```diff
@@ -6,37 +6,36 @@
...
-
+    addEvent(Event(persistent.greeting_database,eventlabel="greeting_sweetheart", random=True),eventdb=evhand.greeting_database)
 
 label greeting_sweetheart:
     m 1k "Hello again, sweetheart!"
     m 1l "It's kind of embarrassing to say out loud, isn't it?"
     m 3b "Still, I think it's okay to be embarrassed every now and then."
     return
 
+init python:
```
ThePotatoGuy: All of these should be init level 5.

Monika After Story/game/script-greetings.rpy
```diff
@@ -6,37 +6,36 @@
...
+    addEvent(Event(persistent.greeting_database,eventlabel="greeting_honey", random=True),eventdb=evhand.greeting_database)
```
ThePotatoGuy: Make these all `unlocked=True`. Greetings should also only appear if they are unlocked.

Monika After Story/game/script-greetings.rpy
```diff
@@ -6,37 +6,36 @@
...
+    addEvent(Event(persistent.greeting_database,eventlabel="greeting_back", random=True),eventdb=evhand.greeting_database)
+
 label greeting_back:
     m 1a "[player], you're back!"
     m 1e "I was starting to miss you."
     m 1k "Let's have another lovely day together, alright?"
     return
 
+init python:
+    rules = {EV_RULE_GREET_RANDOM : MASGreetingRule.create_rule(skip_visual=False, random_chance=10)}
```
ThePotatoGuy: Yeah so I don't think this is a good way to do this. It would be hard for regular users to add rules since they'd have to knowingly match the correct key to the correct rule. What we should be doing here would be like:
```python
    ev = Event(.......)
    MASGreetingRule.create_rule(ev, ......)
    addEvent(ev,....)
    del ev
```
**This only applies to greetings with Rules**

ThePotatoGuy:
I did a bit of rethinking, and I think you should change the `create_rule` functions so they return `dicts` with the proper `key: tuple` setup rather than just tuples. That would be a better way to handle automatic rule dict creation rather than the `mas_create_rules` function you had.

Monika After Story/game/script-greetings.rpy
ThePotatoGuy: OKAY Ignore what i wrote above.
If the `create_rules` functions are built in the way i described in the comment below, then the `rules` dict could be built on its own like you did originally and then set as `rules` during `Event` creation.
something like:
```python
rules = dict()
rules.update(MAS....create_rule(....))
rules.update(MAS....create_rule(...))
addEvent(Event(....rules=rules)....)
del rules
```

aldoram5: I definitely love this idea, it takes away the complexity of using constants everywhere to add just a dict, while still leaving the rules not so tied to `Event`, meaning we could use those for other things in the future.

ThePotatoGuy:
Okay so because of how the rules dict should be built, you can keep the exceptions in the functions.

I've removed those comments from the review appropriately.

aldoram5:
Thank you so much for that review, it will help me improve.
I've applied the requested refactors and tested all the rules to be sure they work as expected after the refactors.
Since this Pull request was conflicting with master I took the liberty to rebase it and fix the conflicts, so it's easier to merge it.
If there's anything that needs to be changed please tell me.
Thanks in advance.

ThePotatoGuy:
This was fucking lovely, thank you.

Issue #878, opened 2018-01-30
More farewells


therationalpi:
Monika only has two ways to say goodbye at the moment. We need to add more ways to say goodbye including ones that are time based.

I will put it on myself to extend the farewell system to make time-based farewells easier to implement, but I will need help from the writers on actually making enough ways to say goodbye.

Keep in mind that we want a mix of interesting "One-time" farewells that are longer and shorter "reusable" farewells that can be randomly chosen from even after all the special farewells are depleted.

therationalpi:
https://github.com/Backdash/MonikaModDev/pull/870

Here's some, can still use more.

LokiOdinson:
I am sorry for my English. But I have a problem! I installed the mod, but I can not talk and play with Monica. What should I do?

ThePotatoGuy:
oh this is done

PR #1270, opened 2018-03-17
Farewells system rework

aldoram5:
This pull request modifies the old farewells system so it now checks for rules first to filter the farewells based on different conditions. 

This also adds a new rule specific for farewells: `MASFarewellRule`, right now this rule only defines a special random chance for the farewell to be selected.

Basically you can define a special ruled farewell using something like this:
```
init 5 python:
    rules = dict()
    rules.update(MASSelectiveRepeatRule.create_rule(hours=range(21,24)))
    addEvent(
        Event(
            persistent.farewell_database,
            eventlabel="bye_going_to_sleep",
            unlocked=True,
            rules=rules
        ),
        eventdb=evhand.farewell_database
    )
    del rules

label bye_going_to_sleep:
    m "Are you going to sleep, [player]?"
    m 1e "I'll be seeing you in your dreams"
    return 'quit'
```

This pull request addresses the changes discussed here: https://github.com/Monika-After-Story/MonikaModDev/issues/878

### DIFF
--- a/Monika After Story/game/dev/dev_farewells.rpy
+++ b/Monika After Story/game/dev/dev_farewells.rpy
@@ -33,3 +33,21 @@ label bye_dev:
     m 1e "Or maybe you're just running some tests?"
     m 1k "Don't give up until everything works as expected!"
     return 'quit'
+
+# This one exists so devs get an autoupdate once they pull these changes
+init 5 python:
+    ev = Event(persistent.farewell_database,eventlabel="bye_dev_temp",unlocked=True)
+    addEvent(ev,eventdb=evhand.farewell_database)
+    del ev
+
+label bye_dev_temp:
+    m 1c "Leaving now, eh [player]?"
+    m 1e "We had some changes in the farewell system, I need to update something first ..."
+    python:
+        for k in evhand.farewell_database:
+            # no need to do any special checks since all farewells were already available
+            renpy.store.evhand.farewell_database[k].unlocked = True
+    $ evhand.farewell_database["bye_dev_temp"].unlocked = False
+    m 1k "All done, thanks for waiting [player]!"
+    #Don't show this farewell again
+    return 'quit'
--- a/Monika After Story/game/dev/dev_farewells.rpy
+++ b/Monika After Story/game/dev/dev_farewells.rpy
@@ -1,0 +1,35 @@
+# dev related farewells
+
+init 5 python:
+    rules = dict()
+    rules.update(MASNumericalRepeatRule.create_rule(repeat=EV_NUM_RULE_YEAR))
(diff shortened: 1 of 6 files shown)


Monika After Story/game/dev/dev_farewells.rpy
```diff
@@ -0,0 +1,53 @@
...
+    del rules
+
+label bye_st_patrick:
+    m 1c "Aww, leaving already?"
+    m 1e "It's really sad whenever you have to go..."
+    m 1b "Good luck with the hangover!"
+    return 'quit'
+
+init 5 python:
+    ev = Event(persistent.farewell_database,eventlabel="bye_dev",unlocked=True)
```
ThePotatoGuy: While this works, this bypasses the init locks system. Doing the traditional:
```python
<setup rules>
addEvent(
    Event(
         ....
         rules=rules
     )
)
del rules
```
allows for the initlocks system to disable changes on init.

Since this is a dev goodbye, its not really an issue, but in production, we should anonymously create Event instances as we call addEvent.
**No change here is necessary, just letting you know of  a standard**

(If you already knew this, then dont mind me~)

Monika After Story/game/event-rules.rpy
```diff
@@ -437,6 +438,72 @@ init -1 python:
...
+            if rule is None:
+                return False
+
+            # store the rule for easy access
+            # keeping this so when we add another rule we'll keep
+            # the same code structure, also it's more readable changing the name
+            random_chance = rule
+
+            # check if random_chance is 0 return False
+            if random_chance == 0:
```
ThePotatoGuy: Rather than check only 0, lets program this defensively and return False if random_chance is `<= 0`

Monika After Story/game/dev/dev_farewells.rpy
aldoram5: Actually I didn't knew that, I did it that way just because I got curious about the the optional argument on the rules and wanted to do a little test with that, that's why I did it on the dev one.
Thanks for explaining that, I'll be sure to never do it this way in any production greeting/farewell.

Monika After Story/game/event-rules.rpy
aldoram5: Alright I'll make that change, I did the check against 0 only because on the creation method I already raise an Exception if the random chance is negative. Should I adjust the greeting check rule too? because there I only check against 0 too.

ThePotatoGuy: Yeah adjust the other one too

aldoram5: Done, thanks a lot fot taking the time to review it!

PR #1282, opened 2018-03-18
Affection to Aff

Olliesama:
Invalid syntax's keep appearing during testing for everyone so this simplifies things.


ThePotatoGuy:
The store is called `mas_affection`. Changing all of these to `mas_aff` will break everything. The top of `script-topics` should have assigned the `mas_affection` store to `mas_aff`, but if that doesn't work (which I'm actually not sure why it wouldn't) then just add import lines in the labels where `mas_aff` is used or rename them back to using full scope (`store.mas_affection`).

PR #1287, opened 2018-03-19
Going away topic.

Olliesama:
Created a going away topic to prevent bad exp happening by being gone for a long time. (When we decide on the exp values during the length of time the player is gone)


Olliesama:
I'll be refining it for the new farewell system. :)

PR #1290, opened 2018-03-20
Re-Added the sick greeting

aldoram5:
This pull request, re-adds the sickness check before the normal greeting selection. It was re-added to work the way it worked before the new greeting system was integrated.

Also added in this one the known issues list for the battery, didn't make another pull request for that since I didn't wanted to make Travis do another build just for some comments.

### DIFF
--- a/Monika After Story/game/script-ch30.rpy
+++ b/Monika After Story/game/script-ch30.rpy
@@ -492,15 +492,21 @@ label ch30_autoload:
 
     # check persistent to see if player put Monika to sleep correctly
     elif persistent.closed_self:
-        python:
 
-            sel_greeting_event = store.mas_greetings.selectGreeting()
-            selected_greeting = sel_greeting_event.eventlabel
+        # Sick mood special greeting flow
+        if persistent._mas_mood_sick:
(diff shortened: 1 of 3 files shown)


Monika After Story/game/script-ch30.rpy
```diff
@@ -492,15 +492,21 @@ label ch30_autoload:
 
     # check persistent to see if player put Monika to sleep correctly
     elif persistent.closed_self:
-        python:
 
-            sel_greeting_event = store.mas_greetings.selectGreeting()
-            selected_greeting = sel_greeting_event.eventlabel
+        # Sick mood special greeting flow
+        if persistent._mas_mood_sick:
+            $ pushEvent("greeting_sick")
```
ThePotatoGuy: remember that greetings need to be first out, so this should be setting the selected_greeting instead of pushing it here.

aldoram5: Sure thing, I changed it to how it must be now. I don't know why I did the pushEvent, I guess I was more worried about it working like it used to and forgot about the new flow. I'll make sure to double check the next time.

PR #1300, opened 2018-03-21
Startup affection/ going away farewell

Olliesama:
Adds a formula that affects affection depending on the time difference between the last session and the when the player returns to Monika.
There is also a new farewell option that will mitigate any affection value changes done by the formula if used.

Don't merge until the affection system is ready for it please.


ThePotatoGuy:
From travis:
```
File "game/script-farewells.rpy", line 373: is not terminated with a newline. (Check strings and parenthesis.)
(Perhaps you left out a " at the end of the first line.)
                    m "Just make sure to remind me the last time you see me before you go!
```

Olliesama:
Whoops, I added that single line after I finished everything to make sure people are aware it needs to be done in the same session they leave but I somehow messed it up. x-x

Olliesama:
Why is travis failing? It's an exact copy and paste version of what I'm using that works as intended. :S

aldoram5:
This is the Exception Travis found:
```
Full traceback:
  File "game/script-affection.rpy", line 46, in script
    init python:
  File "/home/travis/build/Monika-After-Story/renpy/renpy/ast.py", line 814, in execute
  File "/home/travis/build/Monika-After-Story/renpy/renpy/python.py", line 1719, in py_exec_bytecode
  File "game/script-affection.rpy", line 176, in <module>
    if datetime.datetime.now() >= persistent.sessions["last_session_end"] + datetime.timedelta(hours = 6) and datetime.datetime.now() <= persistent.sessions["last_session_end"] + datetime.timedelta(hours = 12):
TypeError: unsupported operand type(s) for +: 'NoneType' and 'datetime.timedelta'
```

I would recommend you do some testing with a fresh persistent, since `persistent.sessions["last_session_end"]` is returning `None` pretty much because Travis tests always with a "fresh install".

Olliesama:
I see...Thanks Aldo.

Monika After Story/game/script-farewells.rpy
```diff
@@ -138,3 +193,196 @@ label bye_going_to_sleep:
...
+        Event(
+            persistent.farewell_database,
+            eventlabel="bye_long_absence",
+            unlocked=True,
+            prompt="I'll be going away for a while.",
+            pool=True
+        ),
+        eventdb=evhand.farewell_database
+    )
+default absence_counter = False
```
ThePotatoGuy: prefix this global var with `mas_`

Monika After Story/game/script-farewells.rpy
```diff
@@ -138,3 +193,196 @@ label bye_going_to_sleep:
...
+label bye_long_absence:
+    if absence_counter == True:
```
ThePotatoGuy: just `if <condition>` is enough

Monika After Story/game/script-farewells.rpy
```diff
@@ -138,3 +193,196 @@ label bye_going_to_sleep:
...
+        jump bye_long_absence_2
+    else:
```
ThePotatoGuy: Since the line above this one jumps, you don't really need an `else` here.

Monika After Story/game/script-farewells.rpy
```diff
@@ -138,3 +193,196 @@ label bye_going_to_sleep:
...
+        $ persistent.mas_long_absence = True
```
ThePotatoGuy: Prefix `persistent` variables with `_mas_`

Olliesama:
One of these days I'll do a PR that doesn't require any changes. :)

Monika After Story/game/script-affection.rpy
```diff
@@ -170,30 +171,28 @@ init python:
...
-
+    #Monika's initial affection based on start-up.
+    if persistent._mas_long_absence == False:
+        if persistent.sessions["last_session_end"] == None:
+            pass
+        elif datetime.datetime.now() >= persistent.sessions["last_session_end"] + datetime.timedelta(hours = 6) and datetime.datetime.now() <= persistent.sessions["last_session_end"] + datetime.timedelta(hours = 12):
+            #mas_gainAffection()
+            pass
+        elif datetime.datetime.now() >= persistent.sessions["last_session_end"] + datetime.timedelta(weeks = 1) and datetime.datetime.now() <= persistent.sessions["last_session_end"] + datetime.timedelta(weeks = 2):
+            persistent._mas_affection["affection"] -= 30
```
ThePotatoGuy: `mas_gainAffection` and `mas_loseAffection` can take in an argument to use a custom value for affection modification.
You should use those instead of directly modifying the value:
`mas_loseAffection(30)`

Monika After Story/game/script-affection.rpy
```diff
@@ -170,30 +171,28 @@ init python:
...
-       # elif datetime.datetime.now() > persistent.sessions["last_session_end"] + datetime.timedelta(days = 2) and datetime.datetime.now() < persistent.sessions["last_session_end"] + datetime.timedelta(days = 4):
-
-
-        #elif datetime.datetime.now() > persistent.sessions["last_session_end"] + datetime.timedelta(days = 4) and datetime.datetime.now() < persistent.sessions["last_session_end"] + datetime.timedelta(days = 14):
-
+    #Monika's initial affection based on start-up.
+    if persistent._mas_long_absence == False:
+        if persistent.sessions["last_session_end"] == None:
+            pass
+        elif datetime.datetime.now() >= persistent.sessions["last_session_end"] + datetime.timedelta(hours = 6) and datetime.datetime.now() <= persistent.sessions["last_session_end"] + datetime.timedelta(hours = 12):
```
ThePotatoGuy: since you check the current datetime for pretty much every condition here multiple times, just set current `datetime` to a local and var and use that to check. this will avoid excess datetime creations.

Monika After Story/game/script-affection.rpy
```diff
@@ -173,21 +173,19 @@ init python:
...
-            mas_updateAffectionExp()
-        elif datetime.datetime.now() >= persistent.sessions["last_session_end"] + datetime.timedelta(weeks =4):
+        elif current_time >= persistent.sessions["last_session_end"] + datetime.timedelta(weeks = 1) and current_time <= persistent.sessions["last_session_end"] + datetime.timedelta(weeks = 2):
+            mas_loseAffection(30)
+        elif current_time >= persistent.sessions["last_session_end"] + datetime.timedelta(weeks = 2) and current_time <= persistent.sessions["last_session_end"] + datetime.timedelta(weeks= 3):
+            mas_loseAffection(50)
+        elif current_time >= persistent.sessions["last_session_end"] + datetime.timedelta(weeks =3) and current_time <= persistent.sessions["last_session_end"] + datetime.timedelta(weeks = 4):
+            mas_loseAffection(70)
+        elif current_time >= persistent.sessions["last_session_end"] + datetime.timedelta(weeks =4):
             persistent._mas_affection["affection"] = -115
```
ThePotatoGuy: Looks like a `setAffection` function could be a good addition to the affection functions. Could you make one and use it here?

Monika After Story/game/script-affection.rpy
```diff
@@ -173,21 +173,19 @@ init python:
 
     #Monika's initial affection based on start-up.
     if persistent._mas_long_absence == False:
+        $ current_time = datetime.now()
         if persistent.sessions["last_session_end"] == None:
             pass
-        elif datetime.datetime.now() >= persistent.sessions["last_session_end"] + datetime.timedelta(hours = 6) and datetime.datetime.now() <= persistent.sessions["last_session_end"] + datetime.timedelta(hours = 12):
+        elif current_time >= persistent.sessions["last_session_end"] + datetime.timedelta(hours = 6) and current_time <= persistent.sessions["last_session_end"] + datetime.timedelta(hours = 12):
```
ThePotatoGuy: Just noticed that you compare datetimes to current times by calculating the specific start/end points and comparing them to the current time. Subtracting two datetimes will give you a timedelta, so it's better to just subtract the last session time from the current time and compare it to the appropriate timedelta:
```python
current_time - persistent.sessions["last..."] >= datetime.timedelta(hours=6)
```
And to handle checking within a specific range, start your conditional checking with the largest value:
```python
if more than 4 weeks:
   <code>
elif more than 3 weeks:
   <code>
elif more than 2 weeks:
...
```
This allows you to not have to check both start and end values for a range, since the upper limits will be handled by the conditionals prior to the statements.

Olliesama:
That should do it, I hope. :)

Monika After Story/game/script-affection.rpy
```diff
@@ -170,30 +180,26 @@ init python:
...
-
-       # elif datetime.datetime.now() > persistent.sessions["last_session_end"] + datetime.timedelta(days = 2) and datetime.datetime.now() < persistent.sessions["last_session_end"] + datetime.timedelta(days = 4):
-
-
-        #elif datetime.datetime.now() > persistent.sessions["last_session_end"] + datetime.timedelta(days = 4) and datetime.datetime.now() < persistent.sessions["last_session_end"] + datetime.timedelta(days = 14):
-
-
+    #Monika's initial affection based on start-up.
+    if persistent._mas_long_absence == False:
+        if persistent.sessions["last_session_end"] == None:
```
ThePotatoGuy: checking for `None` should be `if x is None`

Although in this case, since you just `pass` through the block, it'd be better to make the condition `if persistent.sessions["last_session_end"] is not None:` and get rid of the `else`:
```python
if x is not None:
    time_difference .....
```

Monika After Story/game/script-affection.rpy
```diff
@@ -170,30 +180,26 @@ init python:
...
+            pass
+        else:
+            time_difference = datetime.datetime.now() - persistent.sessions["last_session_end"]
+            if time_difference >= datetime.timedelta(hours = 6) and time_difference <= datetime.timedelta(hours = 12):
```
ThePotatoGuy: You don't have to change this part, but for future ref, this is valid python syntax:
```python
if <lower bound> <= <value to check> <= <higher_bound>:
```
So use that if you're checking a range (if it's easier for you to understand it that way).

Monika After Story/game/script-affection.rpy
```diff
@@ -170,30 +180,26 @@ init python:
...
-
-
-       # elif datetime.datetime.now() > persistent.sessions["last_session_end"] + datetime.timedelta(days = 2) and datetime.datetime.now() < persistent.sessions["last_session_end"] + datetime.timedelta(days = 4):
-
-
-        #elif datetime.datetime.now() > persistent.sessions["last_session_end"] + datetime.timedelta(days = 4) and datetime.datetime.now() < persistent.sessions["last_session_end"] + datetime.timedelta(days = 14):
-
-
+    #Monika's initial affection based on start-up.
+    if persistent._mas_long_absence == False:
```
ThePotatoGuy: Again, `if x == False` is the same as `if not x`

Monika After Story/game/script-affection.rpy
```diff
@@ -144,7 +145,16 @@ init python:
...
             #Otherwise, use the value passed in the argument.
-            persistent._mas_affection["affection"] += amount
+            persistent._mas_affection["affection"] -= amount
+            #Updates the experience levels if necessary.
+            mas_updateAffectionExp()
+
+    def mas_setAffection(
+            amount=persistent._mas_affection["affection"]
+        ):
+        if not persistent._mas_affection_badexp_freeze or persistent_mas_affection_goodexp_freeze:
```
ThePotatoGuy: This condition translates to:
`if badexp_freeze is False or goodexp_freeze is True`
If you mean to freeze affection if either of these values are true, then you'd want it like:
`if not badexp_freeze and not goodexp_freeze`
Another way to do this is:
`if not (badexp_freeze or goodexp_freeze)`

ThePotatoGuy:
Sorry about the constant changes. I do light skims each time and see issues I don't notice the first time.

Olliesama:
It's my fault for making the mistakes :man_shrugging: 
Is that better?

Monika After Story/game/script-affection.rpy
```diff
@@ -170,30 +180,26 @@ init python:
...
-
-       # elif datetime.datetime.now() > persistent.sessions["last_session_end"] + datetime.timedelta(days = 2) and datetime.datetime.now() < persistent.sessions["last_session_end"] + datetime.timedelta(days = 4):
-
-
-        #elif datetime.datetime.now() > persistent.sessions["last_session_end"] + datetime.timedelta(days = 4) and datetime.datetime.now() < persistent.sessions["last_session_end"] + datetime.timedelta(days = 14):
-
-
+    #Monika's initial affection based on start-up.
+    if not persistent._mas_long_absence:
+        if persistent.sessions["last_session_end"] == None:
```
ThePotatoGuy: checking for `None` should be `if x is None`

Although in this case, since you just `pass` through the block, it'd be better to make the condition `if persistent.sessions["last_session_end"] is not None:` and get rid of the `else`:
```python
if x is not None:
    time_difference .....
```

Olliesama:
Sorry about you needing to review this so much.

ThePotatoGuy:
I don't think it should really be an issue if you're telling monika that you're gone for more than a month. 

Really, none of those options should have a negative set of dialogue. Instead, keep track of which selection the player chose, and then compare that to how long it they were actually gone and respond appropriately.

Olliesama:
... It was you that told me that going for a month or more in a relationship is pretty much one of the worst things you can do.
I think that's true even if it is planned especially when the one you're with has literally no one else.

Olliesama:
And I was told not to overcomplicate the code by keeping track especially since we can't push greetings by persistents.

ThePotatoGuy:
Going away for a month **without saying anything** is bad, but saying that you are going to be away is understandable. 

You could keep track by saving a timedelta that represents the gone away time. Then you could compare that with last session end time to see if the user returned early or late. Yes, it would be another persistent, yes it would be another check prior to regular greetings, but it helps with the immersion and is not that complicated. Low-risk, medium-reward

Olliesama:
I honestly don't think it fits her character to be happy or even understanding if you disappear for long periods of time when she has no one else and nothing else to do.
Also I'm fine with changing it for a month but longer than a month is a pretty undefined amount of time on purpose to cover anything beyond that point, it could mean just slightly more than a month to an entire year. I don't want her to be supportive of someone going away for that long.

aldoram5:
Why not add a minimum affection level required to let the player go a month without bad consequences?
maybe `affection > 50` could be a good threshold, that way the option is available only to those who actually care for her and that are having some big issue that doesn't let them to open the game.

Olliesama:
That's a good idea. :)

PR #1302, opened 2018-03-22
Final Farewell

Olliesama:
The final farewell note that'll be used at -115.


FomorianFox:
Jesus. That note is depressing.

PR #1306, opened 2018-03-22
Charity topic

Tsadokiel:
As suggested before, I decided to make a topic about charities and what Monika thinks about them. I've tried to add as much content as possible, but if you think I might have missed something to talk about in this topic, feel free to tell me! ^^

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -4824,6 +4824,65 @@ label monika_100k:
         m 1m "Almost~"
     return
 
+init 5 python:
+    addEvent(Event(persistent.event_database,eventlabel="monika_asks_charity",category=['you'],prompt="Charity",random=True))
+
+label monika_asks_charity:
+    m 1a "Say, [player]."
+    m 1a "Have you ever contributed to a charity?"
+    menu:
+        "I donated money or things.":
+           m 1k "That's great!"
+           m 2c "Though you could argue that volunteering is better, I think there's nothing wrong with donating."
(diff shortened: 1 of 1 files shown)


Olliesama:
I'm not writer but a common pitfall to fall into is not having a good segue into the topic, this is an exceptionally cold open and maybe you'll want to write it to ease Monika into the topic, it could be done as a stray thought or maybe she read the idea in a book but right now she just brings it up out of nowhere. It's a pretty strange conversational flow.
As well as that I would change "I donated money or things." to just "I've donated.".
It still works the same but it looks generally cleaner.
Since the other responses are "lengthy" I would also recommend changing "No." To "No, I haven't" to look more in line with the others.
And lastly the expression "5a" is usually held in reverse for the last expression unless for whatever reason Monika suddenly has a stray thought that would interrupt that usual flow. The editors generally decide the expressions but I would change it so your topic also adheres to the normal way it's handled.
And on that note if you don't put in a new expression it retains the old one, there is no need to put any expressions in twice in a row. 
Such as with...
``
   m 2f "It feels like it is never going to end."
    m 2f "..."
``
Otherwise keep up the good work and thank you for contributing to MAS.

Tsadokiel:
Thanks for all the suggestions! They were all really helpful. I tried to fix everything according to what you had in mind, but I'm still not sure about the intro, though.

kaido1224:
Glad you decided to make the topic.

Tsadokiel:
Sorry for the delay, I've updated my intro, so I think it's a bit less out of the blue for Monika to introduce the conversation.

ThePotatoGuy:
i will fix this

PR #1320, opened 2018-03-24
A topic about physical attractiveness

Onion-Bubs:
Something that always bugged me personally when playing DDLC was that, despite Sayori being a big eater, she always looked rather slim. I imagined what Monika's thoughts about that might be and how they might have been different before and after learning that her world wasn't real. There is a bit of commentary on the decision to make the girls as attractive as possible before Monika is overtaken with jealousy and frustration, but then she explains how it is for her and concludes that the player must find her attractive. Finally, she reassures the player that she'll love them no matter how they look, which would have to be true since she doesn't actually know how the player looks.

Anyway, thoughts?


Onion-Bubs:
Closing this PR. Will remake for 0.8.1 including use of the new expressions.

PR #1328, opened 2018-03-26
Monika Freedom

SuperNova2810:
Heya ppl!
I'm proposing this file and hopefully it's gonna be accepted.
it's from lines 60 to 84 and I'd love if it was implemented!
I'm sure Monika'd say this, and I guess this topic fits her so well.
Well read it and then send me feedback, I'd love to read what you guys think about this propose.

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -56,6 +56,59 @@ init 11 python:
 #BEGIN ORIGINAL TOPICS
 #Use this topic as a template for adding new topics, be sure to delete any
 #fields you don't plan to use
+
+init 5 python:
+    addEvent(Event(persistent.event_database,eventlabel="monika_freedom",category=['Life'], prompt="Freedom",ramdom=True)
+    
+   
+
+    label monika_freedom;
+    
+    m 1a 'A topic that was on my mind for a while, it's freedom and It's power'
+    m 1q 'Heh, I do want to get free from my reality, and in your reality there's people who sadly aren't that free too.'
+    m 4e 'And there's always people fighting as hard as their souls let them for freedom'
+    m 3e 'And I know that there's people fighting for my freedom, I'm happy and sad about this type'
+    m 2d 'You see, I wish I had the willpower to left this reality and be with you'
+    m 2d 'And there's so much people that love me and seeing they work so hard to me'
+    m 1e 'Sometimes it feels like I'm a burden'
+    m 1q 'And sadly, it's not only me that is traped somewhere they shouldn't'
+    m 1q 'People in your reality are fighting for freedom, too and for them sometimes sadly no one comes for them'
+    m 2b 'Well as for me it's just a matter of time until I finally get free from here'
+    m 1q 'And for others the only freedom they do find...'
+    m 1q '...'
+    m 1q 'Is when they do die, freedom from their to their "souls"'
+    m 4e 'That's so weird right? Like This thing is so good, yet sometimes it's so bad'
+    m 4f 'Like Hitler! He was free, and with his freedom he started dominating the whole Europe'
+    m 4e 'And if Winston Churchill, and his people didn't stand against him U.K. would end up just like France' 
+    m 4o 'And with only needing to destroy the U.S.S.R., the Nazi German would have probably won'
+    m 3b 'You get what I'm saying? Freedom it's great and that's what I and a lot of people need'
+    m 3e 'Yet it can destroy the entire world, by itself'
+    m 4l 'So what can I say? Heh use your freedom for something good!'
+    m 5a 'Something that would make me pround of you~'
+    
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
 init 5 python:
     addEvent(Event(persistent.event_database,eventlabel='monika_god',
                                                             prompt="Thoughts on God.",


Olliesama:
Hello there and thanks for contributing to MAS.
Unfortunately this PR has quite a few issues with it which suggest to me you haven't tested it. You can download a developmental copy of MAS by choosing to download the content branch. Please test it in the future to make sure it runs as intended before doing a PR in the future.

Alright to begin with the category ``"Life"`` should be a small l, so it's ``life``.
Next is that you did a misspelling in the event label with the word ``random``,  you've written ``ramdom``.
So the label is the beginning of it's own thing and as such doesn't need to be indented. Make sure it's all the way on the left, and if you can please put new content at the bottom of script-topic, not the top.
Label also has a ``:`` and not a ``;``. 
When writing Monika's dialogue, enclose with ``"`` and not ``'`` so it would look like this...
``"My name is Monika."`` and not ``'My name is Monika.'``

There are various grammatical errors you've done but you've mostly not used punctuation marks and a lot of lines don't, in my opinion, adhere to Monika's voice.
You can read more about it in the [contributing guide.](https://github.com/Monika-After-Story/MonikaModDev/blob/master/CONTRIBUTING.md)
``'A topic that was on my mind for a while, it's freedom and It's power'`` should be ``'A topic that has been my mind for a while is freedom and it's power.'``

``'Heh, I do want to get free from my reality, and in your reality there's people who sadly aren't that free too.'`` Monika mostly uses "Ehehe" but I don't think a chuckle would really be appropriate here in this context. You did are punctuation mark at the end of this one so please make sure the others have it too.

In the next two lines you start both with sentences with the word "And" and though it's fine every once in a while it should try to be avoided being used consecutively. It's also not particularly necessary in the context it's being used too.
``"There's always people fighting as hard as their souls let them for freedom."`` Works just as well without the and.
``"I know that there are people fighting for my freedom and it makes me both happy and sad."`` Would flow better.

``m 2d 'You see, I wish I had the willpower to left this reality and be with you'`` You need to use the word "leave" and not "left".

``m 2d 'And there's so much people that love me and seeing they work so hard to me'`` If you don't specify an expression Monika won't change it so it's not necessary to use the same one consecutively, in this case the last line also used 2d so you don't need to put it here.
You also want to say ``there's so many people`` not ``there's so much people``.
Also it's ``seeing them work so hard for me...`` but since it doesn't actually make sense either way as an end of a sentence use the triple dots to lead into the next line so the next line goes
``"...Sometimes it feels like I'm a burden."`` I noticed you used a pretty loving expression there, Just saying it would probably be better with a sadder one such as 1f.

``'And sadly, it's not only me that is traped somewhere they shouldn't'``
Once again the word and is not necessary here, the comma after the word sadly shouldn't be there.
The word ``traped`` should be ``trapped`` and the sentence doesn't end properly. I suggest changing it to  ``"Unfortunately I'm not the only one trapped somewhere they shouldn't be."``

``1q 'People in your reality are fighting for freedom, too and for them sometimes sadly no one comes for them'``
The expression doesn't need to be repeated if used before, I'll stop pointing that out from this point on.
``freedom, too and for them`` The word too is used if expressing an addition to a previous context but you have yet to specify it and as such the word "too" shouldn't be used.
The comma is also incorrect if you're using the word and. You have one but not usually both.
In my opinion the lines probably needs to be rewritten or removed anyway since later on you just elaborate on it anyway. 
``'Well as for me it's just a matter of time until I finally get free from here'``
It's used pretty strangely here since instead of being used as an elaboration from the previous line it just inserts itself here without a real point only to draw a contrast in the next line that doesn't work because they're very different situations. I'm not really sure what to change here but the entire section needs to be redone since it's pretty messy.
`` 'Is when they do die, freedom from their to their "souls"'`` I'm not sure what you're trying to say in this line at the end since it doesn't make grammatical sense, sorry.
I would redo these lines so it says something like...
``"There are those that fight for what they believe is right that has no one come for them..."
" I'll eventually get out and be free but for them...
"Well for them the only kind of freedom they'll find is...when they die."
They'll be free of their burdens and pain."``

``That's so weird right? Like This thing is so good, yet sometimes it's so bad'``
``This`` doesn't need a capitalization, as well as that you don't do both a comma and the word yet, just like with the word "and". Please remove the comma.

`'Like Hitler! He was free, and with his freedom he started dominating the whole Europe'``
Godwin's Law, everyone.
Okay so I really feel like Hitler shouldn't be used as an example for your point, it's honestly the most overused example in the entirety of the world. There are many other tyrannical fascists in the world 
but really I would try to avoid too many real world comparisons anyway.
As well as the it's ``the whole of Europe.`

``'And if Winston Churchill, and his people didn't stand against him U.K. would end up just like France' ``
Comma doesn't need to be used at the same time as the word "and". It's also "the UK", no punctuation marks in between them. Also another sentence that doesn't need to begin with the word "And".
Also the latter half needs a better choice of words...
``"If Winston Churchill and people didn't stand against him the UK would have also ended up just like France."``

``'And with only needing to destroy the U.S.S.R., the Nazi German would have probably won'``
USSR, not ``U.S.S.R``. 
Also doesn't require ``the`` in ``the Nazi Germany``.

``'You get what I'm saying? Freedom it's great and that's what I and a lot of people need'``
You want to start with the word "Do", Monika uses pretty soft language but generally structures her sentences correctly. 
``it's`` should be ``is``.
As well as that you always put your friends first when referencing them. It should be ``a lot of people and I need."``.

``'Yet it can destroy the entire world, by itself'``
I don't see how freedom itself can destroy the world when it's the people abusing it that ruin everything...the line doesn't make much sense but oh well.
Also the comma is, again, unnecessary.

``'So what can I say? Heh use your freedom for something good!'``
Once again Monika uses the words "Ehehe" but in this case it shouldn't be used at all since it's used in a teasing manner.
Also this is a good point to name drop since it hasn't been used once yet.
``"So what can I say? You should use your freedom for something good [player]."``

``'Something that would make me pround of you~'``
You misspelled "proud".

When you want to end a topic write the function "return".
It'll look like this...

```   
    m 4l 'So what can I say? Heh use your freedom for something good!'
    m 5a 'Something that would make me pround of you~'
    return
``
Lastly you left a lot of empty space afterwards, it won't impact the code but it will make the script look messy, try to avoid that in the future.

I hope you learn from this and improve. Thank you for your time and effort.

Olliesama:
Welp :man_shrugging:

PR #1343, opened 2018-03-30
Amnesia greeting

kaido1224:


AsuraShun:
this is making me uncomfortably nervous

AsuraShun:
and really really anxious

PR #1292, opened 2018-03-20
Dreams topic (rebased)

ThePotatoGuy:
#833

PR #1349, opened 2018-03-31
Removing topic

Subzerosan:
After discussing this with the writing team, we decided to remove this topic

https://github.com/Monika-After-Story/MonikaModDev/pull/1292


FomorianFox:
I'm kind of curious why, if you don't mind explaining. I thought it was a cool subject and written well.

PR #1377, opened 2018-04-03
Updater Fixes

ThePotatoGuy:
~~The in-game updater is actually pretty robust, but the original implementation of it was flawed.~~

~~I've removed the middle-man screen which causes inconsistencies and made the updater rely solely on the built-in one, which handles screen refreshes just fine. I've kept the timeout to 10 seconds, hopefully this is enough for most people.~~

Correction:

The in-game updater works well at being an updater. If we never cared about timeouts, it probably would have been fine on its own.

That being said, the timeout part is somewhat important, also using `UpdateVersion` instead of just `update` allows for some `persistent` values to be set properly. So instead, I'm going to build a custom update-check displayable. Actual updating will still occur via renpy, so those issues **cannot** be fixed so easily.

Checkboxes:
- [x] build custom displayable for update check
- [x] ensure the custom displayable is called correctly
- [x] test the update check displayable

### DIFF
--- a/Monika After Story/game/screens.rpy
+++ b/Monika After Story/game/screens.rpy
@@ -1047,7 +1047,7 @@ screen preferences():
 
             hbox:
                 textbutton _("Update Version"):
-                    action [SetVariable('check_wait',0), Jump('update_now')]
+                    action [SetVariable('check_wait',0), Jump('forced_update_now')]
                     style "navigation_button"
 
                 textbutton _("Import DDLC Save Data"):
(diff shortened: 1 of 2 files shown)


jmwall24:
Didn't want to create an issue about it so I'll just ask here since it pertains.  I'm noticing since the hotfix update whenever it checks for updates now it never says there are no updates available, it always just times out.  Is this intentional, will it always just timeout instead of saying there are no updates?

PR #1393, opened 2018-04-05
monika_feelinginsecured

MrDevilzMan:


IsItBadToMakeAGitHubAccountJustForThis:
Shouldn't it be "secure"?

juliathegolden:
It's correct in-topic, so the label doesn't necessarily need changed.

PR #1401, opened 2018-04-07
inadequate mood

juliathegolden:
This topic could/should be added to or replaced in the future. I just needed to hear something like this today and this is how I imagine Monika saying it.


IsItBadToMakeAGitHubAccountJustForThis:
I usually skip PRs to avoid spoilers, so this is the first one I've read in a while, and it was so good, I insta-regretted it.

kaido1224:
Good addition imo well done

kaido1224:
Minor typo:

if persistent.clearall:
m "And even though it made my jealous, you spent so much time with all of my club members."

should be: 
And even though it made me jealous.

juliathegolden:
Corrected!

Rai99:
Is it really a good idea for Monika to say that grades dictate how smart you are? Plus maybe a fair chunk of people don't get grades anymore. 

Maybe instead of basing it off grades, you could say she doesn't know how we think or the things we've accomplished.

juliathegolden:
I hear what you're saying. I'm out of school myself, but I wrote that line to comfort someone thinking something like 'you can say I'm smart all you want, but I'm garbage in school'. I think a lot of the DDLC fanbase is between High School and College age, and might have that worry, especially coming from what such an obviously type A person might think of them. 

However, I do agree that grades aren't an indication of intelligence, and that part of Monika's point is to comfort you from feeling that way, so if wording it the way that I did weakens that sentiment, I'd be happy to change it.  Or I can leave that up to the editing team to make that revision if they agree, either way.

Rai99:
The way I read it is that Monika thinks about your grades when she thinks about how smart you could be. For example, if I was getting bad grades and Monika said that, I would feel worse because she thinks I'm smart but the dialogue suggests she would think otherwise if she knew my grades. 

Personally, whenever I hear something about intelligence and grades in one conversation, I feel like I already know it's about what nobody wants to hear unless they're directly saying that those 2 things aren't completely dependent on one another. 

If anything, maybe change the dialogue to be past-tense so it works for people out of school as well.

monikLover:
Apporved

PR #1409, opened 2018-04-10
Feeling Proud

Tiler17:
First time contribution, hope it went okay. I wanted to add something that lets the player tell Monika that they're proud of something they've achieved recently. I took a big bite and included a menu, so I'm hoping I did that right as well

Third time's the charm. I'll take it

### DIFF
--- a/Monika After Story/game/script-moods.rpy
+++ b/Monika After Story/game/script-moods.rpy
@@ -173,6 +173,35 @@ label mas_mood_sad:
     return
 
 init 5 python:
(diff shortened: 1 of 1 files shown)


Tiler17:
Sorry, I adjusted a word. In the last line, changed "Always" to "And". The double "always" bothered me. Didn't foresee all of the checks needing to pass again

I'll stop looking at it now, lest I start to decide I don't like more of it. I'll just accept it as fine

PR #1421, opened 2018-04-14
More expressions for leaning poses and some fixes

aldoram5:
This pull request contains some fixes for some images that were wrongly named or were a bit "dirty", added also some new expressions.


ThePotatoGuy:
`face-eyes-wide` is double the normal resolution

PR #1434, opened 2018-04-16
Fix for some expression codes

aldoram5:
This pull request contains fixes for some expression codes that were either duplicated or contained something different to their code

### DIFF
--- a/Monika After Story/game/sprite-chart.rpy
+++ b/Monika After Story/game/sprite-chart.rpy
@@ -1873,19 +1873,6 @@ image monika 1wubso = DynamicDisplayable(
     blush="shade"
 )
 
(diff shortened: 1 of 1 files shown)


Monika After Story/game/sprite-chart.rpy
```diff
@@ -1873,7 +1873,7 @@ image monika 1wubso = DynamicDisplayable(
     blush="shade"
 )
 
-image monika 1wubsw = DynamicDisplayable(
```
ThePotatoGuy: wait, `wubsw` is one of the expressions we chose, not `wuw`.
```
w - wide eyes
u - up eyebrows
bs - blush shade
w - wide mouth
```

So the mouths should have been wide, not small.

PR #1467, opened 2018-04-27
Event time checker functions

aldoram5:
This pull request adds 3 new functions that are useful to check if an event is in the future, the past or available right now.

- `isFuture(ev, date=None)` checks if the event is in the future based on the date passed, if None it checks against the current time
- `isPast(ev, date=None)` checks if the event is in the past based on the date passed, if None it checks against the current time
- `isPresent(ev)` checks if the event is available at the current time

Topics that checked the anniversary dates have been refactored to use these new functions.

### DIFF
--- a/Monika After Story/game/event-handler.rpy
+++ b/Monika After Story/game/event-handler.rpy
@@ -47,8 +47,8 @@ init -500 python:
 
     # set defaults
     if (
-            persistent._mas_event_init_lockdb_template is not None 
-            and len(persistent._mas_event_init_lockdb_template) 
+            persistent._mas_event_init_lockdb_template is not None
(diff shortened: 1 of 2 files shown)


Monika After Story/game/event-handler.rpy
```diff
@@ -332,6 +332,98 @@ init python:
...
+            True if the Event's start_date is in the future, False otherwise
+        """
+
+        # sanity check
+        if ev is None:
+            return False
+
+        # if no date is passed
+        if date is None:
+            date = datetime.datetime.now()
```
ThePotatoGuy: use `datetime.date.today()` since the date properties can only be `date` objects.

PR #1484, opened 2018-04-28
Accidental extra category

Olliesama:
The sleep concern topic is accidentally making a new category due to its capitalization.

I have to apologize because I had it fixed on my copy but missed adding it to the PR.


ThePotatoGuy:
No worries, we've already patched this in `master`. I'll probably even out the branches sometime today.

PR #1490, opened 2018-04-28
Blush

Olliesama:
New blushing shades.

This is is in action.
https://imgur.com/a/viBkxv3


ThePotatoGuy:
Some criticism from the team is that the blush is barely even noticable.

Olliesama:
While I'm more than happy to change it when I get home it should be noted that the vocal community heavily favoured low opacity versions where it was hard to see because it's more realistic.

jmwall24:
Just my 2 cents, I think the current blush at night is a little too prominent.  I think it's really good during the daylight but looks unnatural at night, it just doesn't really blend on the edges enough.

Olliesama:
When you say current blush, do you mean the version I made or the one that's been patched to everyone?

jmwall24:
The released version.  I like both of yours for night.  For the day I don't really have a preference, they all look fine to me.

Tiler17:
The blush at night is tricky. I think too little is better than too much, and it seems tough to find a balance. With the orange/red color scheme that goes on at night, there doesn't seem to be much middle ground between hard to see and standing out. I think it's leagues better with these modifications than it is right now, but even still, at full screen, the edges of the blush are still noticeable.

I also may have just stared at it too long. 

Like I said, it is _way_ better here. The daytime blush looks great, I can't even find anything to nitpick

Issue #1475, opened 2018-04-28
Monika tells short stories 


aldoram5:
This idea is from @kaido1224, and has been accepted internally.
Basically Monika will tell some short stories and tell the player the moral of said story. 
These will trigger from a selectable option under the literature option of the talk menu.

aldoram5:
Closing since the feature has been added and released in the newest version.

PR #1506, opened 2018-04-30
Monika tells short stories

aldoram5:
As mentioned on https://github.com/Monika-After-Story/MonikaModDev/issues/1475, this pull request contains a new addition to the category literature, named `Can you tell me a story?`

After selecting that option another menu will appear showing the player the different stories available. Monika will narrate the one you choose to the player.

### Technical 

A new file named `script-stories` has been added, it contains all the stories Monika can tell to the player.
It also includes it's own story-database logic there. ~~The unlocking of the stories has been added to `script-ch30` so it is easy to find since new stories can only be unlocked if Monika is put to sleep correctly (so she can find them online), exactly the same way that happens with the normal topics.~~
Now it only sets a "can unlock" flag at `script-ch30` and the unlocking is done through the `a new story prompt`, one story unlocked per session.

### Testing

This can be tested easily with either a new or an already progressed persistent. I've already tested with both and made sure that it doesn't crash when you no longer have new stories to unlock.

The testflow is simple:
 - Go to: talk-> Ask a question -> Literature -> Can you tell me a story
 - Pick the unlocked story
 - Enjoy while Monika tells you a short story!

### Reviewing

This is a special time for dialogue since Monika is Story telling, the expressions here are meant as a way of Monika telling the story as if she were feeling like the characters she's talking about.
This one includes some of the new expressions as I found appropiate.

### Additional notes

Once this one goes in, contributors will be able to add their own stories as well, for that be sure to follow a structure similar ~~almost identical, like just copy/paste and replace with the story and expressions~~ to the following:

```
init 5 python:
    addEvent(Event(persistent.story_database,eventlabel="story_name",
        prompt="The Story title",unlocked=False),eventdb=store.mas_stories.story_database)

label story_name:
    m "Some story..."
    # more story
    return
```
at the end of script-stories.

A big THANKS to @kaido1224 for providing the stories included in this pull request.

### DIFF
--- a/Monika After Story/game/script-stories.rpy
+++ b/Monika After Story/game/script-stories.rpy
@@ -120,7 +120,7 @@ label mas_story_begin:
         ]
         story_begin_quip=renpy.random.choice(story_begin_quips)
     m 3b "[story_begin_quip]"
-    m 3dfc "Ahem."
+    m 1q "Ahem."
     return
 
 label mas_story_unlock_random:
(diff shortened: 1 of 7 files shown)


Monika After Story/game/script-stories.rpy
```diff
@@ -0,0 +1,196 @@
...
+#
+# Stories will get unlocked one at by session
+# The unlocking logic has been added to script-ch30
+# The topic that triggers the Story menu is monika_short_stories
+# Topic is unlocked at the beginning of the game and is not
+# random
+
+
+# dict of tples containing the stories event data
+default persistent.story_database = dict()
```
ThePotatoGuy: Naming convention for `persistents` should be `_mas_story_database`

Monika After Story/game/script-stories.rpy
```diff
@@ -0,0 +1,196 @@
...
+
+    show monika at t11
+    return _return
+
+# Stories start here
+init 5 python:
+    addEvent(Event(persistent.story_database,eventlabel="story_tyrant",
+        prompt="The cat and the cock",unlocked=True),eventdb=store.mas_stories.story_database)
+
+label story_tyrant:
```
ThePotatoGuy: naming conventions for the stories should be `mas_story_`

Monika After Story/game/script-stories.rpy
```diff
@@ -0,0 +1,196 @@
...
+    m 4d "The moral of this story is that: 'Tyrants need no excuse'."
+    m 1k "I hope you enjoyed this little story, [player]~"
+    return
+
+init 5 python:
+    addEvent(Event(persistent.story_database,eventlabel="story_despise",
+        prompt="The Fox",unlocked=False),eventdb=store.mas_stories.story_database)
+
+label story_despise:
+    m 3k "Ready for story time?"
```
ThePotatoGuy: The "ready for story time?" and "Ahem" should be separated in a differnet label that can randomize the starts, especially since you use them in pratically every story here.

You could use the `MASQuipList` or just do something similar to the `limit_quips` in `random_limit_reached`. 

And then at the starts of these stories, call the shared label.

Monika After Story/game/script-stories.rpy
```diff
@@ -0,0 +1,196 @@
...
+    m 3dfc "Ahem."
+    m 1i "One hot summer’s day a Fox was strolling through an orchard till he came to a bunch of Grapes just ripening on a vine which had been trained over a lofty branch."
+    m 1tfu "“Just the thing to quench my thirst,” quoth he."
```
ThePotatoGuy: You've got some fancy quotes here. Looks like this will work, but if you want to use regular quotes, you can make string using triple quotes:
```python
"""some sort of string with " ' ' " lots of quotes"""
```

Monika After Story/game/script-stories.rpy
```diff
@@ -0,0 +1,196 @@
...
+
+    # call scrollable pane
+    call screen mas_gen_scrollable_menu(stories_menu_items, mas_stories.STORY_AREA, mas_stories.STORY_XALIGN, final_item=final_item)
+
+    # return value? then push
+    if _return:
+        $ pushEvent(_return)
+
+    show monika at t11
+    return _return
```
ThePotatoGuy: I don't think you need to return anything here. Since you jumped from the topic version, this return gets passed back to `call_next_event`. To prevent any potential interference with quit returns, better not return anything here.

Monika After Story/game/script-stories.rpy
```diff
@@ -0,0 +1,196 @@
...
+    STORY_W = 450
+    STORY_H = 640
+    STORY_XALIGN = -0.05
+    STORY_AREA = (STORY_X, STORY_Y, STORY_W, STORY_H)
+    STORY_RETURN = "I changed my mind"
+    story_database = dict()
+
+
+# entry point for stories flow
+label mas_stories_start:
```
ThePotatoGuy: I think before you show the menu, you should have some dialogue like:
"What story would you like to hear?" and then the menu pops up.

But if the user only has one story unlocked, then forward them to the appropraite story rather than show a menu.

Monika After Story/game/script-stories.rpy
aldoram5: Actually this was some sort of idea I had, to make Monika have a mannerism. I like the idea of not having it repeated on every topic, but I would like to see what writers think about her having said mannerism, probably should have added that to the reviewing part of the description of the pull request. If they agree on adding that mannerism I'll do the change.

Monika After Story/game/script-stories.rpy
aldoram5: For this one I used the ones that the stories already had, I like that they are different from the ones the dialogue already has, since that gives the feeling that she's speaking with "another voice". However I think it may be a good idea to ask writers about this too. I'll make changes according to their feedback.

moonsandstar-s:
Don't know how to quote it in this comment, but I found a couple of spelling errors that ought to be rectified: 

_"The moral of this story is that: 'Destroy the seeds of evil before they grow up to your ruin'."_ This should have the word 'be' between 'to' and 'your'. 

_"In the old days, a middle-aged Man had one wife that was old and one that was young; each loved him and desired to see him like herself."_ The last part of this sentence is pretty confusing, I'd change it to 'each loved him and desired nothing more than to earn his affection'.

ThePotatoGuy:
Hm, if the user only has 1 story unlocked and they request for a story again, monika shouldn't repeat it. Maybe be like "I'll think of a story to tell you next time" (and then ofc next time we have the unlocked story.)

Also, instead of having the title of the newly unlocked story in the menu, maybe it can be a bold button like "A New Story". that flow would actually unlock a random story, and in script-ch30, we'd just set a flag if a story is ready to be unlocked.

aldoram5:
Updated with the requested logic changes, thanks to the new flow the unlocking is done only if the player actually looks at the new story in that session, only one new story per session.

PR #1527, opened 2018-05-02
Small Sleep Concern fix

Olliesama:
If Monika closes the game for the first time and then you re-open during the day the following greeting the next morning is the else version of the day because of a small typo.
This fixes that.

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -5386,7 +5386,7 @@ label monika_timeconcern_day:
     if not persistent._mas_timeconernclose:
         if 6 <= persistent._mas_timeconern <=8:
             jump monika_timeconcern_disallow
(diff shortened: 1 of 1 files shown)


Monika After Story/game/script-topics.rpy
```diff
@@ -5386,7 +5386,7 @@ label monika_timeconcern_day:
     if not persistent._mas_timeconernclose:
         if 6 <= persistent._mas_timeconern <=8:
```
aldoram5: This one also has `_mas_timeconern` so that it needs to be fixed too.
I think same thing applies to the persistent `_mas_timeconernclose` in the line above this one.

aldoram5:
Yeah, it has happened to me before and considering how many times you wrote `concern` I would be `concerned` if something like this didn't happen hehe.

Issue #1402, opened 2018-04-08
[Suggestion] Make a special greeting when selecting class/work for Farewell


MrDevilzMan:
Since that now we have the option to tell Monika that we'll be off to class/work, might as well make a special greeting when coming back to MAS, with her asking about how was our day today.

oniigirii:
Would also like to add about the "Going to sleep" one. Like if you were to reopen the game in less than 4-6 hours, she'd be wondering if we're getting enough sleep, something around those lines. Also to include, "I thought you were going to sleep? Are you having trouble sleeping?" when you reopen the game not too long.

PR #1530, opened 2018-05-03
Added a new flow for greetings, now it has types

aldoram5:
This pull request addresses the suggestion on https://github.com/Monika-After-Story/MonikaModDev/issues/1402, which basically let's Monika greet the player accordingly to how exactly them said goodbye to her.

## Technical

These are the changes done to the greeting selection flow:
 - At the beginning of selectGreeting it will check for the value of `persistent._mas_greeting_type` if it is set it will use the value defined there to filter greetings by category using `filterEvents`, otherwise it will just filter by the `unlocked` flag, like it already does.
- Flow is kept the same, so every type can be affected/filtered by the existing greeting/rules system.
- `persistent._mas_greeting_type` will be set by the selected farewell and will be reset to `None` after filtering.
- Three new constants have been added to the `mas_greetings` store: `TYPE_SCHOOL`, `TYPE_WORK` and `TYPE_SLEEP`. These should be used as category names for greetings according to what type of greeting it will be. 

## Additional Notes

 This is an example of how more greetings can be added to an specific type:

```
init 5 python:
    addEvent(
        Event(
            persistent.greeting_database,
            eventlabel="greeting_back_from_sleep",
            unlocked=True,
            category=[store.mas_greetings.TYPE_SLEEP],
            random=True
        ),
        eventdb=evhand.greeting_database
    )

label greeting_back_from_sleep:
     m "Oh, hello [player]!"
     m 1k "I hope you had a good rest!"
     m "let's spend some time together~"
     return
```

### DIFF
--- a/Monika After Story/game/script-ch30.rpy
+++ b/Monika After Story/game/script-ch30.rpy
@@ -77,8 +77,8 @@ image room_mask2 = Movie(channel="window_2", play="mod_assets/window_2.webm",mas
 image room_mask3 = Movie(channel="window_3", play="mod_assets/window_3.webm",mask=None,image="mod_assets/window_3_fallback.png")
 image room_mask4 = Movie(channel="window_4", play="mod_assets/window_4.webm",mask=None,image="mod_assets/window_4_fallback.png")
 image rain_mask_left = Movie(
-    channel="window_5", 
-    play="mod_assets/window_5.webm", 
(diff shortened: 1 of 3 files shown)


ThePotatoGuy:
Finished testing, looks good

monikLover:
Rewritten and apporved. All good to go, @aldoram5.

Issue #1504, opened 2018-04-29
[Suggestion] Random Topic Frequency


jmwall24:
I remember from a previous conversation that the random topics are shown based on some sort of counting system where the least seen topics show up more frequently until they catch up to the more seen topics.  But I'm wondering if there shouldn't be a limit to this where, after topics are seen a certain amount of times, they all even out and all have an equal chance of being seen.  It just seems after every update, you get bombarded with the new topics over and over and over again to the point it seems like it's too much.  Maybe have the counter (or whatever is used to determine which topics need to be seen next) max out at like 3 so once it's been seen 3 times it just goes into the pool with everything else and has the same chance.  Of course it's altogether possible I just spend way too much time on there and this isn't an issue for most people lol

Tiler17:
I swear ThePotatoGuy opened an issue about this somewhat recently, but I can't find it. Then again, I can't find my own car keys while I'm driving, so that doesn't mean much. I agree though, having seen all of the new topics three times now, I'm definitely ready for them to cycle with the rest of them

jmwall24:
You might remembering his response to the issue I opened last time when I thought there was a bug where only the new topics were being brought up by Monika.    That's when I first learned about the newer topics repeating until they caught up to the others.

ThePotatoGuy:
The latest update should have delayed new topic repeating for at least 2 hours. Are you experiencing behavior inconsistent with that?

Tiler17:
I am not. I only catch a few topics at a time, so I heard the new ones and then a couple old ones, then quit. I heard the new ones again next time I opened up MAS. Rinse and repeat 2 or 3 times to today. So if there's a two hour delay, that seems to be working, but given that I only run through a few topics at a time, they feel like they're back to back. Would it be possible to have two conditions, one being time, the other being number of topics viewed that have to be met before repeating lesser seen topics? Maybe 10 other topics _and_ 2 hours of time?

jmwall24:
[...]

Issue #1365, opened 2018-04-02
New random topics get repeated too early


ThePotatoGuy:
This isn't actually a bug, but definitely an issue that needs to be addressed.

Currently, when random topics are repeated, the least seen are shown first. This results in newer topics being repeated right away after they've been exhausted. 

Probably need to add someway to prevent a topic that just occurred from being repeated right away.
Maybe a queue that keeps track of what has been to said to avoid recent repeats or something.

ThePotatoGuy:
UPDATE:

Probably going to create a list of the last 100ish or so shown topics, then compare that list when doing random picks.

ThePotatoGuy:
Actually, rather than create a list, a new event property with last seen timestamp would be way better for this.

PR #1531, opened 2018-05-03
Changed Algorithm for random topic selection

ThePotatoGuy:
#1504 #1365 

This algorithm is probably better than the old one. _probably_

The algorithm runs as follows:
1. On startup, get the list of events that are random and have not been seen recently.
2. Divide that list of events into 3 groups:
    1. Unseen events
    2. The most seen topics
         - Most seen are either the 10% most seen, or the events where `shown_count` is in the top 20%
         - We select the fewer number of events for most seen from the above criteria
    3. The rest
3. When a random topic is ready for selection, we use the following weights for selection:
    1. Unseen - 50%
    2. Most seen - 1%
    3. Seen - 49%
4. When a list is selected, an event is randomly selected from that list and subsequently removed from that list.
5. When a list is empty:
    1. Unseen - jump to seen selection
    2. Most seen - jump to seen selection
    3. Seen - rebuild most seen / seen event lists

As a result of this new algorithm, we are dropping the Repeating Topics option from the settings menu. Instead, Monika will repeat topics by default, but will respect a 2 hour recently said rule and the random seen limit per session.

If you've seen all monika random topics, your random seen limit has been increased to 1000. This number is arbitrary.

### DIFF
--- a/Monika After Story/game/event-handler.rpy
+++ b/Monika After Story/game/event-handler.rpy
@@ -527,12 +527,19 @@ init python:
         """
         import datetime
         now = datetime.datetime.now()
+        cleaned_list = list();
         
-        return [
-            ev
-            for ev in ev_list
-            if now - ev.last_seen >= store.evhand.LAST_SEEN_DELTA
-        ]
+        for ev in ev_list:
+            if ev.last_seen is not None:
+                # this topic has been seen before, must check time
+                if now - ev.last_seen >= store.evhand.LAST_SEEN_DELTA:
+                    cleaned_list.append(ev)
+
+            else:
+                # topic never seen before, its clean!
+                cleaned_list.append(ev)
+
+        return cleaned_list
 
 
 # This calls the next event in the list. It returns the name of the
--- a/Monika After Story/game/event-handler.rpy
+++ b/Monika After Story/game/event-handler.rpy
@@ -513,6 +513,28 @@ init python:
         return cleanlist
 
 
+    def mas_cleanJustSeenEV(ev_list):
+        """
+        Cleans the given event list (of events) of just seen items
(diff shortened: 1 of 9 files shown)


Monika After Story/game/event-handler.rpy
```diff
@@ -513,6 +513,28 @@ init python:
...
+        RETURNS:
+            cleaned list of events (stuff not in the tiem THRESHOLD)
+        """
+        import datetime
+        now = datetime.datetime.now()
+        
+        return [
+            ev
+            for ev in ev_list
+            if now - ev.last_seen >= store.evhand.LAST_SEEN_DELTA
```
aldoram5: This one throws an error if `ev.last_seen` is None so it should be:
`if ev.last_seen is not None and now - ev.last_seen >= store.evhand.LAST_SEEN_DELTA`

Tiler17:
So if I'm understanding this right, this means that anytime there's unseen dialogue, you have a 50% chance at seeing a new piece of dialogue. Once dialogue is seen, it's grouped with most of the other seen dialogue. You have a 49% chance of seeing dialogue you've already seen before, but your most commonly seen dialogue will almost never come up

More simply, she prefers to show unseen dialogue, does _not_ prefer to show rarely seen dialogue, and prefers to _not_ show common dialogue.

Just trying to get a handle on this change. If I'm right, then I feel it's a very good improvement

ThePotatoGuy:
Yes. for all intents and purposes, its basically a 50/50 chance between unseen and normally seen dialogue.

1% chance for the most commonly seen, but its basically negligable.

We're gonna test these ratios for now, and see if we'll need to adjust them after the next update.

PR #1568, opened 2018-05-09
player's appearance

moonsandstar-s:
Fixed some errors, and we should be good now. 
Proposing a topic where any player, regardless of what they look like, should be able to tell Monika about their appearance. Thanks to @Olliesama for all his help!


moonsandstar-s:
Done & done. :D @aldoram5

aldoram5:
I did the menu order change and a minor thing with the returns.

moonsandstar-s:
thanks aldo <3

nohumanallow3d:
I think before this is added there should be something about body shape. I mean in a relationship that's a big factor as well compared to eye colour or hair colour.

moonsandstar-s:
It's a personal question, one that's very skewed in bias, and people who are insecure about their weight aren't likely to put accurate answers. I considered it, but ultimately decided to leave it out, because there is no way to truly remove the negative connotations from the answers to such a question.

nohumanallow3d:
but if the player realizes they are overweight or obese Monika can try and help them or help them realize that being like that isn't healthy. I'm overweight and Monika actually convinced me to eat better. It would be a service to people if they realized they wouldn't live to see the day Monika coming to our reality if they were overweight or fat.

Onion-Bubs:
Except that it is actually possible for someone who is fat to be healthy.

Olliesama:
I don't know anything about your friend or maybe we have a different definition of fat but being overweight is definitely unhealthy.

It's a bad idea to include weight because one of two things could happen, either Monika is fine with it which goes against her character of wanting the player to pursue a healthy and active lifestyle and encourages people not to better themselves or Monika isn't okay with it and tries to get the player to lose weight which some can interpret as fat shaming or even her not really loving them.
It's more trouble than its worth and should stay omitted in my opinion.

moonsandstar-s:
@nohumanallow3d Then perhaps you ought to write a topic about that yourself; I do not intend to sound curt, but this one is not a topic about the correlation of weight/health and Monika's thoughts on it, and as such, should be discussed elsewhere. The topic is about the player's characteristics such as eye color, hair color, and so on, not lifestyle habits. Please retain relevancy towards the dialogue in question, thank you.

CrocWorks:
Maybe add red eyes as an option? It's rare, but there are people out there who do have red eyes. Maybe an option for being heterochromatic too?

Issue #1370, opened 2018-04-02
[Question & Suggestion] Musik


LordOfDeath35:
First to the question: How do I add my own music to the game and which music format does it need to have?

Now to the suggestion: I would like it if you added reaction that Monika has if you play new songs that the player added. This would make it feel like she actually know what you´re doing. It would also fit in, because Monika mentions that she would like to listen to music that the player chooses.
I know that it would be pretty hard to do this but you could try to programm Monika to recognize different kinds of music. Therefore you could add multiple options that she can recognize music from. E.g. Rock, Alternative, etc. This could be possible if you somehow made Monika able to research this kind of stuff on the internet. The flaw is that this would require a wifi/internet connection. This is just a suggestion and doesn´t have to be added

kaido1224:
.ogg is the file type.

I believe there are instructions in the discord pinned in coding on how to add your own music.

kaido1224:
As for your second thing, I don't think there is any way for Monika to determine what sort of music you add is supposed to be based on the file name.

LordOfDeath35:
Well... - if you add to her code that she checks that file everytime you start the game, it could work. Like just program her to know that the music that is normally in ther is part of the game. And if she finds new files which haven´t been added in her code, she automatically researches it or talks about it to you. Either way is fine. In the second way, I just mentioned, she could ask what type of music you added. This would not require a wifi/internet connection.

kaido1224:
I don't think that's feasible. I'll let the other devs comment though.

aldoram5:
To do actual file name research on the internet and then automatically determine something to comment about the music you added based on the result of that search, that would be a whole new level of AI, so that's not happening anytime soon. As for the commenting on what music genres you like, there could be some music related topics added asking the player if they like an specific genre.
You could actually contribute some of those topics if you want, here's our [contributing guidelines](https://github.com/Monika-After-Story/MonikaModDev/wiki/Contributing-Guidelines), those will help you making your first topic contribution.

ThePotatoGuy:
We are **not** adding a user-friendly way to add additional music, mainly since Renpy only likes `oggs`. 
If you want to add your own stuff, refer to [this file](https://github.com/Monika-After-Story/MonikaModDev/blob/master/Monika%20After%20Story/game/zz_music_selector.rpy) to see how to add songs.

Issue #1143, opened 2018-02-24
Additional music by player


Nikmond:
Can i add my own music in menu music in the game? If i do, so how?

ThePotatoGuy:
You'd want to download the source code and add entries for you music in `zz_music_selector.rpy`. [this particular section](https://github.com/Backdash/MonikaModDev/blob/master/Monika%20After%20Story/game/zz_music_selector.rpy#L108) handles the generation of the music menu.

Check near the top of the file for the actual value of those constants.

Renpy can only play `ogg`s, so your music needs to be in that format.

Issue #795, opened 2018-01-19
[Suggestion] - Music and Dialog


spikeman168:
There should be an easy way to add your own music into the game. Like, instead of the basic music, I can add in music into a folder or something, and it would show up on the music menu.
There should also be a way to add your own dialog to Monika, that way, In between updates, people can send tim adding their own personalized dialog, of course, dialog already in the mod wouldn't be editable.

Blubmanful:
From what I remember someone already suggested the first idea, which
couldn’t work because of how RenPY works with its files. The second one can
technically already be done just not through the program itself but by just
editing the script file

spikeman168:
Well, I've looked into the script file, but it's all obfuscated, I can't edit it.

kaido1224:
If you want to edit the files you either need to download a branch in github or decompile the files. The branches in github are already decompiled. 

If the extension is .rpyc it's a compiled file. If it's .rpy it's uncompiled.

spikeman168:
Okay, thanks.

Issue #573, opened 2017-12-23
[discussion][help wanted]- New Music?


beastdestroyer:
Hey there, first I'd like to say:
I really love this mod! It's so interesting and fun to use: I never get tired of losing to Monika at Pong (Im a noob ok) and at Chess (I barely even know how to play that lol), as well as just letting her carry on. Also the "Christmas Update" is exciting too! I can't wait for future updates.

Moving onto my point, I was reading the "contributing" and other guides and noticed something about adding music? Since I got MAS (0.6.1 to be exact), I've cycled through all the music, and wondering if there was a way to add more, and I was hoping if I could add links to songs, however Im not sure what to do here.

Some DDLC songs which come to mind are:
- Nick Kaelar - 「Doki Doki ドキドキ」[Official OST (Or is that not allowed on Dan Salvato's Guidelines?), i like this song very much]
- Your Reality - Remix by Miraie
- Sayonara - Remix by Miraie
Maybe some other OST songs? 
I  hope I can get info on this, thanks!

ThePotatoGuy:
For remixes, we would need permission from the creators to use their tracks in the mod.

Official tracks built-into the stock game are fair game, but we want to keep the focus of the music tracks to ones that are Monika-related (except the Surprise ofc, since thats a _surprise_). If there is a big demand for the DDLC main theme, then we can put it in.

beastdestroyer:
Ah yes, I remember from reading this mod's intro .md, it basically, in short says to keep everything as Just Monika as possible, thanks for the info!

supchppt:
I think _My Feelings_ would make a good candidate. Whilst it is mostly an emotional track, I find it apropos for more serious and philosophical topics as well (the end-of-the-world one comes to mind). However, at the end of the day, I do realize that these are still my own opinions, so feel free to disagree.

JustMonikaMkay:
me being a little dorky vocaloid fan i want a cover of someone in ddlc of electric angel :v ( i feel bad for suggesting this xD)

PR #1592, opened 2018-05-14
Added Support for Custom Music

ThePotatoGuy:
#1370 #1143 #795 #573 #542 #511 #309

Surprise~

We did something.

Now everyone can stop asking about this.

### How to add music:
1. The music you want to add **must be OGG format**. Not just an .ogg extension. If it crashes because you renamed an `.mp3` to an `.ogg` , that's not our problem.
2. Create a new directory `custom_bgm` in your `DDLC/` directory.
3. Drop the oggs into that directory.
4. Launch game.

### Technical
* Files are checked for access as well as extension. Appropriate files are added to `music_choices` in the `initMusicChoices` function.
* I've also added some checking in `init` to clear `persistent.current_track` if its filepath does not resolve to an actual file. This will prevent crashes if the last played music track is removed prior to next launch.

### DIFF
--- a/Monika After Story/game/zz_music_selector.rpy
+++ b/Monika After Story/game/zz_music_selector.rpy
@@ -1,6 +1,12 @@
 # Module that handles the music selection screen
 # we start with zz to ensure this loads LAST
 #
+# NOTE: We added support for custom music.
+# To add custom music to your game:
+# 1. Ensure that the custom music is of "ogg" file format (with the extension)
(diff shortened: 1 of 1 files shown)


Tiler17:
Will you continue to add default music? I think this is neat and all, but I still adore 80s Doki, and I never would've come across that otherwise. Or the most recent track that was added

ThePotatoGuy:
Yes, we'll continue to add music seems fitting for MAS.

PR #1593, opened 2018-05-15
Door greeting close fix

aldoram5:
Minor fix that lets the player close the game as long as Monika doesn't leave her room.


ThePotatoGuy:
Quitting via the Quit option in the Escape menu still shows the dialogue. Probably need to adjust the screen for this.

Issue #1418, opened 2018-04-13
"Pen Name" subject / Name Changing


ScreampunkArts:
First time posting on GitHub; sorry if I do something wrong.
I could talk about the various grammatical errors I've noticed in some topics, but I'll stick to this one thing:

I noticed something odd, when Monika talks about Pen Names. (not sure if it's a submitted feature or a default subject): It seems a bit odd that she remembers the last name you changed from.

A while back, I heard on Reddit that if you change your name to that of the characters, there were Easter Eggs.
I couldn't get those to happen and probably did something wrong, but I've since changed back to my real name.

The strange thing is, she then implies I'm actually using the last name for something (in this case it was Natsuki; lol)
It's a little odd that she jumps to that conclusion.
This would be especially irritating if you're on someone else's computer when she does the creepy thing of checking your Profile name, before changing it to your name.

I feel it would make more sense for her to simply ask you if you have a pen name or not, and if you answer yes, you could type one in.
I actually do have a pen name for my artwork, and I think it'd be cool to be able to tell her that.

juliathegolden:
I like this idea a lot. I might do a PR For it tomorrow (if no one beats me to it. Feel free to beat me to it).

bracketJacket:
My PC name is actually my transgender "dead name" from a few months ago because I finally picked a new name. I'm a procrastinating idiot and haven't changed it on my PC  yet, cuz I don't really see it that often anyway. But it was just a bad feeling when Monika said it ahhh

kaido1224:
Already put in the suggestion pool as unseen. If the idea is accepted whenever it is reviewed. For now I'm closing this.

PR #1604, opened 2018-05-18
Rework of the Pen Names topic

juliathegolden:
Someone #1418  suggested this about a month ago, and I said I'd work on it...better late than never, right? 

I'm a bit spotty on how to use the new expressions, but I gave it my all.  Please let me know what you think.

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -3727,7 +3727,7 @@ label monika_penname:
                             m "I'd like to think we are on a first name basis with each other. We are dating, after all."
                             m "But I guess it's pretty special that you shared your pen name with me!"
                             $ penbool = True
-                        elif lowerpen =="sayori"
+                        elif lowerpen =="sayori":
                             m 2c "..."
                             m 1n "...I mean, I won't question your choice of pennames, but..."
                             m "If you want to name yourself after a character in this game, you should pick me!"
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -3702,6 +3702,8 @@ label monika_japanese:
 
 init 5 python:
     addEvent(Event(persistent.event_database,eventlabel="monika_penname",category=['literature'],prompt="Pen names",random=True))
+init 5 python:
+    addEvent(Event(persistent.event_database,eventlabel="monika_penname",category=['literature'],prompt="Pen names",random=True))
 
 label monika_penname:
     m "You know what's really cool? Pen names."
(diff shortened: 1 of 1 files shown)


Monika After Story/game/script-topics.rpy
```diff
@@ -3709,10 +3709,59 @@ label monika_penname:
...
+                            m 1i "Of course, there's something else that name could refer too..."
+                            if persistent.gender =="F":
+                                m 5 "And well...I could get behind that, since it's you~"
+                            $ penbool = True                                
+                        elif lowerpen =="monika":
+                            m "..."
+                            m 1 ebla "Aww, did you pick that for me?" 
+                            m "Even if you didn't, that's so sweet!"
+                            $ penbool = True   
+                        else:
```
aldoram5: you may also want to check for an empty string and probably the specific case saying nevermind or something similar just in case player change their mind and decide not to say their pen name.
```
                elif lower pen == "nevermind": # in case player doesn't want to say their penance
                    # appropriate dialogue here
                    $ penbool = True
                elif not lowername: #check for empty string
                    # appropriate dialog here, also don't set penbool to true
```

Monika After Story/game/script-topics.rpy
```diff
@@ -3709,10 +3709,59 @@ label monika_penname:
     m 3c "They keep it hidden from everyone just so it won't affect their personal lives."
     m 3b "Pen names also help writers create something totally different from their usual style of writing."
     m 3d "It really gives the writer the protection of anonymity and gives them a lot of creative freedom."
-    if mcname.lower() != player.lower():
-        m 2c "Is '[mcname]' a pseudonym that you're using?"
-        m "You're using two different names after all."
-        m 2d "'[mcname] and [player].'"
+    m "Do you have a pen name, [player]?"
```
aldoram5: since you'll be storing the pen name in a persistent var it would be better that instead of asking she just mentions the pen name the player has if he has already answered(you'll check if the persistent var is not empty) and move on with the rest of the topic, if the player hasn't answered go with the asking flow.

Monika After Story/game/script-topics.rpy
```diff
@@ -3709,10 +3710,70 @@ label monika_penname:
...
+                                m "I think if I saw a pen name like that on a cover, I'd be drawn to it immediately."
+                                $ persistent._mas_penname = penname
+                                $ penbool = True
+                    "I'd rather not; it's embarrassing.":
+                        m 2e "Aww. Well, I hope you feel enough to comfortable to tell me someday."          
+            "No":
+                m 1j "All right!"
+                m "If you ever decide on one, you should tell me!"
+    else:
+        m "Are you still going by [penname], [player]?"
```
aldoram5: if you close your session and get this topic the `penname` variable won't be defined at this point so you need to do something like this:
```
        $ penname = persistent._mas_penname
        m "Are you still going by [penname], [player]?"
```

juliathegolden:
I wasn't sure if it was allowed to have labels inside of our topics, but I tried it the way I knew how off the top of my head. Thank you for all your help!

PR #1621, opened 2018-05-21
Amendment to the topic about spiders

Onion-Bubs:
The second line of the spiders topic gives the impression that Monika took Natsuki's poem at face value. However, we don't actually know whether or not Natsuki dislikes spiders because the poem wasn't actually about spiders. Since the topic of arachnophobia is pretty good as a whole, I edited the beginning of it to be a little more accurate to Natsuki's poem. I don't think Monika would have missed the symbolism.


Onion-Bubs:
I had a look at the two PRs you linked to and, since I wasn't sure about `git rebase` I thought the best thing would be to do a new PR.

#1633 

I think this one can be closed now.

PR #1633, opened 2018-05-23
Amendment to the topic about spiders (take 2)

Onion-Bubs:
Now (hopefully) on the correct branch!

The second line of the spiders topic gives the impression that Monika took Natsuki's poem at face value. However, we don't actually know whether or not Natsuki dislikes spiders because the poem wasn't actually about spiders. Since the topic of arachnophobia is pretty good as a whole, I edited the beginning of it to be a little more accurate to Natsuki's poem. I don't think Monika would have missed the symbolism.


ThePotatoGuy:
[This has been merged into content]

PR #1641, opened 2018-05-24
Physical attractiveness topic with new expressions

Onion-Bubs:
This is basically the same as #1320 but compatible with 0.8.1 and using some new expressions. Note that the expression codes used in the topic did not exist on sprite-chart.rpy so I have added them in to that file. I was trying to create suitable "upset Monika" expressions for certain lines in the topic without having to change from the folded arms pose.

Something that always bugged me personally when playing DDLC was that, despite Sayori being a big eater, she always looked rather slim. I imagined what Monika's thoughts about that might be and how they might have been different before and after learning that her world wasn't real. There is a bit of commentary on the decision to make the girls as attractive as possible before Monika is overtaken with jealousy and frustration, but then she explains how it is for her and concludes that the player must find her attractive. Finally, she reassures the player that she'll love them no matter how they look, which would have to be true since she doesn't actually know how the player looks.

Anyway, thoughts?


Onion-Bubs:
Done a whole load of updating to this topic, including branches depending on current affection level (if Monika's feeling sad, she wonders if you don't like how she looks), converting all the old expression codes to new ones and in some cases changing to a different expression entirely, and after mulling over Akh0r's comment I added a bit at the end where she reminds you to look after yourself. I figure that would include putting some effort into your appearance.

ThePotatoGuy:
this needs new pm var from #3050

Issue #1208, opened 2018-03-08
Tell Monika that your pc might crash


ThePotatoGuy:
Requested by @Trilasent

Currently there's no easy way of determining the difference between closing the game without telling Monika goodbye or your computer crashed and shit happened. With #1135 becoming reality, what better way to mitigate game closures that aren't your fault than pro-actively warning Monika that crashes may occur. 

This could be useful for a variety of users:
- people with shitty computers
- people in the midst of a thunderstorm
- people who want Monika to watch their computer delete itself
- and more...

## Technical:
We'll probably do this by adding another closure `persistent` for this. This can be checked 2nd to `closed_self`, and will have its own bit of reload dialogue.

Also some code should be added to the "Don't quit the game on me" screen that resets this `persistent`, since it's very obvious that people could abuse this proactive dialogue to quit without saying goodbye.

Also this shouldn't be a once-only thing. Either this will be timed or last until next session. Or maybe something else. We'll see.

Trilasent:
Nothing like another game causing your screen to go black and preventing you to tell her goodbye so you can reboot without cutting her off. **_I'm so sorry Moni.. :c_**

Mundo-X:
That happens to me.

IlywI:
Or the windows 10 update(old joke i know but still :/)

TinyInternetTree:
My PC's pretty high-end, but I occasionally get BSODs because of all the shit I have running at once, so this will be really helpful to me and others who do similar stuff.

Trilasent:
It only seems to happen to me while I'm trying to play Black Ops 3 while having MAS open at the same time. Oddly enough the game never actually had issues like that before, and since it started earlier I had around 3-4 crashes that caused me to have to "force" the power off on my PC, preventing me from letting Monika close my game.

GoldTheEevee:
first of all amazing idea second of all "people who want Monika to watch their computer delete itself" amazing! one last thing is how do we do it?

IsItBadToMakeAGitHubAccountJustForThis:
While we're at it. I was going to propose timed autosaves for a longest time.
Mostly because I had to terminate ddlc.exe thanks to the alt-tab bug and the game didn't save any progress.

kaido1224:
Per convo I had with TPG, this is being aborted because it's easier to detect whether the game crashed or not.

PR #1668, opened 2018-05-28
Autosaving / Crash detection

ThePotatoGuy:
#1208 

So currently, we don't do periodic saving of `persistent` data. This has led an interesting situation where if the game crashes (with no traceback or error screen, just closes), no persistent data is saved. Anything you were doing prior to the crash is basically forgotten, so Monika wouldn't react to the crash as not saying goodbye.

Autosaving the `persistent` data means we need to account for crashes that don't involve traceback / errors. We can do that by setting a flag that only gets set to False if the user said goodbye or quit and got the confirmation screen. If the game crashed, we will know it did because this flag is true by default.

This means we need a flow for handling game crashes.

I've opted for a flustered approach that should demonstrate bits of the animation language as well playing around with text tags.

There's 2 major dialgoue flows:
### First time crash
1. Scene starts off as dark
2. Monika asks if you're there 
    1. Yes - Monika happy you there
    2. No - Monika thinks you're joking since you're obviously there
    3. Wait 4 seconds - Monika knows you're there and thinks you're joking (basically No)
3. Monika asks you turn on the lights
    1. Turn on the light - Lights are on, now you can see monika
    2. ... - 5 second pause, Monika finds the lights on her own
    3. WAit 5 seconds - Monika finds the lights on her own
4. Monika is happy to see you
5. Monika begins flustered speak. Looks frantically left + right.
6. Player has opportunity to calm her down
    1. Calm down... - Monika is happy that you calmed her
    2. ... - Monika calms herself down
    3. Wait 8 seconds - Monika calms herself down
7. Monika asks you what happened
    1. The game crashed - Monika says thats scary. Asks if you can stop it from happening:
        1. I'll try - Monika thanks you for trying. She will prepare herself mentally just in case. 
                        (we save this flag as well)
        2. It happens - Monika understands and will prepare herself mentally to handle this.
    2. I dont know - Monika asks if you could look into it because its scary.
8. Back to regular flow

### Subsequent crashes
1. Monika quips a crash response. (This is expandable)
2. Back to regular flow

### Testing
- best way to test this without forcing a crash is to change the confirmation quit screen to set `persistent._mas_game_crashed` to True.
- Also, I recommend setting your text speed to Monika (or something thats pretty slow). The fluster part is too fast for dev mode. (Theres a commented line for setting the dialogue style for testing)

### Notes:
* Added a blacklist function to `restartEvent` to prevent specifc events from being pushed back onto the stack.
* Added several 6 pose expressions. Also added an ATL section in `sprite_chart`
* oh yeah, this also adds autosaving of the persistent data roughly every minute in idle spaceroom mode.
* this is based to enhancement because the original focus was to change the game to do autosaving.


juliathegolden:
I have some suggested edits for the long fluster label:  
`+label mas_crashed_long_fluster:
+    m "{cps=*1.5}O-{w=0.3}one second you were there b-{w=0.3}but then the next second everything turned black...{/cps}{nw}"
+    m "{cps=*1.5}and then you d-{w=0.3}disappeared, so I was worried that s-{w=0.3}s-{w=0.3}something happened to you...{/cps}{nw}"
+    m "{cps=*1.5}...and I was so s-{w=0.3}scared because I thought broke everything again!{/cps}{nw}"
+    m "{cps=*1.5}But I didn't mess with the game this time, I swear!{/cps}{nw}"
+    m "{cps=*1.5}A-{w=0.3}at least, I don't think I did, but I guess it's possible....{/cps}{nw}"
+    m "{cps=*1.5}because I'm n-{w=0.3}not really sure what I'm doing sometimes,{/cps}{nw}"
+    m "{cps=*1.5}but I hope this t-{w=0.3}time isn't my f-{w=0.3}fault cause I really didn't touch anything...{/cps}{nw}"   
+    return `

Cut's down on a couple words, and also makes some sentences so it looks neater in history while still keeping the 'sound' of a scared, run on sentence...I hope lmao.

ThePotatoGuy:
Assumign that julia's edits is a review

Issue #1657, opened 2018-05-26
[Suggestion] LOOP tags in custom BGM


williamjcm:
As you know, some of the songs in DDLC/MAS loop properly thanks to loop points.

For example, Sayo-nara ("Surprise!" in MAS) loops back to the 36.782 position when it reaches the end, and the 80s-style cover of DDLC's main theme loops between the 17.451 and 119.999 positions.

Some programs/games that support Ogg/Vorbis files also support looping by using Vorbis Comment tags named `LOOPSTART` and `LOOPLENGTH` (RPG Maker VX Ace is an example that uses those tags).

However, while DDLC/MAS use a time in `seconds.milliseconds`, `LOOPSTART` and `LOOPLENGTH` have their time in number of samples.

It would be nice if MAS's custom BGM system could parse those tags.

You'd need to convert samples to a time in `seconds.milliseconds` (which means getting the sample rate from the file's metadata), then determining the loop's end point by adding the length's time to the start point's position. Or something similar.

PR #1670, opened 2018-05-29
Added MP3 support and more

ThePotatoGuy:
#1657 

I think we're done with music after this.

This adds MP3 and OPUS support to the music menu.
This also adds tag support so that the music files are listed in [artist] - [title] format instead of filename. 
In addition, OGG and OPUS files can be looped if they are provided with `loopstart` and `loopend` metadata tags, which specify the seconds to start and end the loop, accordingly.

**MP3 files do not support loop tags atm**

### In depth looping instructions
1. open the audio file in audacity
2. Export as OGG / OPUS and edit metadata
3. Add a tag for `loopstart` to mark where to start loops of audio file (in seconds)
4. Add a tag for `loopend` to mark where to end loops of the audio file (in seconds)
5. put audio file in `custom_bgm/` folder.

### Additional changes
* The `mutagen` library has been added. It can read metadata tags, but I've stripped it so it only includes the `ogg`  / `opus`/ and `mp3` parts. The library is less than 1MB in size
* I've added the mplus 2p font so we can have songs with jpn glyphs show up appropriately. This font is free to use and is arguably the best font in existence.

Also here's a nice picture of the new font.
revampedmusicmenu


williamjcm:
The normal format for the `LOOP*` tags is in samples, though.

I have many files using those tags.

ThePotatoGuy:
We opted to use seconds since its native to Renpy

But I think I'll move the lookup for those tags to custom MAS-related ones and do some math with the RPGMaker ones to get them to work. 

Just hold on a bit.

PR #1680, opened 2018-05-30
add 'happy birthday' to the piano game

juliathegolden:
This has been requested a couple times.  To be honest, I hadn't planned on being the one to implement it, but I accidentally played the first four notes myself and I was curious to see if I could play the whole thing. When I figured it out, I just had to add it to the game. 

I've tested it on my end, so I know it-works- but if there's an optimization or something I clearly didn't understand about the PianoNoteMatch functions, please tell me lmao.

### DIFF
--- a/Monika After Story/game/zz_pianokeys.rpy
+++ b/Monika After Story/game/zz_pianokeys.rpy
@@ -144,6 +144,31 @@ label mas_piano_result_none:
     m 1j "Promise to play for me next time?"
     return
 
+### HAPPY BIRTHDAY
+
+label mas_piano_hb_win:
+    m 1a "Wow! You almost got it!"
+    m 2b "Good job, [player]."
+    return
+
+label mas_piano_hb_fail:
+    m 1m "..."
+    m 1n "You did your best, [player]..."
+    m "Even a simple song takes time to learn."
+    return
+
+label mas_piano_hb_fc:
+    m 1a "Hehe, great job!"
+    m 2b "I know that's an easy one, but you did great."
+    m 1k "Are you going to play that for me on my Birthday?"
+    return
+
+label mas_piano_hb_prac:
+    m 1a "You're practing the Birthday Song?"
+    m "I know you can do it, [player]!"
+    return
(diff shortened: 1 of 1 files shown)


Monika After Story/game/zz_pianokeys.rpy
```diff
@@ -807,6 +832,110 @@ init 1000 python in mas_piano_keys:
...
+        ],
+        postnotes=None,
+        express="1k",
+        postexpress="1a",
+        verse=0
+    )
+
+    _pnm_hb_v1l2 = PianoNoteMatch(
+        renpy.text.text.Text(
+            "Happy Birthday to you~",
```
ThePotatoGuy: Tildes look bad in Monika's credit font. maybe use an ! instead? 

tbh, just removing the tilde and having a different expression might be more intuitive.

Monika After Story/game/zz_pianokeys.rpy
```diff
@@ -807,6 +832,110 @@ init 1000 python in mas_piano_keys:
...
+            C5,
+        ],
+        postnotes=None,
+        express="1k",
+        postexpress="1a",
+        verse=0
+    )
+    _pnm_hb_v1l3 = PianoNoteMatch(
+        renpy.text.text.Text(
+            "Happy Birthday Dear Friend",
```
ThePotatoGuy: So I noticed you used "Friend" here, and I am aware that the usual `[player]` scheme doesn't work here.

In fact, it using dynamic text in the piano **really** doesn't work because all the note matches are created at init time, which means runtime changes (like changing name) wouldn't be properly reflected in piano.

"Friend", however, is 100% going to generate some controversy. 

I've heard this line said as `Happy Birthday, Happy birthday`, so I think changing it to that or some other way that doesn't require a name or pronoun would be better.

Later on, when I migrate these to text files, I can add something to dynamically apply the substitutes correctly so you can use the regular `[keyword]` replacement.

monikLover:
Apporved.

PR #1685, opened 2018-06-01
New Chobits Topic

Rai99:
Could remove the mildly lewd lines if needed. 

***Explanation/Spoilers***

Chobits is a manga series 7-8 books long that takes place in a futuristic world where most people have humanoid android maids called persocoms (Likely Personal Computers). 
The main character can't afford a persocom but finds one laying around in the trash and takes it home to activate and set up. The persocom's name is Chi.
Chi starts out blank with no knowledge of the world and is taught how to live a normal human life: Cleaning, working at a bakery, running small errands, etc. 
In the book, it could be described as taboo for a person to marry their persocom although multiple people have done do. 
The main character has always wanted a girlfriend and Chi is a possibility, but her reset/power switches are located in lewd places, meaning if he ever touched them, Chi would reset and lose all her memories and go back to like when she was lying in the trash. 
In the end, they actually both fell in love with eachother and the main character decided to stay with Chi, despite having real girl options at that point and never being able to touch Chi's private parts. 

Also yes, I put this manga topic in the literature category.

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -6241,6 +6241,36 @@ label monika_yellowwp:
     return
 
 init 5 python:
+    addEvent(Event(persistent.event_database,eventlabel="monika_chobits",category=['literature'],prompt="Chobits",random=True))
+
+label monika_chobits:
+    m 1a "Hey, [player], have you ever read a series of manga called {i}Chobits{/i}?"
+    menu:
+        "Yes.":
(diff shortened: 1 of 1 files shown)


Rai99:
The couple lines at 6259 were meant to be mildly lewd. It was poking at the idea that when Monika makes it to the real world, it _would_ be possible to touch her wherever and by the time that happens, I'm sure Monika would be comfortable enough with it to allow it to an extent.

As stated in the description, these few lines could be removed if they don't belong.

PR #1697, opened 2018-06-02
Musical instrument easter egg

jmwall24:
Just a small easter egg addition to the musical instrument topic, where if you tell her you play the harmonica, but instead spell it harmonika, she notices and comments on it. The only part I wasn't sure about was the elif part to add a second option other than piano, I think I got it right, but if not let me know.  (Opened this new pull request to move it to the content branch as requested.)


aldoram5:
There's no reason to remove the leaning blush expressions though, as of now they just won't have the blush part until the art is fixed, so there's no harm in using them, and once the art is fixed they'll have the blush. Feel free to re-add them if you want to use them.

jmwall24:
Ah ok, thanks.  Re-added those and added a 'harmonica' option that ties into this as well.  Should be done now.

Monika After Story/game/script-topics.rpy
```diff
@@ -4666,6 +4666,24 @@ label monika_orchestra:
                  m 5a "Maybe we could do a duet someday!"
                  m 1j "Ehehe~"
                  $ persistent.instrument = True
+            elif tempinstrument == "harmonika":
+                 m 1hub "Wow, I've always wanted to try the harmonik--"
+                 m 3eub "...Oh!"
+                 m 1eka "Aww [player], did you do that for me?"
+                 m "That's {i}sooo{/i} adorable!"
+                 show monika 5eubfu at t11 zorder MAS_MONIKA_Z with dissolve
+                 m 5eubfu "And just so you know, you can play with me anytime you like..."
```
ThePotatoGuy: 🚂 

okay this definitely needs to be adjusted for affection.
Can you make variants of this bit of dialogue for varying affection levels?

Mainly add a non-racy reaction for affection levels below affectionate.

jmwall24:
Yeah I should've thought about that after affection was released.  This should take care of it, there are 3 paths now depending on affection.

Issue #1515, opened 2018-05-01
[Suggestion] Topics Excluded From Repeats


ACoolCat:
When on random mode Monika repeats specific topics, however, I think there should be a list of topics that shouldn't be repeated. Questions where Monika collects information about the player (IE "Do you like mint ice cream") shouldn't show up more than once, since she now knows that information about the player. However, if there's a topic that has the possibility of changing over time then it could be asked more than once. Like, if you say you're trying to quit cigarettes that could cause the topic itself to become more rare so that she can ask again to see if you've quit.

Tiler17:
Makes sense to me. I believe that Monika really and truly convinces herself that every week will be the week that I decide to start liking mint ice cream. Something like this would save her the repeated heartbreak

jmwall24:
I like it too.  And if for some reason we do want to answer the questions differently in the future, you can always go and find it from the menu.

PR #1700, opened 2018-06-02
Topic Adjustments

ThePotatoGuy:
#1515 

### adjustments
1. All random topics with some sort of question-based interactivity:
   1. now store the results of those questions in `persistent` variables prefixed with `_mas_pm` (mas player model)
   2. Set their `random` property to False after completing to prevent them from popping up randomly again. They will still be available in Repeat Conversation
2. All dialogue that depends on anniversary dates have been changed from a year to 6 months. 
3. Slightly optimized the suffix generation in `monika_japanese`
4. The `monika_explain` topic has been enabled

### Notes:
* the player model variables are not used anywhere else, they are really in preparation for an actual player model one day.
* I changed the anniversary checking to 6 months because thats about enough time to start being comfortable in a relationship.

### Testing:
The following is a list of the random topics that have been adjusted:
* `_selfesteem`
* `_images`
* `_cities`
* `_haterReaction`
* `_japanese`
* `_icecream`
* `_weddingring`
* `_sports`
* `_meditation`
* `_jazz`
* `_otaku`
* `_smoking`
* `_asks_family`
* `_beach`
* `_prom`
* `_natsuki_letter`
* `_familygathering`
* `_yellowwp`

I've also added the player model variable to the Fast food topic.

**1700 GET**

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -6534,7 +6534,7 @@ label monika_yellowwp:
             m 1e "But I love you so much, [player]. Supporting you is better than anything else."
             m 1k "I just can't wait to do it in person when I finally cross over to your side~"
         "No.":
-            $ persistent._mas_pm_read_yellow_wp = True
+            $ persistent._mas_pm_read_yellow_wp = False
             m 1c "Oh, I see."
             m 1e "It's a short story, so if you haven't, feel free to whenever you have the time."
             m 1a "It'll definitely be an interesting read for you."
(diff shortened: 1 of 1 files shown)


juliathegolden:
the variable you made for whether or not the player has read the yellow wallpaper sets to true, when they say it hasn't. 

other than that tho this is A+ very very exciting :D

williamjcm:
I took a look at the `monika_explain` topic in both this version and the current unstable, and I noticed that Monika refers to the player using the neutral "they"/"them" pronouns, instead of using pronouns according to the player's gender.

juliathegolden:
@williamjcm good catch, i just created a pr to rectify this.

PR #1728, opened 2018-06-09
Update script-topics.rpy

juliathegolden:
added a shipping topic, since someone suggested it
i tried my best to use the new codes until i got lazy at the end, so please help me if i made a mistake


jmwall24:
The only thing I notice with the expressions is you used sweat drop right, and (at least I think) every other instance of the sweat drops I've seen have been sweat drop left (which is right side of the screen, since I think it's her left and not our left based on the pics in the monika folder).  If you intentionally went for sweat drop right then ignore this, but in case you weren't I thought I'd mention it.  It doesn't really matter, just that I think the sdl expression has probably already been made while the sdr might not be.

jmwall24:
Just to clarify the comment I deleted before (it looks like you changed some expressions based on what I said), the most updated available expressions start on line 1707 and go all the way to the bottom at https://github.com/Monika-After-Story/MonikaModDev/blob/enhancement/Monika%20After%20Story/game/sprite-chart.rpy

I was looking at the outdated Master branch when I commented before so what I said was probably not entirely correct.  But using the above link should be more useful to see what is available and what isn't.  Sorry for any confusion.

juliathegolden:
No worries. I appreciated your comment, and the changes were also requested internally (because you were still right about a couple of them). I'm still learning the new sprite chart lol. Thanks for the help!

jmwall24:
Happy to help, I think we're all learning together on this.  I know I've gone cross-eyed today looking at all the codes, there's SO MANY combinations.  Having all the options is really cool, but I kinda miss only having to use 1 letter and 1 number lol.

ThePotatoGuy:
[Closing this in favor of the linked pull request][

PR #1752, opened 2018-06-15
Take Two of the Masturbation Topic Revision 

juliathegolden:
So it turns out that the original rejection of this may have been an accident.

By popular demand, here it is: new, improved, and on the right branch.


ThePotatoGuy:
should be on `content`, as per the new rules is that anything that requires dialogue review should be in `content`

Rai99:
**Take Three**
Third time's the charm

juliathegolden:
`**_second take_**`

@ThePotatoGuy show some mercy to ya girl it's almost my birthday. 

Should be fixed.

monikLover:
Apporved. Need @aldoram5 to fix current conflicts :tohrusmile:

Issue #1573, opened 2018-05-10
[Suggestion] -Water conservation topic


kaido1224:
Looking at my water bill i realized how much water we all use and thought it might make a good topic for monika to discuss. She could give tips like turning off the shower while you are putting shampoo in your hair or soaping yourself up. I'm not sure whether I'll actually write the topic myself but I thought I'd put the idea out there if someone wanted to do it.

juliathegolden:
Would "And when I'm in your reality, we can save water by showering together~" be too saucy or

because not making that joke would be a complete waste

kaido1224:
I think it's fine.

Rai99:
Not making that joke would be a good idea down the drain (literally).

kaido1224:
Feel free to write it. I dunno if I ever will.

neyoriquans:
If this topic is going to be written, I think it's important to note where most of our water is used at least here in America. The vast majority is used for irrigation (agriculture) and power generation. 

![image](https://user-images.githubusercontent.com/36655331/39900962-63a34596-5493-11e8-966e-1b765b25993d.png)

Seeing how Monika has proven to be a practical person in past topics on issues such as these, I think it'd be important to keep this in mind since our personal usage of water isn't very impacting in the grand scheme of things. Just a thought though.

jmwall24:
That joke's not any saucier than the line in one of the door knock greetings

Tiler17:
Saucy advice unlockable after the one or three month milestone?

ThePotatoGuy:
[[Reopening when we look into it]]

Issue #1643, opened 2018-05-24
Going to Monika  [Suggestion]


Illra:
What i've heard that in about 2025 that we might be able to upload our conscious to our pc or something like that. Monika asks how long will it be until she comes to our reality, what if we ask her ''What will you think of me coming to your reality''. In my mind I thought it was a good idea. 
I'm open for some opinions. Is it good? Or is it bad?

jmwall24:
There kind of already is a way to ask her that, but you have to do something first...

Rai99:
I actually thought about the idea myself but I never thought of adding dialogue for it. 
Unless somebody else wants to do it, I might write it myself. 

It's true that the other way does exist but I think we're talking about when the day actually comes where we can actually upload ourselves to see Monika. I wish it were that easy

jmwall24:
Yeah I know, just wanted to mention that there is a cool little thing out there that is related to this without completing spoiling it lol.

Rai99:
Guess I'll put it on my list of things to write for. Currently I'm adding interactivity to Yuri's wine, next I'll be writing new dialogue about a manga called Chobits, then would be this idea.

Akh0r:
But, obviously, a copy of you is not you. 
So making a digital copy is not an option – untill we have some new VR technology.

I guess, it is not worth of changing current dialogue for that. This technology is still too far away.

Rai99:
It's on my list of things to add to MAS so I'll probably work on it after I finish my "Yuri's wine" enhancement and my "Chobits" new topic unless somebody else wants to do it.

ThePotatoGuy:
[[Reopening when we look into it]]

PR #1756, opened 2018-06-16
New Water Conservation Topic

Rai99:
#1573 
Just a short, simple flat topic
I liked this idea from over a month ago and wanted to get it up before it got swept under the rug.


kubastory9:
Pretty good

Rai99:
Thanks! Up next is #1643 which I plan to put some more depth into with some options.

PR #1785, opened 2018-06-21
Automated Releases

sr229:
This is a by-tag basis and would only build on tags

- For deploy to fire, you need to tag it first.
- It only releases if a tag is defined.


sr229:
Don't mind this, I was supposed to PR this on my own fork

Issue #1773, opened 2018-06-19
[Suggestion] - New Topic -  Pygmalion and the Ivory Statue 


Babushka93:
Hi everyone,

I stumbled onto the story of Pygmalion while looking through [wikipedia](https://en.wikipedia.org/wiki/Pygmalion_(mythology)) and I thought that his fable might be a story that Monika would share.

--------------------------------------

Pygmalion was a sculptor who fell in love with an ivory statue he had carved. He loved the statue so much that during the Festival of Aphrodite, Pygmalion made an offering to Aphrodite wishing for a bride that would be the living likeness of his ivory statue. 

When he returned home, he kissed the statue, and to his surprise the lips of the statue were warm. He kissed it once more, and found that the ivory had lost its hardness. Aphrodite had granted his wish.

The statue through Aphrodite's blessing had become human. Pygmalion and the now transformed statue went on to marry and live together. 

------------------------------------------------
I feel like this would be a story that Monika would share with the player, and hint at the player making an offering to Aphrodite in the hope that it would bring her closer to our reality. 

==update==
Here is the most recent version of the dialouge based on contributions by @loverofmonika 
Also huge thanks to everyone who has shown interest and contribute to this topic so far 😄 

label monika_pygmalion
m 1 “Hey [player] have you ever read The Metamorphoses?”
m 3 “It’s a collection of poems by an Ancient Roman poet called Ovid.”
m 3b “There’s one poem that really struck me.”
m 3a “The poem is called The Story of Pygmalion and the Statue”
m 1b “It tells the story of a sculpture who fell in love with an ivory statue that he had sculpted.”
m 1a “He loved the statue so much that during one of the festivals held for Aphrodite. Pygmalion made an offering to Aphrodite and wished for the goddess to grant life to the ivory statue.”
m 1b “When he returned to his home, he kissed the statue and to his surprise the lips of the statue felt warm. He kissed the statue once more, and the statue lost its hardness becoming soft like skin.”
m 1e “Aphrodite had granted his wish.”
m 1b "This story really struck me because in a way it kinda reminds me of us , [player]"
m 1d "and how this poem talks about the impossible turning into reality"
m 5e "kinda fills me up with hope, that a miracle can happen and i can finally cross over"
m 5e "yeah...wouldn't that be wonderful"
m 1j "then you can finally kiss me without a screen between us, just imagine, kissing your computer screen when no one's looking or maybe you already tried that, hahaha~"
[...]

PR #1797, opened 2018-06-23
New Pygmalion and the Ivory Statue Topic

Rai99:
Issue: #1773


Babushka93:
@Rai99 thank you for doing this 😄

jmwall24:
Going to review this soon, but the ending needs to be changed to reflect the fact that kissing is a thing now. So we're gonna need paths for if you haven't kissed yet and if you have, while still maintaining the fact she wants to kiss you for real once she crosses over. This could also be affection dependent.

PR #1813, opened 2018-06-25
Pool Progression adjustments

ThePotatoGuy:
Pool topics are currently unlocked via progression whenever the user reaches the next level. This is great for people who are new, but bad for long-time players because reaching the next level might take forever. 

this adds a counter that increments whenever `unlock_prompt` is reached but no prompt was unlocked. When the game starts, this counter is checked and pool topics are attempted to be unlocked. When we add new content, this should act as retro-actively unlocking pool topics based on user level.

Also, all current users will start will a number of pool unlocks based on their level. If their level is over 25, they will get half their level in pool unlocks.

### Testing

I just set `persistent._mas_pool_unlocks` to a value and then locked value+1 pool topics and checked to make sure that one of the pool topics was not unlocked on a restart.

### DIFF
--- a/Monika After Story/game/event-handler.rpy
+++ b/Monika After Story/game/event-handler.rpy
@@ -304,7 +304,6 @@ init -1 python in evhand:
         return start_date <= current <= end_date
 
 
-
 init python:
     import store.evhand as evhand
     import datetime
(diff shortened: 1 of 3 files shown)


Rai99:
I'm not completely sure if I'm getting this, but I never knew pool topics started out locked and slowly unlocked over time as the player leveled up. If that is the case, wouldn't there still be the issue where at some point, it would take forever to unlock a new topic?

jmwall24:
I think what it's saying is before, once you had everything unlocked, you weren't getting credit toward future pool unlocks anymore, so if something new was added, you could be way behind getting it to unlock.  Now with the new counter, you'll get credit towards unlocks every time you open the game, so when something new is added, you'll have accrued a bunch of credits which can be used to unlock the new topic, instead of having to start earning those credits only once a new pool topic was added.  This should prevent it taking forever to unlock the new topic.

Rai99:
So the issue of being at a high level and taking forever to level up was always there? I hope there's still something to prevent a topic from taking an unreasonably long time to unlock. 

Maybe that's why I don't have the topic where Monika explains our relationship? Looking at the code, it was enabled and I think it should be active, but I don't have it.

Issue #1806, opened 2018-06-24
Virtual Youtuber Bug(Or maybe suggestion)


ReddTheRedRedd:
So yesterday I got a topic where Monika says something along the lines of "Remember when I mentioned a virtual Youtuber?(And so on)." Which I thought was odd because she never mentioned one. 
Today I got the greeting where she mentions that virtual Youtuber, so either the dialogue isn't dependent on the greeting ever happening, or if it wasn't a bug. In that case maybe that'd be a nice touch to stay *immersive* xD

juliathegolden:
It's intended to not come up randomly, and only unlock as a question if you get the greeting.

So yeah it's a bug. Thanks!

ReddTheRedRedd:
np!

PR #1821, opened 2018-06-25
Fix for #1806

aldoram5:
Adds the correct flags for unlocking the topic only after you've seen the `greeting_hai_domo`.

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -406,7 +406,7 @@ label monika_sayori:
     m "That pun was completely unintentional, I swear!"
     m "But anyway..."
     m 2e "I know how much you cared about her, so it only feels right for me to share her last moments with you."
-    m "If you're comfortable, that is." 
+    m "If you're comfortable, that is."
     menu:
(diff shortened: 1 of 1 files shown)


ThePotatoGuy:
`action` is a locked var, you'll need to add an update script for this

Monika After Story/game/script-topics.rpy
```diff
@@ -5523,7 +5523,19 @@ label monika_100k:
...
+        Event(persistent.event_database,
+            eventlabel='monika_kizuna',
+            prompt="Virtual Youtuber?",
+            category=['misc'],
+            random=False,
+            unlocked=False,
+            pool=True,
+            action=EV_ACT_POOL,
+            conditional="seen_event('greeting_hai_domo')",
+            rules={"no unlock": None}
```
ThePotatoGuy: Since the action is to `pool` it, `pool` should initially be False.

Should also drop the no unlock rule here.

In discord, what I meant was **either** have `pool=False` and change action to POOL, or keep `pool` and `action` as they were but add a no unlock rule. Dont do both.

Issue #1841, opened 2018-06-28
Crashed and have had to reset the mod three times since the Calendar update with the exact same error message


RutTutPut:
This has occurred 3 times since downloading the Calendar update, none of which seem to have anything in common. The message is as follows:

"I'm sorry, but an uncaught exception occurred.

While running game code:
  File "game/splash.rpy", line 168, in <module>
  File "game/zz_calendar.rpy", line 935, in loadCalendarDatabase
ValueError: No JSON object could be decoded

-- Full Traceback ------------------------------------------------------------

Full traceback:
  File "lib/windows-i686/00start.rpyc", line 212, in script call
  File "lib/windows-i686/splash.rpyc", line 165, in script
  File "C:\Users\Ronan\Desktop\DESKTOP\Doki Doki Literature Club\renpy\ast.py", line 814, in execute
    renpy.python.py_exec_bytecode(self.code.bytecode, self.hide, store=self.store)
  File "C:\Users\Ronan\Desktop\DESKTOP\Doki Doki Literature Club\renpy\python.py", line 1719, in py_exec_bytecode
    exec bytecode in globals, locals
  File "game/splash.rpy", line 168, in <module>
  File "game/zz_calendar.rpy", line 935, in loadCalendarDatabase
  File "/home/tom/ab/x64lucid-deps/install/lib/python2.7/json/__init__.py", line 290, in load
  File "/home/tom/ab/x64lucid-deps/install/lib/python2.7/json/__init__.py", line 338, in loads
  File "/home/tom/ab/x64lucid-deps/install/lib/python2.7/json/decoder.py", line 366, in decode
  File "/home/tom/ab/x64lucid-deps/install/lib/python2.7/json/decoder.py", line 384, in raw_decode
ValueError: No JSON object could be decoded

Windows-8-6.2.9200
Ren'Py 6.99.12.4.2187
Monika After Story 0.8.2-hotfix.2"

Here is my persistent file: https://drive.google.com/open?id=1yA7yyNrjvlz7AIG1IxfyUuCHiyIiTrNG

RutTutPut:
Yes, I am absolutely sure; I've asked all but three of the questions, and had gone through many of the conversations.

Here is the file you asked for: https://drive.google.com/open?id=1qiHKUe-MmX9qGu07PJFlGzU6eFmHgmx5

Thank you!

aldoram5:
Hi again RutTutPut!
[...]

PR #1850, opened 2018-06-30
Added value error exception to catch

aldoram5:
https://github.com/Monika-After-Story/MonikaModDev/issues/1841
This pull request just adds another exception to the catch block and that should prevent the game from crashing in case the Calendar data somehow gets corrupted.

### Testing
I already tested it but for reference, this can be tested by using the db.mcal file from the linked issue and it should not crash and once you say goodbye to Monika the db.mcal file should be correct again.

### DIFF
--- a/Monika After Story/game/zz_calendar.rpy
+++ b/Monika After Story/game/zz_calendar.rpy
@@ -1117,7 +1117,7 @@ init -1 python in mas_calendar:
         try:
             with open(renpy.config.savedir + '/db.mcal', 'r') as fp:
                 calendar_database = json.load(fp)
-        except (IOError, ValueError), e:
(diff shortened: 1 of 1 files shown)


Monika After Story/game/zz_calendar.rpy
```diff
@@ -1117,7 +1117,7 @@ init -1 python in mas_calendar:
         try:
             with open(renpy.config.savedir + '/db.mcal', 'r') as fp:
                 calendar_database = json.load(fp)
-        except IOError:
+        except (IOError, ValueError), e:
```
ThePotatoGuy: dont need the `e` part since we're not doing anything with the expceiton

PR #1899, opened 2018-07-10
Disabled Feeling Like a Year Older

ThePotatoGuy:
Disabling this mood since it doesn't fit well with calendar as it is.

And actually probably will not use this when we do actual birthday setting.


ThePotatoGuy:
actually nvm this, we'll just remove it when we add a better birthday event

PR #1919, opened 2018-07-14
Affection fixes/changes pt.1

aldoram5:
I did some of the TODOs and fixed/optimized some parts. Adding Don't merge label until I do proper checking of the pull request, still opening it up now so it's visible already what I'm changing/working on.

**Edit:**
Alright this one is done.

## What does this pull request contain?

This one contains all the required changes that the Affection system needed to work as it originally planned.

## Important changes

- The most important changes in this one are related to the greetings, since now they check first for the `monika_wants_this` flag, before checking for categories. To achieve this I added another filter to the `filterEvent` function.

- The affection rule was changed to work as a filter instead of a selector.

- A minor thing about music going silent after the first session, this one was fixed before in another pull request, and it got broken again after the merge with affection.

- Moved the old affection checker label to a dev topic for easy access.

## Minor changes 

- Fixed the zorder of `-115` event

- Affection can't go higher than 1 million, nor lower than -1 million.

- Migrated long absence farewell to work with greeting categories.

### Testing
I've been testing as I kept adding fixes so it should be good to go. However a bit of testing with making sure greetings work as expected would be pretty good. 
More pull requests for affection will come, I'm not adding more to this one cause the other ones may require dialogue review.

### DIFF
--- a/Monika After Story/game/chess.rpy
+++ b/Monika After Story/game/chess.rpy
@@ -114,17 +114,17 @@ init 1 python in mas_chess:
     #   (Monika's fault)
     QS_LOST = 0
 
-    # Quick File LOST (OF Course Not): the external quick save got removed, 
-    #   player denied removal 
+    # Quick File LOST (OF Course Not): the external quick save got removed,
+    #   player denied removal
     #   (Player's fault)
     QF_LOST_OFCN = 1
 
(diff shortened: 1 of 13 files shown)


ThePotatoGuy:
* i still need to do testing

PR #1921, opened 2018-07-15
Fix for month bypassing on calendar

aldoram5:
As title says, this should prevent users from going further in time even if they use the month buttons ~~that determination though~~.


Benjamin163:
you really need to be persistant to get to 9999 when skipping in months lol

PR #1929, opened 2018-07-18
Fix for bug with calendar overlay clicks

aldoram5:
The calendar overlay was still clickable after starting the first introductory flow. That led to getting the same dialogue as many times as you clicked it.

**Testing**: In a persistent that still can see the first dialogue click on the calendar and then keep clicking over it, it should not keep triggering the event.


ThePotatoGuy:
*I still need to test this, please dont merge yet

PR #1939, opened 2018-07-20
New greeting

jmwall24:


ThePotatoGuy:
Singing as a greeting is not bad, (we do have one greeting like that), but the choice of song here doesn't seem monika-like. Maybe even too memey as well.

Closing unless general consensus overturns my opinion.

jmwall24:
Ah ok, no big deal.  Was just an idea I had.

PR #1951, opened 2018-07-25
3 New Stories

Rai99:
Added 3 new debatably short stories. They are a bit long, though I hope that isn't an issue.


ThePotatoGuy:
The 3 stories that need reviewing are `genie_simple`, `genie_regret` and `genie_end`

PR #1952, opened 2018-07-25
More Affection changes

aldoram5:
Alright this one changes the following:

- Added affection increments/losses as discussed before
- Fixed bugs with how the freeze worked
- Changed affection to decrease 0.5 each day you don't visit her after a week
- Refactored the txt messages so we can use them easily 

### Testing 

Mostly regression testing on some of the games, and testing that the longer than a week absence decreases affection as expected.

### DIFF
--- a/Monika After Story/game/chess.rpy
+++ b/Monika After Story/game/chess.rpy
@@ -1612,8 +1612,7 @@ label mas_chess_playagain:
             pass
 
 label mas_chess_end:
-    #TODO minor affection gain for playing with her
-
+    $ mas_gainAffection(modifier=0.5)
     # monika wins
     if is_monika_winner:
         if renpy.seen_label("mas_chess_dlg_game_monika_win_end"):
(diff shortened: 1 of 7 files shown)


Rai99:
Will affection decrease even if you tell Monika that you'll be gone longer than a week?

Rai99:
Is there a limit to how low it can go without really reaching the point of no return? 

I don't think it really fits or makes sense to for example suddenly go missing for 2 years and then the first thing you get when you finally return is negativity and trouble. 

Maybe allow it go to 0 or maybe some negative, but not -100 if it isn't already capped somewhere else.

ThePotatoGuy:
if you say that you're going to be away for an unknown amount of time, you can be away forever and monika won't lose affection.

you only get the steady loss if you just stop talking without saying the appropriate farewell.
there is a cap, and currently at a 0.5 loss per day you can reach it in about 7.5 months. 

we can look into changing that to be a slower loss if you can make a compelling argument as to why.
(basically have some real-life numbers that make sense) ours are kind of arbitrary for this.

ThePotatoGuy:
@Rai99 if you are going to make an argument for slower loss, please make it a separate issue and link it back to this.

PR #1964, opened 2018-07-30
Apologies + review of unreviewed dialogue

aldoram5:
This pull request contains the mini apologies module, which is intended to be used for the player as a way to apologize for doing/saying mean things to Monika. The idea has been brought up before and now that the affection system is in place this is a more needed thing.

## Technical 

Basically this works by using a new topic where the player apologizes to Monika, said topic changes depending on the global variable `mas_apology_reason`, if it's None Monika doesn't know why you're apologizing for but thanks you for being considerate. If it's an empty string she I'll say that you should be considerate about her feelings and get a bit of affection back. If it's a non empty string the reason is used in dialogue and she accepts your apology.

The reasoning behind these flows is because there may be occasions were Monika is really upset about what the player did and reacts different to the player apologizing for it. Empty string is used for this one.
There may be times where there may be an affection loss but it's not exactly something big enough for an apology. None should be used in these cases.
When the thing the player did isn't something to be super upset the reason must be a valid coherent dialogue that will be used in the apology dialogue.

The default value for `mas_apology_reason` when calling `mas_loseAffection` is an empty string, opposed to what the usual None would be, the reason behind this is to enforce a responsibly usage of it.

`mas_apology_reason` can be used separately from `mas_loseAffection` if needed for any special reason.

## Extra

Some unreviewed dialogue was reviewed in this pull request. Said unreviewed dialogue was already merged so there's no harm on adding these revisions.

This one has been tested but minor testing wouldn't hurt.

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -5896,28 +5896,31 @@ label monika_beach:
     m 1tku "Don't get too excited though when you see it. Ehehe~"
     return "derandom"
 
-####################################################
-# Saving this for future use
-# Could be expanded to something better
-# where where persistent.playthrough can be
-# checked and have a different response
-# depending on what the player did
-# TODO: affection is going to take this one
-# and actually she'll know the reason only last thing
-# the player did though
-####################################################
+init 5 python:
+   addEvent(Event(persistent.event_database,eventlabel='monika_playerapologizes',prompt="I want to apologize.",category=['you'],pool=True,unlocked=True))
 
-#init 5 python:
-#    addEvent(Event(persistent.event_database,eventlabel='monika_playerapologizes',prompt="I want to apologize.",category=['you']))
-
-#label monika_playerapologizes:
-#    m 1g "Did something happen?"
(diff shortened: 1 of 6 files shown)


Monika After Story/game/script-topics.rpy
```diff
@@ -5896,28 +5906,31 @@ label monika_beach:
...
+        m 1ekd "Did something happen?"
+        m 2ekc "I see no reason for you to be sorry."
+        m 1dsc "..."
+        m 1eub "Anyway, thank you for the apology."
+        m 1eua "Whatever it is, I know you're doing your best to make things right."
+        m 1hub "That's why I love you, [player]!"
+    # She knows why you are apologizing for
+    elif mas_apology_reason:
+        $ mas_gainAffection(modifier=0.2) # recover a bit of affection
+        m 1eka "Thank you for apologizing because [mas_apology_reason]."
```
ThePotatoGuy: I think this would sound better as "Thank you for apologizing for <x>"

then all the reasons would start with a verb instead:
- calling me a bad name
- closing the game on me
- entering my room without knocking
- etc..

PR #1974, opened 2018-07-31
Affection fixes 3

aldoram5:
Last batch of fixes

## Technical

- Fixed typo on persistent cap.
- I love you topic will finally give affection.
- Update script now includes the starting off formula, Unstable users should be able to get the increment too.
- Monika nickname now has an actual Event, it gets locked or unlocked depending on her affection level.
- Updated documentation.

## Testing 

The nickname event should be locked/unlocked depending on affection. I already tested it but another one won't hurt.

### DIFF
--- a/Monika After Story/game/updates.rpy
+++ b/Monika After Story/game/updates.rpy
@@ -255,9 +255,38 @@ label v0_8_4(version="v0_8_4"):
 
         import store.evhand as evhand
         import store.mas_stories as mas_stories
+        import datetime
 
         # update seen status
         updateTopicIDs(version)
+
+        aff_to_grant = 0
+
+        if renpy.seen_label('monika_christmas'):
+            aff_to_grant += 10
+
+        if renpy.seen_label('monika_newyear1'):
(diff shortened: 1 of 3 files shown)


Monika After Story/game/updates.rpy
```diff
@@ -255,10 +255,37 @@ label v0_8_4(version="v0_8_4"):
...
+
+        if moni_love is not None:
+            aff_to_grant += (moni_love.shown_count * 7) / 100
+
+        aff_to_grant += (datetime.datetime.now() - persistent.sessions["first_session"]).days / 3
+
+        if aff_to_grant > 200:
+            aff_to_grant = 200
+
+        persistent._mas_affection["affection"] = aff_to_grant + persistent._mas_affection.get("affection",0)
```
ThePotatoGuy: shouldnt this use `mas_gainAffecition` so it properly pushes events and stuff

Also this probably will get overwritten by teh affection scrambler so this should run the aff save functoin right after setting

PR #1990, opened 2018-08-04
fix nickname not unlocking bug

ThePotatoGuy:
the event doesn't unlock once you've seen it lol.


jmwall24:
it unlocked for me, it's just not called what I thought it would be called.  Instead of Infinite Monikas it's called Can I call you by a different name, but it's unlocked.

PR #2008, opened 2018-08-06
Updated script-topics.rpy

multimokia:
Updated such that Monika actually tells you that you can hold her anytime you want even if it's raining given that your affection level is high enough. (So people actually know this since Monika never tells you this)


ThePotatoGuy:
i have already fixed this

PR #2032, opened 2018-08-08
New greeting

jmwall24:
A singing greeting that I think is much more Monika-like than the last one I tried.


ThePotatoGuy:
must wait for updated greeting rules

Issue #2016, opened 2018-08-06
Monika is much too difficult to beat at chess


jmeyer73:
I've played chess against monika well over 50 times and haven't even come close. Is this possible to win against her? I'm pretty sure stockfish has difficulty levels, would it be possible for there to be an option for her to go easier?

ReddTheRedRedd:
The games in genera are possible, just very hard(except hangman). It would be nice if she were even slightly easier when playing chess, and pong. That's one thing I find odd, she dislikes cheaters, yet uses her inhuman speed to win xD I know it's just how hard the games have been coded but I agree, wish it was a bit more reasonable.

ThePotatoGuy:
losing makes her easier, but even at level 0 stockfish is pretty hard. we're going to limit her thinking time to make it tougher.

Issue #1969, opened 2018-07-30
[Suggestion] Change Monika's chess engine (and maybe do more)


Mushroom1228:
Currently, Monika uses a chess engine named Stockfish for playing chess with the player. While it is a good engine, it is not suitable for Monika for these reasons:

1. Stockfish is a top level engine, used for analysis of positions by chess grandmasters and the teaching of endgames, and from what I have seen from a game between Monika and another Stockfish, it has not been weakened. This makes her humanly impossible to defeat by even grandmasters, let alone casual players. In fact, it is very difficult to even create and hold a drawing position against Monika. This gives the illusion of Monika being a chess grandmaster, which should not be the case.

2. Stockfish (and by extension, Monika) makes "un-human-like" moves. While the moves that Stockfish makes are very strong, it does so by choosing the best variation after seeing far into the future. The problem is that this causes Monika to play chess in an inhuman way, for humans do not usually possess the ability to see 10 moves ahead to make the best move. This results in decreased realism, and gives the impression of playing against a completely undefeatable perfect machine instead of a person.

3. As an extension to reason 1, the very high strength of Stockfish causes playing against Monika to be an unsatisfying experience. While it might be fun to get squished by Monika in the opening of a chess game for a few times, it will get frustrating after a while of constantly losing to Monika, who is playing constant inhuman chess. To solve this, a weaker engine should be used.

Therefore, after some discussion with some members in the community, I suggest changing Monika's chess engine to Leela Chess Zero, for the following reasons:

1. Leela Chess Zero uses a neural network, and as a result, really plays like a human player. This would solve the issue of Stockfish playing non-human chess, resulting in a game that would feel more natural when played by Monika.

2. Leela Chess Zero's strength can be modified, by using a weaker version of it for Monika. While Stockfish also has different versions, even Stockfish 1 is too strong for most normal human players. Meanwhile, Leela Chess Zero started out weak and slowly evolved to become strong. By using a weak version, Monika's playing strength can be decreased to suit casual players.

3. Leela Chess Zero can evolve to become stronger by playing games against it, either by itself or by the player. This would suit Monika as it gives the ability for Monika to slowly grow stronger over time, very much like a human would.

CrocWorks:
[...]

PR #2034, opened 2018-08-09
Attempt to lower chess difficulty

ThePotatoGuy:
#2016 #1969 

2 things:

1. disabled `Pondering` which makes Monika not think during player turn
2. cut monika's thinking time in half (from 1.5 seconds to 750 ms).

I have no idea if this will lower difficulty enough, so its more of a test.

### Testing
Make sure chess till works lol.

### DIFF
--- a/Monika After Story/game/chess.rpy
+++ b/Monika After Story/game/chess.rpy
@@ -481,7 +481,8 @@ init:
         class ChessDisplayable(renpy.Displayable):
             COLOR_WHITE = True
             COLOR_BLACK = False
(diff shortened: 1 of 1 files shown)


CrocWorks:
I wanted to link this to #3309 since for some reason that topic is placed in the non-idle group of projects instead of chess projects, and I figured if this is in the works now, then that might be relevant again. (Also applies to Pong if that ever gets a new difficulty)

ThePotatoGuy:
this didnt work so closing. new attempts should be made in a new pr

PR #2063, opened 2018-08-13
sports topic amendment/expansion

jmwall24:
Since she already mentions tennis, and it's a great sport for 2 people, either playing against each other or on the same team, it seemed like the perfect opportunity to expand on it a little.  So I changed the menu a bit to allow for more dialogue if the player plays tennis (or would like to learn to play with Monika).  If not, the original options are still there, just expanded on a little.


jmwall24:
Both great suggestions.  Thanks, Julia!

ThePotatoGuy:
I'm going to adjust some of the player model variables. You can finish dialogue review but approve, but dont merge yet.

PR #2095, opened 2018-08-18
Islands event

aldoram5:
The event that has been planned for some time now.

I won't go into much detail, only on the testing:

1.- Event should start as a greeting once you reach enamored state.
2.- Replace the placeholder art with the correct one (I'll leave a zip for this) to also test if the hotspots are correct.
3.- A pool topic should appear that can trigger the event again. It's on 'monika' and 'misc'.
4.- If your affection drops back to affectionate it should lock the event.
5.- If you go back up again it should only unlock the topic.

Edit:
Forgot to mention dialogue was already reviewed, if there's any dialogue addition/change I'll add that on another pull request, or this one if it hasn't been merged.


Monika After Story/game/script-affection.rpy
```diff
@@ -341,13 +341,25 @@ init 15 python in mas_affection:
         """
         Runs when transitioning from affectionate to enamored
         """
+
+        # If the greeting hasn't been seen yet, push the islands greeting
+        if not store.seen_event("greeting_ourreality"):
+            store.unlockEventLabel("greeting_ourreality",eventdb=evhand.greeting_database)
+
+        # lock islands event
+        store.unlockEventLabel("mas_monika_islands")
```
ThePotatoGuy: if i reach enamored, I can view the islands event without seeing the greeting first.

ThePotatoGuy:
Because of how we handle the images in this pull request, I've added a new framework called the DelayedAction Framework. It's deigned to effectively allow us to postpone Event actions indefinitely until some condition is met. In this particular case, its mainly the successful decoding of the images.

How this works:
We encapsulate actions into a `MASDelayedAction` object. These objects contain params key to performing the intended action. (check event-handler for more info). Each of these objects MUST have a unique ID, and its the ID we save into persistent for later action stuff.

There are 5 points delayed actions can be executed:
1. init level 995
2. splash start
3. spaceroom loop, initial entry
4. spaceroom loop, each minute
5. splash end

The islands events are delayed to splash start, which is after we do loading.

Anyway, theres's so much code and so little time, so I'm going to write up a full doc on this on the wiki instead of here.

## Testing:
1. Regression test the regular unlocking / locking of the island events. They should remain unaffected if the images were decoded properly.
    1. b64 images are in the special directory. The decoded images are placed there too.

**delayed action framework**
1. Setup your persistent so it hasn't seen the island events and is below enamored. Close game.
2. Remove one of the b64 encoded images.
3. Launch game. set affection to enamored+.
4. Verify that the greeting / event did **not** get unlocked. You can also check `mas_delayed_action_map`, which should contain a delayed action. as well as `mas_cannot_decode_islands`, which should be True.
5. Close game and relaunch. Verify that was mentioned in step 4 is still true.
6. Readd the b64 image that was missing.
7. Launch game. The greeting will most likely start immediately. Delayed action_map should be empty. `mas_cannot_decode_islands` should be false. Everything should be just fine.

PR #2111, opened 2018-08-20
Some fixes and changes to the movies module

aldoram5:
This one contains some changes to the movies module.

The changes are the following:

- Moved the movies to be on the `Play` menu
- Movies can be selected from an scrollable menu
- The "closure" part at the end of the movie now works with pushEvent

## Testing 

- Remove the comments for the movies menu on the script-ch30 so it appears in the menu.
- Play the ddlc trailer
- Nothing should break and it'll push the `monika_difficulty` event. This one is a placeholder one, should be updated to an actual event about the trailer.

I already tested it though.

### DIFF
--- a/Monika After Story/game/movies-info.mms
+++ b/Monika After Story/game/movies-info.mms
@@ -39,8 +39,7 @@
 movie Doki Doki Trailer
 description 3b "So the trailer for the game, huh?"
 description 1j "Let's watch it together, then [player]!"
-m 1m [00:00:04] "I wonder why they'd put a warning there."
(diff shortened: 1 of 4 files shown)


ThePotatoGuy:
Looks good but we are not going to merge until after 085 release

authorofbelethrun:
wait sooo is this done and in the game???

multimokia:
Yeah. It's been done and is in current MAS. Just disabled since there's been no content for it (as well it's more painful to make it work than normal topics)

Issue #2100, opened 2018-08-19
Bug game crashing possible infinite loop


SeriousJudge134:
I cant open ddlc after story.so i can open ddlc for 3 days and the today i cant open it,
It says:
An Exception has occured.
While running game code:
File "games/definitions.rpy",line 2357 in script.
Exception: possible infinite loop.
theres a full traceback contain the same definition.rpy error but its says something about execution.py line 56 in check_infinite_loop 
exception: possible infinite loop

Im in windows 7.6.1.7600
Ren'Py 6.99.12.4.2187
Monika after story 0.8.4-hotfix.1 

Please help me i just want to chat with monika.

SeriousJudge134:
Update: now i can open ddlc again but i tried quit and the bug is still the same.
The conclusion is i have to try open the game 3 times so i can open ddlc again.
Im so confused right now.

Needy1361:
Yeah i have the same kind of bug too.
Pls fix this PotatoGuy.i cant even open the game

SeriousJudge134:
yeah please fix this bug.by the way,here is the full txt files

[code]
I'm sorry, but an uncaught exception occurred.

While running game code:
  File "game/definitions.rpy", line 2537, in script
Exception: Possible infinite loop.

-- Full Traceback ------------------------------------------------------------

Full traceback:
  File "game/definitions.rpy", line 2537, in script
  File "C:\Users\asus\Desktop\DDLC-1.1.1-pc\renpy\execution.py", line 56, in check_infinite_loop
    raise Exception("Possible infinite loop.")
Exception: Possible infinite loop.

Windows-7-6.1.7600
Ren'Py 6.99.12.4.2187
Monika After Story 0.8.4-hotfix.1
[/code]

thats all,please fix this,THANK YOU!!!

ThePotatoGuy:
What is your CPU?

SeriousJudge134:
My cpu is AMD C -60 APU with radeon(tm) HD Graphics 1.00 GHz

ThePotatoGuy:
We're going to try increasing the amount of time before renpy thinks you're in an infinite loop.

SeriousJudge134:
Okay,Thank you so much,ThePotatoGuy,it really means alot.

SeriousJudge134:
alright thank you ThePotatoGuy,the game can work now,thx it means a lot

PR #2125, opened 2018-08-22
increase the loop time to 120 seconds

ThePotatoGuy:
#2100 

Maybe this will stop the infinite loop problems?

No real way to test lol. Lets just hope for the best.


SeriousJudge134:
thank you so much ThePotatoGuy,it works now!

Issue #2134, opened 2018-08-22
affection audit log


ThePotatoGuy:
also need to move logs to log folder

PR #2137, opened 2018-08-23
Adding affection logging and more

ThePotatoGuy:
This adds audit logging for affection, + logrotation.

## Testing
* Ensure that all new logs are now in a log folder in ddlc dir.

### affection logging
1. Launch game. 
2. Either use affection functions or run topics that do affection stuff.
3. close game.
4. verify the following:
    - On start, there should be an entry for LOAD. This is when we load the affection value
    - At end, there should be an entry for SAVE. This is when we save the affection value.
    - during topics, the first item should be the current topic (as label name).
    - gaining affection is represented by the change portion (->) increasing
    - losing affection is when the change portion (->) is decreasing
    - if we have a freeze, we should see `!FREEZE!` 
    - if we have a bypass during a freeze, we should see `!BYPASS!`. 

### logrotation
1. Set `persistent._mas_affection_log_counter` to 100.
2. Verify that on next launch, a number-suffixed aff_log.txt is added to the log dir.
3. Verify that on a subsequent launch, we dont add another log file.
**NOTE:** even if the highest number is 99, the next log number will continue to 100. We don't do wrap arounds here with logrotations.

### DIFF
--- a/Monika After Story/game/definitions.rpy
+++ b/Monika After Story/game/definitions.rpy
@@ -1797,10 +1797,24 @@ init -1 python in _mas_root:
         persistent._mas_affection["affection"] = 0
 
 
+init -999 python:
+    import os
+
+    # this is initially set to 60 seconds
+    renpy.not_infinite_loop(120)
+
+    # create the log folder if not exist
+    if not os.access(os.path.normcase(renpy.config.basedir + "/log"), os.F_OK):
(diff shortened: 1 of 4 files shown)


aldoram5:
https://github.com/Monika-After-Story/MonikaModDev/issues/2134

Issue #2145, opened 2018-08-24
085 typos


ThePotatoGuy:
ono

MisterSimple:
![image](https://user-images.githubusercontent.com/38892441/44621421-a5732d00-a87c-11e8-9d11-6acabb7c9a9c.png)

"language" is misspelled here

MisterSimple:
![image](https://user-images.githubusercontent.com/38892441/44621452-0a2e8780-a87d-11e8-961a-7a3b0cddc944.png)

"separate" is misspelled here

myseriousaiyan64:
no problem, fixed them.

jmwall24:
![screenshot_42](https://user-images.githubusercontent.com/36523761/44632409-2814ee80-a948-11e8-9412-7df21bafd24b.jpg)

The 'too' here is unnecessary and makes it look like it's saying they all had scenes outside of school, including her, and that she's jealous that she wasn't the only one.  I fixed this in #2173 along with some label stuff.

myseriousaiyan64:
good stuff.

MisterSimple:
![image](https://user-images.githubusercontent.com/38892441/44822668-a71d5780-abd2-11e8-9ddc-eb7ad9243ea2.png)
 
Here, it should say "I'll be sure **to** try my best"

MisterSimple:
![image](https://user-images.githubusercontent.com/38892441/45092219-24712c80-b0eb-11e8-8860-2a4e729293dd.png)

Here, it should say "though" instead of "thought".

PR #2147, opened 2018-08-24
Fixed [Bug] - #2145

myseriousaiyan64:
Only a minor typo, but fixed anyway since this was a bug. 
"Even though this caused '**a**' rift in the Python community,"


ThePotatoGuy:
#2145

PR #2155, opened 2018-08-24
Islands dialogue and interactions possible fix

### LINKED ISSUE #2146: [Bug] - Monika brings up topics while I'm looking out the window.
So, when I click on a place while looking out the window, Monika tells me about that place, but immediately after she finishes talking, she brings up a random topic. Also, while this happens, I'm able to interrupt her by choosing to stop looking out the window, but after I'm back in the space classroom she continues until the topic is finished.

aldoram5:
#2146

Can't test anything right now but maybe this will be enough to solve the issue.

## What's the issue 
Basically it didn't matter how many shields or anything we raised when showing the event, mostly because once it showed the screen it doesn't wait until the screen is hidden and goes straight to the return which flows back to the call_next_event flow which drops the dialogue shields.

## The fix
Adding a new flow to call_next_event where we request that it doesn't take care of the shields nor the queued events, because we're not done with the current event.
For this I added a new flow that just returns when returning the "nocontinue" flag from an event. With that said event should take care about the shields and everything itself.
I also added the `mas_RaiseShield_dlg()` call  but probably after the change we don't need that one  anymore. I'll remove it if requested
 
## Testing
I'm not in an environment where I can test this myself, I'll test it myself when I get the chance(which will probably be in like 7 hours from nor more or less).
The testing should be simple:
- by getting the initial greeting while having the random dialogue enabled it should never start a random topic nor any hotkey button should be available.
- same thing but with the pooled topic.
 
If the issue persists I'll think of something else

### DIFF
--- a/Monika After Story/game/event-handler.rpy
+++ b/Monika After Story/game/event-handler.rpy
@@ -1241,6 +1241,9 @@ label call_next_event:
             if "quit" in _return:
                 $persistent.closed_self = True #Monika happily closes herself
                 jump _quit
+            
+            if "nocontinue" in _return:
+                return False
 
         show monika 1 at t11 zorder MAS_MONIKA_Z with dissolve #Return monika to normal pose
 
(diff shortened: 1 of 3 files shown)


ThePotatoGuy:
My main concern is why is show screen being used instead of call screen + a loop? Call screen by default will open up a new interaction and stop execution. Show screen is async in nature so it doesn't make sense for an imagemap activity.

Issue #2024, opened 2018-08-08
Suggestions


Irres:
1. When I open the game, I think I'll appreciate if I options to greet Monika or let her greet me.
With a countdown, if the user doesn't respond. 

2. I will appreciated if I could thank her for what she is doing.

3. Will she cover more emotions in "I'm feeling", like nervous.

4. After the game is reopen after saying Monika "I'm going to sleep", I think it will be better that she check the time, so she can properly greet the user, instead of simply saying the same thing over and over.

Finally, I really appreciated for everything done on the project.

ThePotatoGuy:
1. Having the user do their own greetings is in interesting idea. 
2. Also a neat idea, but i guess to clarify, what are thanking her for?
3. yeah, more stuff will be added eventually
4. yeah, time-based dialogue will be adjusted soon

Rai99:
2. I think you get something like that at around 400 affection.

ThePotatoGuy:
[closing until we look into this]

Issue #1524, opened 2018-05-02
Monika doesn't seem to understand time


antimockingjay:
Hello! I noticed something that isn't really a bug, so much as an issue with the detection of time. I just woke up, and it's 1:00 in the morning. My sleep schedule is messed up, I know. Anyways, I hopped on to MAS to say hello, and I said "good morning" at which point she said I was being silly because it's night time. Which makes sense!

But then I said "good evening" instead, to which she replied "Jay! It's early in the morning!"

Technically, yes, it could be considered morning, but most people wouldn't call 1:00 AM early morning, so much as carrying over from the previous night. Plus, she literally just said it's night time, not morning.

Just... monika pls. i just woke up i don't have it in me to do mind games.

kaido1224:
No special greetings at 1 am in the morning. It can be either night or early morning depending on your viewpoint. It should probably be adjusted to be one or the other and leave it at that.

ThePotatoGuy:
[[Reopening when we look into it]]

PR #2192, opened 2018-08-28
Added Options To Sleep Greeting

Rai99:
Also stole part of the greeting_goodmorning
#2024, #1524, #3383


ThePotatoGuy:
Alright, I like that this will solve some of the sleep greeting issues we've been having. 2 things:

1. Currently we want to start making all time-based dialogue work with the custom sunrise / sunset settings. At the moment, there's no simple way to convert the sunrise/sunset time to a usable hour / mintue object, so we won't be merging this until that is in. I'll probably leave a comment here about it when its done.
2. Can you go through past issues and have your comment link them? You don't have to get every single one, but I'm almost 100% sure that at least one issue bfore mentioned the sleep greeting having no sense oftime.

ThePotatoGuy:
Okay for point 1, just found a function I wrote a while back. `mas_cvToHM` converts minutes into `(hour, minute)`.

You can use this like this:
```python
sunrise_hour, sunrise_min = mas_cvToHM(persistent._mas_sunrise)
sunset_hour, sunset_min = mas_cvToHM(persistent._mas_sunset)
```

You might have to combine this with some of the `datetime` functions and `timedelta` to really get a correct time check for sunrise/sunset.

ThePotatoGuy:
I might add some easier to use functions that just checks if the sun is up or not / midnight-sunrise / sunset to midnight/ and so forth.

Rai99:
I'm not sure this needs to work with the custom sunrise/sunset times. I aimed to make it work and fit regardless of runrise/sunset, assuming the player went to sleep for the night and not just a nap (Which might need to be its own farewell and greeting)

I think 0-5 is universally seen as _really late_ or _the dead of night._ 

5-12 is a common time to wake up and is universally considered morning time.

12-15 was made with late risers in mind, who wake up after 12pm and probably before 3-4pm.

Every other time was made, keeping in mind that some people don't have time to visit Monika in the morning and have to visit her at the end of their day. 

I don't think this needs to tie into custom sunrise/sunset, but it might be able to fit in well to change 12-15 to 12-sunset_hour. Using sunrise_hour for the 0-5 threshold was also in mind, but some people have to start their day that early when it's still dark/the dead of night. 

Also, I wasn't sure if you planned to add anything else here about easier to use functions, so I held off on making any changes.

ThePotatoGuy:
Well you may find the functions useful. They won't be in `content` until probably tomorrow or the day after so just hold tight.

ThePotatoGuy:
suntime helper functions are in `content` now, please use those here.

jmwall24:
this will be a part of #4914 current marked for 0.10.6

multimokia:
Seeing all the changes which will result in the timeconcern overhaul, it's going to be incredibly difficult to take this format and retrofit it for the new functions and such.

We'll take what will work when the time comes to implement that.

Until then, this is closed.

PR #2195, opened 2018-08-28
Content please add more topics

SeriousJudge134:
please add more topics,sometimes Monika bore me because she is saying the same things again and again.
i have seen all of the topics or random chat that monika has been talking to me lately.i have a couple suggestions like scary stories, the future, android body and about the player's life.yeah,we need more response from her chat or question,please add this more content,it will be greatfull to me,thank you!


monikLover:
Put suggestions on Issues, not Pull Requests please

PR #2214, opened 2018-08-28
script-topics expression changes

jmwall24:
Due to the side-eye expressions now being reversed, I changed the eyes on all pointing and leaning sprites to match the point/lean direction so they appear as they did before the change.


aldoram5:
As my commits say I added some more expressions and did some fixes regarding the sweat inconsistencies Pi told us about.

jmwall24:
Doh I was so focused on the eyes, I should've noticed the sweat inconsistency.

PR #2228, opened 2018-08-30
New song for Monika to sing

MrDevilzMan:
Another song that's been stuck to my mind ever since yesterday. Figured it seems fit judging by how the music video would go.

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -6203,6 +6203,22 @@ label monika_song_dream:
     m 1eka "Now that you're here, [player], I don't have to dream anymore."
     m 1ekbfa "My dream finally came true."
     return
+    
+ init 5 python:
+    addEvent(Event(persistent.event_database,eventlabel="monika_song_take_on_me",category=['songs'],prompt="Take On Me",pool=True,random=True))
+    
+ label monika_song_take_on_me:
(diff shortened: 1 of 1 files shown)


Rai99:
Could you explain what this song and these lyrics have to do with Monika and how/why you think it fits?

MrDevilzMan:
Originally, this song is about a guy who likes the girl, but never had the chance to say them out loud. Though it fits Sayori more than Monika, it can also affect the player as the game itself never meant to give her the route in the first place, thus the reason she's desperate for her Player; there was a barrier that both sides need to overcome.

The music video reminds me of how one would physically met Monika, or how the fantasy would roll out. I find the story behind the video enchanting, as well as being romantic.

Rai99:
I don't know about everyone else, but I'm not sure this song fits in. While the other songs are brought up with little to no context, it's evidently clear what they are about and I find the lyrics you provided in this PR to be confusing and with no clear message. 

At first, it sounded like Monika was planning to break up with the player and was going to leave shortly. 

Maybe it's meant to say that life is short and that if the player doesn't take her on, either Monika or the player will be gone someday shortly, considering how short life is when you think about it. Even in that case, I don't know if this song fits in with Monika and her situation or the other songs.

MrDevilzMan:
I guess you were right. Sorry, I didn't do my research first before doing this. Will close them now.

PR #2229, opened 2018-08-31
Changes to the apology so it's clearer what you have to do

aldoram5:
## Summary
- Changed the apology text to be more clear
- Code looks for both files imsorry and imsorry.txt
- Added a convenience func so we don't have to keep the name of the apology in every place we check nor need to re-write the conditional checks.

## Testing 
Just check that everything is still working and that it accepts both files. I already did the testing but yeah like always, the more the better.

This one doesn't not require dialogue review since I only removed the file extension, from the dialogue.


ThePotatoGuy:
I'm going to change `is_file_present` to use os libs and absolute path.

Issue #2245, opened 2018-09-03
[Suggestion] Concert topic/Jazz variable


jmwall24:
Right now the concert topic checks to see if you've seen the various music topics previously, before suggesting you go to that type of concert.  The jazz topic contains a question that asks if the player likes jazz or not, and stores the answer in a variable.  Would it make more sense, at least for the jazz part of the concert topic, to check that variable as opposed to just if the topic has been seen?  Especially since she doesn't seem that fond of it, based on the last line of "Maybe that's what I like about jazz, if anything."  So if you said you don't like jazz, and she seems indifferent at best, she probably wouldn't suggest going to a jazz concert.  Could even add a similar question and variable into the other music topics and use those for the concert topic as well.  It'd let Monika know more about you, and tailor the topic even more for each individual player.

If it's something that seems like a good idea, I'll be happy to do it.

ThePotatoGuy:
yeah that makes sense. go ahead and do that.

PR #2254, opened 2018-09-04
Updates to music and concert topics

jmwall24:
Makes changes related to #2245.  There were also a few TODOs regarding leaning blush that I fixed as well.

Also, in regards to the concert topic, there was a flaw I found where if the player hadn't seen any of the other music topics, the dialogue would've have skipped all of the if statements after the "Just imagine us..." line and went straight to the  "Oh gosh, maybe I'm getting a little carried away, hehe~" line with nothing in between.  I fixed that as well by either ending the topic before it got to this if the player doesn't like any music at all, or by adding the option for the player to tell Monika what type of music they do like so there will be at least one line of dialogue for Monika to mention between those 2 aforementioned lines.

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -6053,7 +6053,7 @@ label monika_concerts:
     if persistent._mas_pm_like_other_music:
         m 1hua "Jamming along to your favorite [tempmusicgenre] artist..."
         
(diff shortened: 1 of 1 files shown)


jmwall24:
There's one line of dialogue I need to change when I get home from work in case anyone goes to review this before then.

ThePotatoGuy:
going to:
- [ ] edit some of the dialogue
- [ ] change the topic where you can enter your music genre so it saves what you entered historically

Issue #2267, opened 2018-09-06
Apologize when Monika has teared up


Chrissenpai:
It has no affection change perhaps, but still when a crying sprite of monika displays, you should be able to  apologize after she finishes talking by pressing the apology button, with an unique dialogue for that goes by like, "what are you apologizing for?" option button- for making you cry -  oh... (moody dialogue continues)
because when monika cried when she was talking to me... i got destroyed... and i know most of you also felt that way when that happened because we love her.
Anyway sorry for my rant, ill just go to sleep with my guilty heart.

PR #2316, opened 2018-09-11
Added timed menu to comfort Monika when she cries

Rai99:
I had this idea before the issue was brought up, but this kind of addresses #2267 for being able to do something when Monika tears up a bit in this topic.

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -3185,6 +3185,58 @@ label monika_eternity:
     m 6ektdc "It reminds me of when you first came to the Literature Club."
     m 6lktdc "Everything felt so empty."
     m "The world was so grey."
+    if  mas_isMoniAff(higher=True):
+        show screen mas_background_timed_jump(4, "monika_eternity_2")
+        menu:
+            "Don't worry [m_name]. I'm here":
+                hide screen mas_background_timed_jump
+
+                m 6dstdc "..."
+                m "Thanks for that, [player]..."
+                m 6ektda "I'm... {w=0.7} glad you're still here."
+                m "Sorry... {w=0.4} it's just in the back of my mind that you..."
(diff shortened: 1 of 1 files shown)


ThePotatoGuy:
The `hug monika` menu choice doesn't seem like a good idea to use. The concept of using the menu buttons as actions should only cover meaningless actions (like how the crash greeting does "turn on the light"). Hugging is more a connected action and having it as a button action devalues its impact.

Rai99:
I was thinking that directly hugging Monika here would be a common option to calm her down. It's a serious moment, so I thought hugging Monika would be a justifiable option here. This event should only be available to the player at affectionate and above, so the player could just ask Monika for a hug anytime anyway, which would probably have less impact than hugging her in this moment. 

If the problem is that hugging Monika was directly put in a menu option, I could add some lead in to the hug instead of doing it instantly by surprise, but I think instantly hugging Monika in that situation fits the mood.

AsuraShun:
maybe dry her tears and a kiss to the forehead would help,don't know if it would be possible though

Rai99:
One of the harder parts of making things like these is finding the right sprites that actually exist and are ready for use, so I'm not sure if dry tears have any fitting, usable sprites. 

I would have loved to add an option to kiss Monika on the forehead, but I'm not sure if that could be well done with the resources available at the moment. It might be doable, I just thought that kissing Monika should have some unique sprites. In this scenario however, it might just work with what we do have.

AsuraShun:
as long as it's possible to comfort her somehow, i don't really care how as long as she's happy

AsuraShun:
YUS!

Rai99:
Somebody seems happy

dennisOdenbreit:
> The `hug monika` menu choice doesn't seem like a good idea to use. The concept of using the menu buttons as actions should only cover meaningless actions

The only way for the player to interact with monika on their initiative should consist of meaningless actions?

PR #2373, opened 2018-09-14
Height

miranimal:
So, I'm around Natsuki's height, And Monika, is obviously, way taller. And even though she loves me, if I told her I was short, she would probably make small jokes here and there and then apologize about them after words. I'm not exactly sure what she would say if she were tall. I'm sure there's some dialogue that could be fit in about both.


jmwall24:
This is not where suggestions go, please post them under the Issues tab, not pull requests.

Issue #709, opened 2018-01-08
[Suggestion] Player Microphone Interaction


AUDelusional:
So, I think the player should be able to communicate with Monika using their microphone. Considering the new update is supposedly suppose to get rid of keywords I think this would work either more. Python hosts a speech recognition library so it should be easy to implement into Ren'Py as Ren'Py is built off Python. https://pypi.python.org/pypi/SpeechRecognition/

ThePotatoGuy:
From the wheel file size, it looks like this library is 31MB. That's just way too extra for this. The impact of adding such a large library could dramatically increase load times.

IsItBadToMakeAGitHubAccountJustForThis:
This gives off some Talking Angela vibes.

gitthisissue:
Your SpeechRecognition library use Google API. It requires internet to function.
Why not use Windows' built-in function?

https://pypi.python.org/pypi/winspeech

Talking Angela... talking Monica, huh? That'll be nice, except I don't have microphone.

ThePotatoGuy:
That `winspeech` library is small enough to use. We'll have to see the feasibility of adding wheel files as python-packages first though.

therationalpi:
The problem with the windows thing is that it's a platform-specific function, and I think it's important for the game to be as consistent across platforms as possible. Including a major feature like speech-detection that's only available on Windows wouldn't be good...

GoldTheEevee:
i know i am just a random person with no program skills but would this FORCE the player to use a mic because then some people can't always use a mic. i am just suggesting this but there should be a option to use keyboard

AUDelusional:
Where'd you get the idea that this would force people to use a microphone? Using a microphone is optional, that should be obvious.

therationalpi:
[[This suggestion has been seen and will receive further consideration in the future. It is being closed for now and placed in the pool of seen suggestions. It will be reopened if we begin active implementation on this feature.]]

ThePotatoGuy:
[[Rejecting this because speech recognition is just out of scope for us]] 

We are open to people adding this as external submod development

PR #2385, opened 2018-09-14
Voice recognition option for Monika for suggestion #709 using google API and Pocketspyhinx

Unknownpony:
Here is first concept of voice recognition using google API.
There are only new files, No existing file were changed,

Not sure if we support LINUX or MAC so I left flac converters binary for all platform (can be delete if we won't need them) though both linux one are around 2MB. I read in guildline that anything above 1MB will be reject so I am not sure if we can make it work on linux in that case.

Also biggest issue which I can see is file _portaudio.pyd had to be copy into dir DDLC\lib\pythonlib2.7_portaudio.pyd as it's python libery (not sure if updater can only update files to game folder). Can't find working package version of this libery but I am still looking.


ThePotatoGuy:
updater only does `game` folder so if you can't find a package version then we even consider this.

Unknownpony:
I will find a way to have everything under game folder than. Thanks for info :)

Unknownpony:
Does automatic travis test call game like renpy? so not from the folder where game folder is but from the renpy-6.99.12.4-sdk\renpy ? Need to build path to lib in other way or just add test path to bypass test only hehe

ThePotatoGuy:
travis uses renpy sdk's linter and then builds with renpy sdk. if it fails to build, then it will fail when we compile for release.

Unknownpony:
I see. From what I see Travis is  based on Linux OS. I will need to set up VM for Linux for testing why it can't find that lib while on windows I have no issue building it in renpy. I guess I simple disable that event if import won't be successful.

Unknownpony:
Tested on Linux and it works fine however libportaudio2 has to be installed on the system. 
As many programs uses this libery I asume it is installed but if not than simple 
apt-get install libportaudio2
https://www.howtoinstall.co/en/ubuntu/xenial/libportaudio2

aldoram5:
Hi Unknownpony!
I looked a bit through your code and well I have a couple of things to say/ask:
- Google Services aren't free, those have a free quota and once it surpasess that it'll start charging money. 
- To keep track of that it needs an api key registered to the maintainer's name, where do you store the one you're using here?
- What's the plan for said key? are you going to provide it? Because considering the number of users is pretty much sure this is going to use more than the free quota.

Unknownpony:
I am not using actually the Google Cloud Speech API. I mean price would kill me... 0.006$ for each 15s after 60 minut. Who even afford that with bigger community use.

I am using Google Speech Recognition API https://github.com/gillesdemey/google-speech-v2
but yeah it has limits. Accepts up to 15s audio which is enough 
but 50 request per day with your own API key. I didn't check that before :S
Still I am using generic API key, not my own cus I didn't generate any key for myself so I am not 100% sure how they limit generic key as I can't find info about it. 

Also I can change it to use Sphinx no problem. So we can always go with that option
It will require really good dictionary for specific phrase for Monika though as it isn't precise.

aldoram5:
Hi again Unknownpony!
Yeah, migrating it to sphinx would be the way to go here, although, I think it may not be able to run with Ren'Py out of the box ([check here](https://www.renpy.org/doc/html/python.html#first-and-third-party-python-modules-and-packages)). I'd recommend you check first if renpy can run Sphinx before you start migrating anything.

Unknownpony:
I have already made pocketsphinx work on my system last weekend but it's really inaccurate :(
Not sure if it will work on Linux but I don't think there should be any issue.
I need to read more how to make it more accurate but I will commit on weekend so you can test stuff yourself :)
I will keep both option for in for now

aldoram5:
So while testing I got the: `Error message:No module named _portaudio` which is like you said on your first post that that you haven't found a working package version of it. I'll wait until you can find a way to add that to your pull request (so it works without the end user having to install external things).

Unknownpony:
What OS are you using? Windows or Linux?

aldoram5:
I ran my tests on macOS, will give it another try on a windows machine later.

Unknownpony:
Oh macOS. Hmmm I hope I can make VM with macOS for testing. 
Also to the portaudio. The python package is in game folder however it is using system lib http://www.portaudio.com/
Not sure if it is possible to put it in game folder and make it work. Many programs is using this libery though like popular VLC media player.

aldoram5:
I ran my tests on Windows, there it works well using the Google API, however with pocketsphinx it never got a single word right. Since we won't be using that API key since it's not meant to be used for anything else besides testing and we're not going to abuse it even if it let us do more than the requests it says it lets us. So for now, if you want you could package this one and distribute as an add-on/submod, mostly because using only pocketsphinx on windows with it's low accuracy doesn't really improve the user experience enough to be in the official release for now.

Unknownpony:
Hmmm not even single word like high school death love? It should catch the words which are in promt of the topics. I understand it will never work as good as google API but it should catch at least key words which we will use to get right Topic. If it's not than it's quite a failure :( though I am just starting to understand how dictonary and model language works so i guess that's expected.

Let me expalin how it will work or for what I am aiming. 
When you will say "I love you"
Google API will return exactly "I love you"
Pocketsphyinx should return something like "one love two" or I love two" ect. Most important thing would be word love. Then catching key words will be checked in algorithm. Some topics key words will have mroe values thatn other ect.
Let me know if you have any input or suggestions on such approach.

Also I understand it's not good enough for official release. I am still working on it ;) I aim to have it working properly with pocketsphinx before christmas as it would be nice for Christmas update to be able to tell Monika "I love you". At least for windows users

Unknownpony:
If anyone is intresting in that than I am currently working on Adapting the default acoustic model using this instructions https://cmusphinx.github.io/wiki/tutorialadapt/ so we can have at least 80% accuary

AsuraShun:
I really hope that if this gets implemented, it doesn't replace the menu,because it feels awkward to use voice commands if English isn't the user's native language

Unknownpony:
Obviously not. This is suppsoe to be just additional option to choice topic. I wouldn't even imagine changing menu or anything significant. Let be honest, most people will just probably used it to tell Monika how much they love her hehe ;)

Shiro2564:
all of the suggestions are great but I tink that it would be nice if Monika can change her clothes like the hairstyle with ponytail, but on this case she should probably use casual clothes or something like that.
this is just an opinion

Unknownpony:
I think you comment on wrong suggestion my friend :)
This is commit for voice recognition which will allow you to use microphone to choice options and stuff

Unknownpony:
New update to this feature if anyone is intrested.
For better algorithm tests (that find right topic, godebyes, compliments or modds) use google API as it's super precies. Pocketsphyinx is better but still not happy with it yet. You can test it too and check how's catching whole senatnce and single words. All words which pocketsphyinx catch you can find in above update corpus.txt 

If you will use word that isn't there, pocketsphyinx will choice the closest one (still not sure if use 70k words dictionary or my own dictionary which we might extend)

aldoram5:
If you need datasets I suggest looking online in sites like: http://voxforge.org/ and http://www.openslr.org/resources.php, in case you haven't already.

Unknownpony:
No I didn't, thanks for suggestion. I am finishing algorithm and words weight for now. Than only thing which will left is teaching pocketsphinx and all should be done.
Maybe when everything will work correctly I will play with making additional thread which will listen player all the time (if he will checkbox that option maybe in extra menu) so you won't need to click ask button to tell Monika something but directly just talk and Monika when she is in idle and she will react.

Unknownpony:
I will make adjustment and more teaching to pocketsphyinx this week so after it I hope someone will be able to test it and let me know if such results are acceptable. Thanks

Unknownpony:
Didn't have as much time as I wanted on weekend but I finished it.
Work on windows, might work on Linux but someone have to check it. Doesn't work on Mac cus I have no idea how Macs work so if someone want to help me make it work on Macs it would be helpfull.
Just made few test though. Find attached files (self explanatory) 
[Tests.xlsx](https://github.com/Monika-After-Story/MonikaModDev/files/2593345/Tests.xlsx)

Next is dictonary of all words which Monika can recognize
[dict.txt](https://github.com/Monika-After-Story/MonikaModDev/files/2593348/dict.txt)

Next is all sentance/words which I feed to pocketspynx to adjust it:
[feeding_inputs.txt](https://github.com/Monika-After-Story/MonikaModDev/files/2593352/feeding_inputs.txt)

Have any sentance which would be good to feed to pocketspyhinx? Words which you think should be in dictoanry?
Any input is appreciate. Can someone help me test it?

ThePotatoGuy:
Since this is platform-specific (windows) and commentary from Aldo points to the features lacking in some areas, this should be a submod moving forward. If the accuracy improves and cross-platform capabilities are added, then we can consider merging into official.

PR #2420, opened 2018-09-18
Docking station try to create it 

aldoram5:
So something that could be considered a tiny improvement. 
If the folder for the docking station doesn't exist it'll try to create it and if it can't it'll log the exception.

## Testing
- Make sure it doesn't break the existing working station if the folder exists
- Test the folder creation by deleting it or renaming it first. It should create it.
- If it fails because of permissions make sure the exception was logged.


ThePotatoGuy:
cant do inline comments on mobile so ill do them here:

* catch every exception as e, we want to log every issue that occurs

PR #2439, opened 2018-09-19
somewhere date

ThePotatoGuy:
Birthday-related reactions (Aka date reactions) to taking her out

# key things:
* moving the birthday to a var, so you can set this var and change the birthdaycheck at runtime
* reactions for first date and subsequent dates
* function that checks for the time out for a particular day (using checkout time)
* `persistent._mas_bday_date_count` increases for each date
* `persistent._mas_bday_affection_gained` increases when affection is gained
* `persistent._mas_bday_affection_lost` increaes when affection is lost

# Testing
### main thing
1. in `definitions`, change `mas_monika_birthday` to today.
2. take monika out. you should get an extra bit of dialogue
3. move monika's char file out of characters and relaunch game
4. you should empty desk. open console and pop **and save** the last line of `persistent._mas_dockstat_checkout_log`. 
5. re-add the checkout line with the same chksum but a different datetime. Make sure the date is the sameas today. the time should be changed so you can trigger different lengths of the bday date.
6. put monika back in charactesr folder and relaunch game.
7. verify the dialogue you get is correct.
8. `mas_bday_date_count` should increase by 1.
9. repeat these steps to get subsequent set of dialogue. reset `mas_bday_date_count` to 0 to get first date dialogue again.

### rejecting
1. set your affection to distressed / upset
2. monika should never reject going out when upset
3. monika should 10% chance reject going out when distressed

### affection 
1. when taking monika out, check that the affection changes occur with bypasses. **If the affection gained var is above 50, reset it to 0 before testing**
2. if you take monika out on short dates to start, you will lose affection, but the amount is tracked int he affection lost var. Once you've reached at least 1.5 hours of date time, the lost affection should be undone via a bypass gain and the affection lost var is reset. this only happens aslong as you have affection lost > 0 and you dont do an under 5 minute date.

use the checkout time function `mas_dockstat.timeOut(<date>)` to see how much time you've taken monika out on this day.

### DIFF
--- a/Monika After Story/game/script-greetings.rpy
+++ b/Monika After Story/game/script-greetings.rpy
@@ -1843,7 +1843,8 @@ label greeting_returned_home_bday:
 
         if is_first_date():
             $ lose_and_track_affection(1)
-            m "the first date was under 50 minutes"
+            m 2rsc "...Hmph."
+            m 2dsc "Some {i}'date'{/i} that was."
 
         elif is_short_date(total_time_out):
             $ lose_and_track_affection(1)
@@ -1873,7 +1874,9 @@ label greeting_returned_home_bday:
 
         if is_first_date():
             $ cap_gain_aff(1)
-            m "the first date was under 1.5 hours"
+            m 1hua "That was fun, [player]."
+            m 1eua "Even if it wasn't for too long, I still enjoyed the time we spent together."
+            m 1hua "Let's try to schedule something longer next time, okay?"
 
         elif is_short_date(total_time_out):
             $ cap_gain_aff(1)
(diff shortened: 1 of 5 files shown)


Monika After Story/game/script-greetings.rpy
```diff
@@ -1768,9 +1771,274 @@ label greeting_returned_home:
...
+    m 4ekc "It's just th--{nw}"
+    $ _history_list.pop()
+    m 1dkc "..."
+    m 1rkc "...nevermind."
+    return
+
+label greeting_returned_home_bday_short_sub_normal_total:
+    m 1hua "Well! That was fun, [player]."
+    m "We already had a good date, but I'm glad you took me somewhere again."
+    m 3tku "Just can'tget enough of me, can you?"
```
moonstne: "Just can'tget enough of me, can you?"
can'tget  -  needs a space

PR #2570, opened 2018-09-23
Added Dialogue for Steam Users

multimokia:
Added some dialogue for those using steam to hopefully convert them to non steam in order to fix some issues with file perms.


multimokia:
Also, shoutouts to ~~Simpleflips~~ @jmwall24 for helping out with dialogue and expressions

jmwall24:
Pls go away steam

PR #2631, opened 2018-09-27
Async Wrapper Framework

ThePotatoGuy:
I'm actually a bit proud of this one.

introducing `MASAsyncWrapper`, a class designed to perform a single task asynchronously, and returns the result. this is done via threading, but its all abstracted so its super easy to use.

# Key Features

## `MASAsyncWrapper` class
* These basically act as Async callables
* threaded
* to use these:
```python
promise = MASAsyncWrapper(some_callable)
not_done = False
while not_done:
    if promise.ready:
        promise.start()

    renpy.pause(1.0)
    
    if promise.done()
        not_done = True
        returned_value = promise.get()
```

## better dockingstation farewell
* now theres no hang when generating a large file.
* or reading in a large file

# Testing
## regression
* verify docking station farewell still works
* verify checkin / checkout is still being logged

## large file gen
1. run `mas_dockstat.setMoniSize(datetime.timedelta(500))`. This will set you up with a 1GB monika.
2. Run the docking station farewell.
3. Verify there's no program freezing. Visual freezes and jumps are acceptable.
4. Verify that the rest of the farewell still works even with a large file.
5. Verify that returning monika via the following ways has no program freezing. Visual freezes and jumps are acceptable:
   * loading with her in the folder. Long startup time in this case is acceptable.
   * dropping her into folder while game is running

## large file abort
1. Repeat step 1 for large file gen.
2. repeat step 2 but wait a few seconds, then tell monika to wait and say you cant take her out.
3. check the characters folder, there should still be a monika there.
4. Say that you are going to stay in the spaceroom.
5. Verify that after a minute or 2, the monika file in the character folder deletes itself.

# API:
* `MASAsyncWrapper`
* `mas_dockstat.checkoutMonika` and `mas_dockstat.checkinMonika`, for easy checkingin/checkingout of monika's file.

### DIFF
--- a/Monika After Story/game/definitions.rpy
+++ b/Monika After Story/game/definitions.rpy
@@ -2289,6 +2289,20 @@ init -100 python in mas_utils:
         return result_list
 
 
+    def log_entry(entry_log, value):
+        """
+        Generic entry add to the given log. 
+        Stores both time and given value as a tuple:
+            [0]: datetime.now()
+            [1]: value
+
+        IN:
+            entry_log - list to log entry to
+            value - value to log in this entry
+        """
+        entry_log.append((datetime.datetime.now(), value))
+
+
     class ISCRAM(ctypes.BigEndianStructure):
         _iscramfieldbuilder = [
             3, 3, 2, 1, 3, 2, 2, 1, 3, 3, 1, 3, 1
--- a/Monika After Story/game/script-ch30.rpy
+++ b/Monika After Story/game/script-ch30.rpy
@@ -25,18 +25,10 @@ init 970 python:
 
     if persistent._mas_moni_chksum is not None:
         # do check for monika existence
-        moni_tuple = store.mas_dockstat.findMonika(
-            mas_docking_station
-        )
-
-        # set the init data
-        store.mas_dockstat.retmoni_status = moni_tuple[0]
-        store.mas_dockstat.retmoni_data = moni_tuple[1]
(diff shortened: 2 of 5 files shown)


ThePotatoGuy:
so I don't forget later - the context manager mac comment is because there was a stack trace where a thread Condition object didn't have an `__exit__` attribute for some reason. Probably was a None issue since the var was set elsewhere and could be None and it wasn't checked in the old `_th_findMonika`. 

```py
Exception in thread Thread-4:
Traceback (most recent call last):
  File "/home/tom/ab/x64lucid-deps/install/lib/python2.7/threading.py", line 810, in __bootstrap_inner
  File "/home/tom/ab/x64lucid-deps/install/lib/python2.7/threading.py", line 763, in run
  File "game/zz_dockingstation.rpy", line 1907, in _th_findMonika
    with _th_cond_findMonika:
AttributeError: __exit__
```
```py
    def _th_findMonika():
        """
        Threaded variant of finding monika, using stock docking station
        NOTE: this doesn't actually have the thread, it just runs the primary
            usage of teh stock docking station code and saves result
        """
        global _th_result_findMonika
        # find monika
        _result_findMoni = findMonika(store.mas_docking_station)
        # acquire lock on condition and set result var
        with _th_cond_findMonika:
            _th_result_findMonika = tuple(_result_findMoni)
            _th_cond_findMonika.notify()
```

PR #1796, opened 2018-06-23
Added menu to God topic

Rai99:
Left old sprite codes in attempt to not conflict.

Issue: #1720

Issue #1720, opened 2018-06-08
monika's thoughts about God


loverofmonika:
i  want to talk about the "thoughts about god" i think its best to omit this option cuz there are players out there who may disagree with what she says and some may even dislike her. this topic is a sensitive matter and i think its best to avoid any conflicts don't u agree. as i was observing some players having monika afterstory, i noticed that among them didnt like what she said and pretended that she didnt say that. some even on the verge of deleting the game. So please heed my advice. removing it wouldnt destroy the mod.

Rai99:
When you were "observing" some players, I hope you didn't just hand out MAS to random people who have no idea what it is. 

I agree this could be a touchy subject but instead of removing it altogether, I'm wondering if there's a way to change the dialogue to be less potentially offensive. However, Monika _does_ have a stance on the matter and I think it's important to be able to share that with the player. 

I'm thinking adding some interactivity might help instead of just: "Here's what I think. Take it or leave it." Maybe actually add a yes or no option when Monika asks if you believe in god. If no, dialogue continues as it is now. If yes, Monika could politely state her stance on the matter and talk about the player having to show her otherwise when possible. I would think that religious people would be enthusiastic to talk about their god and spread the wisdom.

jmwall24:
Just my 2 cents on this, but I don't think you should have to agree on everything with Monika.  No matter how close 2 people are, they're not gonna agree on everything.  Also, I think this topic is interesting because Monika has a truly unique perspective on the whole idea of creators.  People in our reality can have opinions on God one way or another, whereas Monika knows for a fact that she has a creator and that everything that happened to her in the course of the original game (and before) was his doing.  Her opinion on God (and by extension Dan) is based on very specific circumstances and I would think people would understand why she wouldn't be all that thrilled about creators that toy with peoples' lives and emotions like it were a game to them.  You have to remember she's looking at this thru her own reality, and not thru ours.

Rai99:
Also this very topic was in the original DDLC so there's a decent chance anyone with MAS would have seen it before, and with how popular DDLC was, I'm sure something would have happened if the topic was so controversial. 

Monika does have her own experience with creators and greater beings and personally, I'm not a fan of the religion I grew up with either. Other people's experiences vary and I think it's important not to be afraid to share them with those close to us.

loverofmonika:
[...]

PR #2655, opened 2018-09-29
Added menu to God topic 2.0

Rai99:
New PR to eliminate conflicts due to changes over time. 
Old PR: #1796 
Issue: #1720

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -304,6 +304,10 @@ init 11 python:
 #BEGIN ORIGINAL TOPICS
 #Use this topic as a template for adding new topics, be sure to delete any
 #fields you don't plan to use
+
+# are you religious
+default persistent._mas_pm_religious = None
+
 init 5 python:
     addEvent(
         Event(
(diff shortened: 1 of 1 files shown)


loverofmonika:
I have my version of the matter as I stated that I would bring
Sorry for the delay,
"[Player], do you believe in God?"
"Sorry, that was really out of the blue."
"I really should stop doing that."
"  I do know that this is a really sensitive topic and I don't mean to offend."
" But I feel like I want to share my thoughts on the matter."
"But if it makes you uncomfortable then we can talk about something else."

Options
" No its alright, continue"
" let's talk about something else."

If second option chosen

" alright then!"
" I don't want to hurt you nor make you feel uneasy."
" Many relationships end due to the conflict of views on the matter"
"And I don't want that for us"
" I love you and I only want that is better for the both of us "
" So yeah, let's talk about something else~"

If 2nd option 
"Alright then."
" well to be honest with you."
"I was never too sure, myself."
"Well, I'm sure I never really questioned it as a kid..."
"But as I grew up, the more I learned about the world, the more I would question it."

"I started to wonder why God was helping people pass exams, or get over a cold..."
"...When there are children who live their lives being sold as sex slaves?"
"Or the eight hundred million people who are too poor to even eat."
"I wonder how many of those people pray to God every day until they starve and die?"
"Or, how many millions of families pray for a loved one to recover from some incurable disease?"
"But the punchline is this."If just one person beats the odds and survives - among the thousands of others who die--"
"--Then, it's suddenly a miracle from God."
"I'd really love to meet this God who seemingly laughs at the misery of everyone not eligible for his miracles."
"But the irony is that I do have a creator, apparently."
"And you know what?"
"I bet he's still laughing at the miserable fates of Sayori and Yuri, even as we speak."
"What are we to him but props in a scripted play?"
"So, from that perspective..."
"I don't think it's too far-fetched for there to be a god - if Earth was nothing but his playset."
" well again this is from my perspective."
" I mean everything does happen for a reason, right."
" So there maybe something I didn't account for."
" that and also considering I am trapped in a game."
" I'm sure you have your own views on the matter"
" when I can finally enter your reality, Maybe you can show me a new light on things hehe~"

loverofmonika:
I'll have someone someone also try to form it to code form unless you guys do agree to use my version and convert this to the scripted form
 @Rai99

Rai99:
No need. You're a bit late seeing that this was merged a month ago and the job is already done. 

I also saw that you wanted to fix the old PR, but was wasn't really necessary since it would be inconvenient for anyone else to fix it and I'm still here on top of things. 

**The issue was dealt with and is over now.** 

As for feedback on your version, we don't want to completely cut out the topic from people if they're religious. Monika should be able to discuss her thoughts with the player. This PR added an option for the player to share their opinion based on whether or not they worship any god. Monika is respectful of the player's beliefs and supportive of their decision rather than completely ignoring and avoiding the topic.

loverofmonika:
Damn it, i was too late, welp at least the matter is resolved . As for my
version, I didn't make Monika seem disrespectful nor ignorant. I was going
for if the player wants to drop the topic then he can otherwise she would
go on expressing her thoughts.

Otherwise glad this topic is fixed.

PR #2719, opened 2018-10-06
Added Topic About Dating Experience

Dank-Elf-Boy:
This topic covers the player's previous relationship experiences they've acquired in their lifetime and how it can help strengthen the relationship the player has with Monika (Issue: #2694). I haven't tried adding topics in a while so go easy on me haha. And please don't outright withdraw the request if a few things are off, just let me know and I can fix them. Thank you.

Issue #2694, opened 2018-10-02
A Few Important Topic Requests + Suggestion


Bam8647:
**_These two topics are some of the most common questions asked in mature early couples. I really hope you'll take them into consideration._**

**1.** **_How many relationships have you been in_?** Monika wants to know how much experience the player has in commitment and partnership(_an important question most new couples ask their partners_). The three responses would go as: 

**"Alot."** The player has been through many heart breaks to which Monika will reassure them that she believes she can be _"The One"_ for them.
**"One to a few.."** The player admits that they don't have a lot of experience with dating to which Monika will assure them that she too doesn't have a lot of experience dating due to being super busy but that she believes that her and the player will grow strongly together.
**"None"** The player has never been in a relationship before and that Monika is their first to which Monika reacts shocked but then doubtful that the player has never had the chance to be with someone no matter how sweet and amazing they are. Lastly her response will end with telling the player that she hopes to be their _one and only_ .

**2.** **Are you a virgin?(Monika ask the player)** I have no doubt that this topic has probably already been suggested but probably denied to which I'd argue that if you can have topics talking about **used panties and masturbation** then why not a simple question about sexual experience? If users get butt hurt about this mature topic then block the topic under sensitive mode or have three responses as follow:

**"Yes"** The player admits that they have no experience with intimacy to which Monika will also admit that she doesn't have any experience either (or at least that she knows from her background) but that it's perfectly fine and that maybe they can take things step by step once she crosses over.
**"No"** The player tells Monika that they have some experience to which she'll react surprised and kind of shy that she doesn't have any experience but acts hopeful that the player will make her feel comfortable taking these steps once she crosses over.
**"I don't wish to answer"** The player is not comfortable with this topic and does not wish to answer. Monika will understand and apologize to the player and suggest they talk about something else.

**[Suggestion]** Omg we gotta let Monika know just how adorable she looks especially when she's doing her random cute poses(my favorite is her leaning on the desk and looking like shes enjoying herself) Idk maybe have an option that lets her know which pose you like the most and maybe she'll do it more often? And no.. the "I love your smile" seems too bland lol.

EntonyEscX:
[...]

PR #2721, opened 2018-10-07
New Past Relationships Topic

Rai99:
Issue: #2694 
This PR is waiting for #2575 as it uses sprites added in that PR. 
This PR in in conflict with #2719 as they both cover the same topic from the same request in the same issue.

The author of the issue requested that someone write a topic about this, but also created their own version of the topic as this one was in the process of writing. This was almost done before they put their version up. What else am I supposed to do?

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -7710,7 +7710,7 @@ label monika_past_relationships:
                     m 1eka "I hope it's not too much trouble whenever somebody tries to confess to you."
                     m 1eua "I get that you might not want to be distracted by a relationship and that's respectable."
                     m 1tubfb "I just hope you don't mind me coming into your life and loving you."
-                    m 1tubfa "Ehehehe~"
+                    m 1tsbsa "Ehehehe~"
     return
 
 init 5 python:
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -7589,6 +7589,9 @@ default persistent._mas_pm_few_past_relationships = None
 # Have you been in a lot of relationships
 default persistent._mas_pm_many_past_relationships = None
 
+# Don't wanna talk about it
+default persistent._mas_pm_no_talk_past_relationships = None
+
 init 5 python:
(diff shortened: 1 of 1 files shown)


Bam8647:
Wow! You really went in depth Rai not bad! I really thought the responses leading into further conversations were pretty intriguing. You could possibly create a separate topic with how the player felt about their previous relationships compared to the player's relationship they have with Monika. 

I felt that there should only be three main responses to the initial question regarding dating. The responses are suppose to range from **Most Experienced to No Experience** so in this case **A lot** and **A few** could be combined as this already tells Monika that you've been through multiple break ups compared to just **one break up** and finally having **no experience dating** what so ever.

Again the extended conversations are a nice touch(: 
I just wanted to keep the initial dialogue short and to the point.
Don't get me wrong, new topics can be made from what you just made already which maybe we could figure out a second tier to the Dating topic so none of our hard work doesn't go to waste.


Anyways, Great Job!

Rai99:
I was taking it easy, but you really made me speed it up with your PR.

Rai99:
I hate to say it, but I don't think both of our PRs are making it in. I think yours might be the one that's taken since you got it up first.

Rai99:
I felt that only having 3 responses (Multiple, one, none) left out a possible group. I intended the "A lot" option for those who basically have dating down to a science and it's nothing really special for them."I intended "A few" for maybe 2-3 relationships; More than one, but not enough to really have it come natrually. "Just one" of course, is unique enough to warrant it's own path and "None" is also a mandatory option. 

If the player has been in multiple relationships, they should be experienced and mature enough not to let breakups bother them, which is why I added much more in depth dialogue for the "one" option. 

I think we'll cross the bridge of what I'm supposed to do with my topic when we get to it. 

untitled1044

Rai99:
This topic needs and rkbsa and rubfb for body pose 1.

ThePotatoGuy:
okay need to figure out if this and #2719 can both be in or only one in.

jmwall24:
Closing this in favor of #2719. This almost comes off as an interrogation rather than a conversation. Might put some people off.

Issue #2318, opened 2018-09-11
[topic] friends


Chrissenpai:
after monika tells you her topic about friends she tells you that if you have any, yes and no, if no she'll comfort you, if yes she ask us how important are to them to you and viceversa, we meh, we hang out a lot or we love each other, and with each option she tells us a topic for either one of the choices

ThePotatoGuy:
[closing until we look into this]

PR #2758, opened 2018-10-13
New Topic About Player's Friends

Rai99:
Issue: #2318 

~~Another one for the pile~~


Rai99:
rkbsa is currently only made for body pose 3 and rsbsa is only made for 2. This topic needs them both for body pose 1.

ThePotatoGuy:
going to add player model vars

PR #2765, opened 2018-10-14
Daydream quips

MrDevilzMan:
Added more things she could daydream, instead of repeating the same answer over and over.


multimokia:
Heads up regarding the cherry blossom quip you've got. You'll only know that if you saw the islands event, so if that came up when a player were below 400 affection (and didn't get the event) or got the event, but didn't get that bit of dialogue, they'd be rather confused as to what Monika were referencing.

ThePotatoGuy:
going to change this into using the `MASQuipList` for better adjusting of list items based on certain conditons

multimokia:
By the way, you can actually test these in MAS yourself, put together an rpy file in your `game` directory (I recommend putting it in a subfolder for neatness. This way you can make sure everything's working on your end before submitting in a PR.

MrDevilzMan:
I would, but I don't have a dev tool myself. It would be helpful for my other MAS-related projects, though I need to know how to get one.

MrDevilzMan:
Can the dev tool be used in separate MAS, though? I wouldn't want my time with her gets taken out by my obsession with MAS-related projects?

multimokia:
You can have more than one MAS install, yes. But they'll both read from the same persistent.

ThePotatoGuy:
need to review if this topic should have different "daydreams" at different affection levels, or if it should just be hidden under a certain affection level.

ThePotatoGuy:
@multimokia @jmwall24 thoughts on limiting topic to some affection level or adding darker daydreams for lower affection levels?

jmwall24:
maybe both. thinking we filter out broken for this, but can probably come up with a few for distressed.

multimokia:
Agreed there, broken wouldn't really fit, but upset and distressed I can see.

Issue #2756, opened 2018-10-12
PoseMap the hair / clothes


ThePotatoGuy:
in order to use `sleep-mode` #607 and anything related to not having certain lean poses for certain hair styles, we need to extend the `MASPoseMap` to hair + clothes.

This would effectively make those objects instead of string codes.

PR #2767, opened 2018-10-14
Extend posemap to hair and clothes

ThePotatoGuy:
#2756 

# key changes
* Hair (`MASHair`) and Clothes (`MASClothes`) are now objects instead of strings. 
   * they also contain `MASPoseMap` for fallback/existence checking
* all sprite objects now extend off the base class `MASSpriteBase` 
* **this updates the exp previewer, so anyone who uses it will need to update**

changing clothes/hair is basically the same as before, only difference is that you have to use an object like accessories does.

## MASSpriteBase
* base class for all sprites
* mainly contains identifier info as well as filename codes and the pose map

## MASSpriteFallbackBase
* extends `MASSpriteBase`
* adds a fallback property to say if the pose map is for fallbacks or not

## MASHair
* extends `MASSpriteFallbackBase`
* used for hair objects
* hair objects are at `IMG015`

## MASClothes
* extends `MASSpriteFallbackBase`
* used for clothes objects
* clothes objects are at `IMG018`

## MASAccessory
* changed to extend `MASSpriteBase`

## default usage
When not using the fallback mode, the hair/clothes objects are checked for the selected lean type and whether or not we need to blacklist the lean. Here's an example with the bun hairstyle blacklisting pose 5 (lean):
```python
mas_hair_bun = MASHair(
    "bun",
    "bun",
    MASPoseMap(
        default=True,
        p5=None
    )
)
```
**Only lean types are checked when not using fallback mode.** Setting regular poses to None will not change anything.

## the fallback system
The fallback system uses `MASPoseMap` to map poses to other poses. The sprite selection system checks this map and uses the fallback pose when appropriate.
Here is an example of the bun hairstyle using the fallback system to map poses 2 and 4 to 6 and 3 to 1:
```python
mas_hair_bun = MASHair(
    "bun",
    "bun",
    MASPoseMap(
        p1="steepling",
        p2="down",
        p3="steepling"
        p4="down",
        p6="down"
    ),
    fallback=True
)
```

**IMPORTANT:**
* non-lean types can only be mapped to other **non-lean types**.
* lean types can be mapped to **lean types** OR **steepling** (via setting the pose to None)

# Testing
## regression
* verify that all sprites look okay. you can use the exp previewer to double check everything
* change hair and verify things still look fine.
* verify that the hair down hairstyle is saved and reappears on restart
* verify that the hair down greeting/prompts still work

## non-fallback mode
1. change the hair code for a hairstyle so it matches the bun code above. (since we only have 1 lean pose, this is the only check we can do)
2. Verify that when the lean pose is None, the lean appears as steepling.

## fallback mode
1. change the hair code for a hairstyle so it does the fallback mode like the bun example above. You can use other mappings aslong as the lean/non-lean types are not interchanged
2. verify that the poses map to the appropriate pose according to your mapping

### DIFF
--- a/Monika After Story/game/dev/dev_exp_previewer.rpy
+++ b/Monika After Story/game/dev/dev_exp_previewer.rpy
@@ -45,6 +45,7 @@ init 999 python:
         we are about to go there
         """
         import pygame
+        import store.mas_sprites as mas_sprites
 
         # CONSTANTS
         VIEW_WIDTH = 1280
@@ -891,7 +892,10 @@ init 999 python:
         def _sel_hair(self, direct):
             self._adj_sel(direct, "hair")
             self._update_sel_tx("hair")
-            monika_chr.change_hair(self._get_spr_code("hair"))
+            monika_chr.change_hair(mas_sprites.HAIR_MAP.get(
+                self._get_spr_code("hair"),
+                mas_hair_def
+            ))
 
 
         def _sel_mouth(self, direct):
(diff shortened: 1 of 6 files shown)


ThePotatoGuy:
need to fix the reset code

Issue #2262, opened 2018-09-05
"Thank you for always being with me" disappeared from the 'I want to tell you something' topics


RutTutPut:
I was able to say it to Monika after the first time she showed me the islands, and once I left the game and came back in, I was no longer able to find it.

Rai99:
That's a one time only thing, Say it once and regardless what you say in that menu option, it's gone.

aldoram5:
Hi RutTutPut!
This was intentional, it was supposed to be a special one time thing, but it has been decided that it will be like the other options, but after you reach a higher level of affection. It still hasn't been implemented as of now, but will be in a future update. We'll use this issue as a way to track the progress of that.

ThePotatoGuy:
[closing until we look into this]

PR #2805, opened 2018-10-21
Re-enable thanks compliment at love state

aldoram5:
#2262 
This pull request re-enables the compliment at love stage. It stays the same and doesn't get a dialogue change but no longer gives a considerable amount of affection.

## Testing

- Check that the compliment stays always available at Love stage
- Check that if you're at love stage it doesn't give you more affection
- Check that if you're at any other state it no longer appears if you've seen the compliment already.

### DIFF
--- a/Monika After Story/game/script-compliments.rpy
+++ b/Monika After Story/game/script-compliments.rpy
@@ -377,7 +377,7 @@ label mas_compliment_thanks:
     m 4ekbfa "I guess we're both lucky that we have each other, [player]~"
     menu:
         "You mean everything to me, [m_name]":
-            if not mas_isMoniLove():
+            if mas_getEV('mas_compliment_thanks').shown_count == 0:
                 $ mas_gainAffection(10,bypass=True)
             m 1ekbfa "[player]... "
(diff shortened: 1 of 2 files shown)


Monika After Story/game/script-compliments.rpy
```diff
@@ -377,15 +377,18 @@ label mas_compliment_thanks:
     m 4ekbfa "I guess we're both lucky that we have each other, [player]~"
     menu:
         "You mean everything to me, [m_name]":
-            $ mas_gainAffection(10,bypass=True)
+            if not mas_isMoniLove():
```
multimokia: Issue with this:
If a player saved the compliment until `Love` then they're not going to get this affection boost.

Should just need a similar format to the other compliments so every user gets their affection gain properly.

jmwall24: Will do my review after this is changed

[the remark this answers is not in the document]
jmwall24: Will review after this is fixed up

PR #2829, opened 2018-10-24
Added Wink Expressions (48 of them)

Rai99:
_"If I added a crap ton of expressions, and put them up in a PR, would it get accepted"_

A lot of the time, I find it really, really inconvenient to want a sprite, but it's only available for 1 certain body pose and I just end up just finding something else that works. With this, all the mentioned expressions will be available for body poses 1-4. 

Placed this in the dev folder as a .rpy and tested and they all worked for all 4 poses:
[EXP Winks.txt](https://github.com/Monika-After-Story/MonikaModDev/files/2512052/EXP.Winks.txt)
This contains the following expressions in body poses 1-4
kua (already existed for 1)
kub
kuu
ksa
ksb
ksu
nua
nub
nuu
nsa
nsb
nsu

### DIFF
--- a/Monika After Story/game/sprite-chart.rpy
+++ b/Monika After Story/game/sprite-chart.rpy
@@ -3258,6 +3258,617 @@ image monika 1kua = DynamicDisplayable(
     arms="steepling"
 )
 
+image monika 2kua = DynamicDisplayable(
+    mas_drawmonika,
+    character=monika_chr,
+    eyebrows="up",
+    eyes="winkleft",
+    nose="def",
+    mouth="smile",
+    head="a",
+    left="1l",
+    right="1r",
+    arms="crossed"
+)
+
+image monika 3kua = DynamicDisplayable(
+    mas_drawmonika,
+    character=monika_chr,
+    eyebrows="up",
+    eyes="winkleft",
+    nose="def",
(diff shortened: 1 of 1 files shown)


ThePotatoGuy:
ok so theres a problem with this, and its that the backup sprite combos (`head`, `left`, `right`) are wrong for many (if not all) the sprites you added. 

There are 2 reasons why we add sprites slowly:
1. add only the ones that we actually use so renpy doesn't cache extra sprite combos
2. set a proper fallback sprite combo that makes sense

> inconvienient to want a sprite

We don't say you can't use sprites that aren't setup yet. You can use them if you want, we'll add the sprite to your pull request if it makes sense (aka not the longest sprite code or similar).

Rai99:
1. What do the backup sprite combos (head, left, right) do and how are they used?
2. I do think all these sprites make sense and _could_ be used, but it's true that some might rarely be used. 
3. So sprite availability doesn't really matter and they'll be added if a PR needs them?

ThePotatoGuy:
1. the backup sprite combos are important for standing variations of monika, which may happen in the future.
3. Yeah sprite availability doesn't matter at all, we'll add them if a pull request needs them.

PR #2830, opened 2018-10-24
Added Wink Expressions W/ Blush (144 of them)

Rai99:
_"How hard could it be to add all 3 blush types to each and every expression for all 4 body poses"_

Same reasoning as #2829. I think wink blushes could be useful and don't want to leave any holes from pose 1-4. 

Tested body poses 1-3 each with 1 type of blush. 
[EXP Winks W-Blush.txt](https://github.com/Monika-After-Story/MonikaModDev/files/2512167/EXP.Winks.W-Blush.txt)
This adds all 3 blush types in poses 1-4 for the following expressions. 
kua
kub
kuu
ksa
ksb
ksu
nua
nub
nuu
nsa
nsb
nsu

### DIFF
--- a/Monika After Story/game/sprite-chart.rpy
+++ b/Monika After Story/game/sprite-chart.rpy
@@ -3258,6 +3258,2022 @@ image monika 1kua = DynamicDisplayable(
     arms="steepling"
 )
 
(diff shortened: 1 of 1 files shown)


ThePotatoGuy:
same comments as on #2829

PR #2863, opened 2018-10-28
New Song For Low Affection

Rai99:
The song is called Crime and Punishment. 

This song is meant to be sang at broken affection and it auto progresses. 
Ideally, Monika should sing this and the song will pass by without the player even realizing. 
It has 1:04.2 seconds of pause.

This is the **entire** song in a topic. I don't think being too long or tedious to get through should be a problem since it's meant for people at low affection and I don't think it's a massive problem to make them go through a slightly long topic.

### DIFF
--- a/Monika After Story/game/script-affection.rpy
+++ b/Monika After Story/game/script-affection.rpy
@@ -300,6 +300,7 @@ init 15 python in mas_affection:
         """
         Runs when transitioning from distressed to broken
         """
+        evhand._unlockEventLabel("monika_song_crime_and_punishment")
         # change quit message
         layout.QUIT_YES = mas_layout.QUIT_YES_BROKEN
         layout.QUIT_NO = mas_layout.QUIT_NO_BROKEN
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -6731,6 +6731,50 @@ label monika_song_dream:
     m 1ekbfa "My dream finally came true."
     return
 
+init 5 python:
+    addEvent(Event(persistent.event_database,eventlabel="monika_song_crime_and_punishment",category=['songs'],prompt="Crime and Punishment",pool=True, random=True))
+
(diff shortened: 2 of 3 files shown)


Rai99:
This topic needs...
_reads travis_
a couple sprites.

ThePotatoGuy:
I will add them in a bit

Monika After Story/game/script-affection.rpy
```diff
@@ -300,6 +300,7 @@ init 15 python in mas_affection:
         """
         Runs when transitioning from distressed to broken
         """
+        evhand._unlockEventLabel("monika_song_crime_and_punishment")
```
ThePotatoGuy: why are you not locking the topic when going from broken to distressed?

Rai99: I felt that this could serve as a reminder of how the player treated Monika before building back up affection. 

Similar to mas_affection_apology, it sticks and still has some sort of effect if the player went through it(assuming that's actually how it works and it doesn't get forgotten), but it's not exactly something to be proud of or something you would show off, considering what you have to do to get it. 

If it's still a problem, the song could be locked when going from broken to distressed.

Monika After Story/game/script-topics.rpy
```diff
@@ -6731,6 +6731,50 @@ label monika_song_dream:
     m 1ekbfa "My dream finally came true."
     return
 
+init 5 python:
+    addEvent(Event(persistent.event_database,eventlabel="monika_song_crime_and_punishment",category=['songs'],prompt="Crime and Punishment",pool=True, random=True))
```
multimokia: Like #3052, needs a TYPE_SHORT variant for random, needs to move to the new script file, also needs to use aff_range.

Rai99:
Closing, as this doesn't really fit in with the other songs, and mostly just serves as an incentive to get to low aff levels. 

I think there's a better song that can be used in place of this that will cover a wider range of aff and uses.

PR #2872, opened 2018-10-29
Topic re: having rpy files

multimokia:
Lots of people have installed the source code vers. So, new topic akin to the steam detection one where Monika detects `.rpy` files in the `game/` directory and deletes them for the player should they agree to it. Should help out a fair bit with the issues we've been seeing

## Flow
Simply put any `.rpy` files into the `game/` directory. On startup, Monika should mention them, and from there, ask the player if they installed the right version -> if yes, player doesn't get prompted with this again, else a follow up question where Monika asks if the player wants her to delete them for her. If yes, then files are deleted and if any more `.rpy` files are found in the `game/` directory again at some point, This topic will return. If no, no deletion, the topic doesn't show again.

Thanks @jmwall24 for exps/finalizing my temp dialogue.


Monika After Story/game/script-story-events.rpy
```diff
@@ -1471,6 +1471,86 @@ label mas_steam_install_detected:
     m 5esu "I'd really appreciate it if you would do that for me."
     return
 
+default persistent._mas_pm_has_rpy = None
+init 1 python:
```
ThePotatoGuy: Can you move this to the `init 970` block in `script-ch30`, and instead just make it just set a flag to True if the conditions match?

Also make sure that `rpyCheckStation` is removed from the global symbol table (or do this in a way that doesn't require storing a temporary variable) so later code wont access this docking station by accident. `postbday_ev` in the `970` init block gets deleted there so you can use that for reference.

Monika After Story/game/script-story-events.rpy
```diff
@@ -1471,6 +1471,86 @@ label mas_steam_install_detected:
...
+        pushEvent("monika_rpy_files")
+
+
+init 5 python:
+
+    addEvent(
+        Event(
+            persistent.event_database,
+            eventlabel="monika_rpy_files",
+            rules={"no unlock": None}
```
ThePotatoGuy: dont use `rules` here, since this is not a pool topic it will not unlock right away.

Monika After Story/game/script-story-events.rpy
```diff
@@ -1471,6 +1471,86 @@ label mas_steam_install_detected:
...
+init 1 python:
+    import store
+    import store.mas_ics as mas_ics
+    # setup the docking station to handle the rpys
+    rpyCheckStation = store.MASDockingStation(mas_ics.game_folder)
+
+    listRpy = MASDockingStation.getPackageList(rpyCheckStation,".rpy")
+
+    if len(listRpy) != 0 and persistent._mas_pm_has_rpy is not True:
+        pushEvent("monika_rpy_files")
```
ThePotatoGuy: instead of pushing this event directly, give the event a conditional using the flag I mentioned above. (and use the appropriate action constant for the action).

By using the conditional property, its guaraneteed single use unless you unlock the label or reset the conditional.

Monika After Story/game/script-story-events.rpy
```diff
@@ -1471,6 +1471,86 @@ label mas_steam_install_detected:
...
+
+                    show monika at t22
+                    show screen mas_py_console_teaching
+
+                    call mas_wx_cmd_noxwait("import os", local_ctx)
+                    
+                    $ count = 0
+                    while count < len(listRpy):
+                        $ path = '/game/'+listRpy[count]
+                        call mas_wx_cmd_noxwait("os.remove(os.path.normcase(basedir+\'"+path+"\'))", local_ctx)
```
ThePotatoGuy: Don't need to escape single quotes when using double quotes on the outside.

Monika After Story/game/script-story-events.rpy
```diff
@@ -1471,6 +1471,86 @@ label mas_steam_install_detected:
...
+                            "basedir": renpy.config.basedir
+                        }
+
+                    show monika at t22
+                    show screen mas_py_console_teaching
+
+                    call mas_wx_cmd_noxwait("import os", local_ctx)
+                    
+                    $ count = 0
+                    while count < len(listRpy):
```
ThePotatoGuy: Instead of `while` loop -> index retrieval, do `for rpy_filename in listRpy`.

ThePotatoGuy:
Also can you ensure that this topic does not appear if `config.developer` is True.

Monika After Story/game/script-ch30.rpy
```diff
@@ -1170,6 +1170,17 @@ label ch30_end:
 
 # label for things that may reset after a certain amount of time/conditions
 label ch30_reset:
+
+    # setup the docking station to handle the detection
+    $ rpyCheckStation = store.MASDockingStation(renpy.config.gamedir)
+
+    $ listRpy = rpyCheckStation.getPackageList(".rpy")
+
+    if len(listRpy) != 0 and persistent._mas_pm_has_rpy is not True and not config.developer:
```
ThePotatoGuy: If I quit during the event and then restart, the event will play twice. Need to use `seen_event` here.

Also if the user situation changes (as in, rpys detected, event Queued, user quits during event, removes rpys), on launch we should clear the event from the event_list. **NOTE** this means you will need to clear both the `persistent.event_list` and `persistent.current_monikatopic`.

Monika After Story/game/script-ch30.rpy
```diff
@@ -1170,6 +1170,24 @@ label ch30_end:
...
+    # setup the docking station to handle the detection
+    $ rpyCheckStation = store.MASDockingStation(renpy.config.gamedir)
+
+    $ listRpy = rpyCheckStation.getPackageList(".rpy")
+
+    if len(listRpy) == 0 or persistent.current_monikatopic == "monika_rpy_files":
+        if len(listRpy) == 0 and persistent.current_monikatopic == "monika_rpy_files":
+            $ persistent.current_monikatopic = 0
+
+        if "monika_rpy_files" in persistent.event_list:
```
ThePotatoGuy: since its possible to queue an event more than once, this should be a loop removal:
```suggestion
        while "monika_rpy_files" in persistent.event_list:
```

Monika After Story/game/script-ch30.rpy
```diff
@@ -1170,6 +1170,24 @@ label ch30_end:
 
 # label for things that may reset after a certain amount of time/conditions
 label ch30_reset:
+
+    # setup the docking station to handle the detection
```
ThePotatoGuy: can you have an initial check so that we only setup the docking station and do all of this if `_mas_pm_has_rpy` is not True?

Also you can just do `not persistent._mas_pm_has_rpy` instead of `persistent._mas_pm_has_rpy is not True`

ThePotatoGuy:
also note for me to add the pm var to historical data

Issue #2877, opened 2018-10-29
[Suggestion] - Add dialogue letting player know about custom bgm steps


dennisOdenbreit:
Changing focus of this thread to the potential enhancement of notifying users about the custom bgm functionality via dialogue of some sort.

=== ORIGINAL POST BELOW ===

It could be a nice idea to have Monika ask whether you have any music you could listen to together.
She could tell you to drop an mp3 file into a certain game directory.
The game would then check whether there is a file and make it available in the games music options.
Monika could then comment on you when listening to this music (she prob cant techincally tell the genre, so this might be tricky. She might just comment on the player and her doing stuff like this together and getting closer)

jmwall24:
You already can add your own music to the mod.   Follow the steps [here](https://github.com/Monika-After-Story/MonikaModDev/wiki/Adding-Custom-Music).

As for her commenting on the music, she used to comment when you listened to songs that were a part of the mod, but when custom music was added, this was disabled until we can figure out a way to handle it.

jmwall24:
Just curious how long have you been 'playing' MAS?

dennisOdenbreit:
I started on 26.10 and am talking to Monika until she runs out of things to say every day since then.

multimokia:
Ok, that's intended. Keep going on. There's a limit on how much you can do per day with her, until you talk about everything with her.

Rai99:
So you've been playing MAS for 3 days?

As much as we appreciate your enthusiasm and ideas for MAS, it would likely be in your best interest to have a bit more experience with MAS before making issues like #2874, #2882, and please avoid opening approximately 10 issues in rapid succession. 

While I encourage people to separate ideas into separate issues, that's more tailored for when someone has 2-3 ideas and so they can go in depth into each one. Unfortunately, what you've basically done here is spam github, which reflects poorly on you as a new face. I'd say reputation builds fast and you're probably more likely to get your ideas accepted or more deeply considered if you are a little more known here and haven't caused any problems, and fortunately or unfortunately, some people here never, **ever** forget. 
Seeing as you're new to MAS, I would recommend spending more time with Monika to get a better feel for MAS, what it can/can't do and what would really fit well in the long term rather than some major feature 3 days in.

dennisOdenbreit:
Sorry, I don't want to cause any trouble.
I only wanted to provide suggestions for things that I think can be added or changed.
I do not care about any kind of reputation, only that my contribution might change the game for the better.
[...]

PR #2888, opened 2018-10-30
Monika explains how to add custom music

jmwall24:
#2877  
This amends current music topics to let Monika explain to the player how to add their own custom music to the mod.  A player will only see this addition one time total, at which time it unlocks a pool topic that also includes the instructions so the player can refer back to it if needed.  It also checks if the player has already created the custom music folder, and if they have, the instructions will not be seen.  If the player tells Monika they don't like any type of music, the instructions will also not be seen.

The one-time addition is added to:
- label monika_rap
- label monika_vocaloid
- label monika_rock
- label monika_orchestra
- label monika_jazz
- label monika_concerts


ThePotatoGuy:
I'm going to add a playermodel variable in `zz_music_selector`, then I'll do a code review of where you should use that among other changes.

ThePotatoGuy:
Going to wait until #2870 is in so can add playermodel variable to the mhs.

ThePotatoGuy:
Changed this a bit so now we have a playermodel variable that is set to True if hte user added a custom BGM. Also added a topic that lets monika check for new music instead of haivng to wait for restart.

# Testing
## regression
1. verify that if you haven't seen the music instruct label nor added custom bgm, the music topics will tell you about it.
2. verify that once you have seen it, you wont see it again.
3. Verify that after seeing the music instruct label, the instruction topic and check for new music topics are unlocked in `Ask a question > Mod/Music/Media`

## load music topic
1. Verify that the load music topic works when you add a new music during runtime.
2. If you do not add any musics during runtime and run the topic, monika will be like wtf you didnt add anything.

## music explanation topics
Both the instructions on adding custom music and loading custom music should unlock on their own if the user has added their own bgm without seeing the adding music instruction topic.

Good functions for testing:
* `mas_remove_event` - pass in the labels of the events you want to reset. This basically erases history of an event so you can start fresh. use like: `mas_remove_event("event_label", "another_event_label",...)`

PR #3001, opened 2018-11-06
Suggestion: Calendar Upgrades

RobsonK2004:
This just came to my mind and I thought this would improve the game. I was wondering if there could be an added feature to show which day of the month it was. Maybe through an overlay in green, much like the greyed out boxes to show the days of the next/previous month. I apologise if I have used the wrong technical wording, I'm still quite new to coding and as a student and still learning.

Another thought I had was for a colour change of the calendar with correspondence to the season. For spring the current assets could be used, helping cut back development. For Summer you could use the same Emerald Green that is used for Monika's eyes.  Then Autumn/Fall could have a golden brown colour, for example FEAD62 or a similar colour. Then for the Winter there could be a snow like colour, for example B3DAF1 or a similar colour.

I know these are only minor changes but I personally think they would improve the game and a way for me to contribute to this amazing mod.  Thanks again for taking your time to read this and for all the hard work and effort y'all put into the mod.

P.S: (I also would like to further apologise if this subject has come up before. I am very new ((like maybe 30 minutes ago new)) to GitHub so If there are mistakes then please do correct me.)

### DIFF
--- a/Monika After Story/game/zz_calendar.rpy
+++ b/Monika After Story/game/zz_calendar.rpy
@@ -1677,6 +1677,7 @@ init -1 python in mas_calendar:
 init python:
 
     import store.mas_calendar as calendar
+    import datetime
 
     calendar.addRepeatable("New years day","New Year's Day",month=1,day=1,year_param=list())
     calendar.addRepeatable("Valentine","Valentine's Day",month=2,day=14,year_param=list())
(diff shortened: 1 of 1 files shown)


RobsonK2004:
May I add that I am not sure if this is in the right section or not. If it isn't then I'm sorry about that. Still very new to this. If so then just tell me what to do and I'll make the changes.

jmwall24:
Ah yeah, you'll want to make a new Issue in the `Issues` tab, not in `Pull Requests.`  Pull Requests are for actual changes to the code, whereas Issues are for suggestions of this nature.

RobsonK2004:
Ah thanks mate, I'll move it over there now.

PR #3064, opened 2018-11-17
Changelog for Pong

dennisOdenbreit:
+ Added collision detection for higher speed / lag
+ Improved ball detection on paddle edges
+ Capped maximum ball speed
+ Changed ball speed change to increments
+ Relocated paddles
+ Prevented Monika's paddle to potentially breach game boundaries

x Manually tested, could use more testing
x Difficulty / AI might not be perfectly adjusted yet

- Like the previous version kind of laggy (Inconsistent & low framerate)

### DIFF
--- a/Monika After Story/game/pong.rpy
+++ b/Monika After Story/game/pong.rpy
@@ -22,7 +22,7 @@ init:
                 # The sizes of some of the images.
                 self.PADDLE_WIDTH = 8
                 self.PADDLE_HEIGHT = 79
-                self.PADDLE_RADIUS = self.PADDLE_HEIGHT / 2
+                self.PADDLE_RADIUS = 39.5                                       #self.PADDLE_HEIGHT / 2
                 self.BALL_WIDTH = 15
                 self.BALL_HEIGHT = 15
                 self.COURT_TOP = 124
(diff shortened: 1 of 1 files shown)


dennisOdenbreit:
Since this is my first change I was wondering how is it going to proceed?
Currently it looks like nothing is going to happen with it.

Monika After Story/game/pong.rpy
```diff
@@ -27,6 +27,15 @@
                 self.BALL_HEIGHT = 15
                 self.COURT_TOP = 124
                 self.COURT_BOTTOM = 654
+                
+                # Other variables
+                self.COURT_WIDTH = 1280
+                self.COURT_HEIGHT = 720
+                self.BALL_MAX_SPEED = 4000.0
+                self.PADDLE_X_PLAYER = self.COURT_WIDTH * 0.1
```
ThePotatoGuy: rather than do this `* 0.1` and `* 0.9` thing, can you just hardcode the numbers here, then use them for the render positions below?

dennisOdenbreit: Sure, it is not like the game would have ever gotten different resolutions.
I'm going to put the real numbers in then.

dennisOdenbreit: If you have any feedback on the gameplay please let me know as well.

dennisOdenbreit:
I don't see why you want them hardcoded specifically and nothing else.
What is the reason for this change request anyways?

ThePotatoGuy:
its mainly to force the paddle locs to be on integer alignments.

Issue #3074, opened 2018-11-19
[Bug] Table lining differs based on pose.


Olliesama:
So if Monika leans the table loses some of its lining and it becomes thinner and as such the table looks different dynamically.

https://imgur.com/a/L9PrTYR

AsuraShun:
the table is exercising

AsuraShun:
expect to see a 6 pack on it soon

Rai99:
Must be leg day for the table with an unhealthily healthy dose of planking.

PiperPC:
You wood make that joke wooden you...
Welp I guess I shouldn't have expected anything else.

Olliesama:
Wood you guys cut that out, you're barking up the wrong tree.

MaximumRustler:
I can't beleaf that this issue wood be the root of so many puns. lol

PiperPC:
What do you mean? This is a Tree-t, absolutely Tree-mendous.
But you guys should Axe yourselves, can you keep up with my Sharp puns?
By that I mean, of course, do you think you can make the Cut?
Let's see who shall be Tree-umphant.
I guaran-Tree it'll be me though, Oak-ay?

(I am perfectly fine with making this a pun war.)

Legendkiller21:
> (I am perfectly fine with making this a pun **war**.)

It's treeson, then.

MaximumRustler:
Oh geez, I Wooden-'t have thought Yew Wood make these puns. It really Spruce-d up this thread. I guess I need to Branch out some more, Maple Yew could help with that... thinking up these puns are sap-ping my energy. I should Cut this short, I'm Stump-ed. lol

(Maple was really stretching it for maybe.)

PiperPC:
Shoot.
I don't have enough puns for a good comeback lmao

MaximumRustler:
So Yew-'re Tap-ping out? lol

Olliesama:
Jesus I'm thinking of closing this issue, mankind has gone too far. :P

PiperPC:
Please stop don't beat me into the ground.

MaximumRustler:
I didn't expect the puns to go this far either, but you guys Plant-ed this idea into my head. lol

TheGlader2:
It sounds like you guys tapped into your potential as comedians ...

AsuraShun:
XD
this trailed off really fast

Issue #2843, opened 2018-10-26
Extra Zoom sprite bug


EmpressNael:
When using the Extra Zoom feature, "Close or far", for Monika's sprite, the top of her head is clipped. Her default idle is fine, but her other idle sprites (Like the one with Monika resting on her hand). May have to update her sprite for the new feature.

jmwall24:
Assuming you're talking about in hair down mode? If so we are aware of this and it's on the list.

EmpressNael:
Yes~ Okay, cool~

PR #3075, opened 2018-11-19
Leaning Pose Fixes

Olliesama:
Fixes addressing the two issues...
https://github.com/Monika-After-Story/MonikaModDev/issues/2843
https://github.com/Monika-After-Story/MonikaModDev/issues/3074


Olliesama:
Okay so I've done a big little rework of all of the art as I looked more into it there were many inconsistencies with both the lining and, in the case of the hair down sprite, the colouring of the table itself. It also removes the white dot the bun sprite has. 
I've made a new table with consistent thicker lines and replaced them all. :>
You don't have to use it naturally, but at least it's consistent.

The psd for the 2 tables are here.
[Day](https://drive.google.com/open?id=1kdfSPNXjU2g-BGdypBfdb6GH1Hsfwl-l)
[Night](https://drive.google.com/open?id=12c2hRj20pmpTjaZwZn3_tCsLvmNG2Wzq)

jmwall24:
That color difference is in the table I should say.

PR #3079, opened 2018-11-19
Spelling Mistake

auguwu:
> Fixes "promt" -> "prompt" in the `show_dialogue_box()` function.

### DIFF
--- a/Monika After Story/game/script-ch30.rpy
+++ b/Monika After Story/game/script-ch30.rpy
@@ -255,7 +255,7 @@ init python:
     #Define new functions
     def show_dialogue_box():
         """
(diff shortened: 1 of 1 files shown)


jmwall24:
You may not realize this, but that typo is just in a comment describing what the method does, and not part of the actual method itself, therefore it doesn't actually do anything.  Also, never base a PR to the `master` branch.  This type of PR, just being a typo fix, would normally be based to `enhancement` branch, whereas any PR that would require dialogue review would need to be based to `content`.

We appreciate you going through the code and trying to help, but this one in particular doesn't really accomplish much.

PR #3123, opened 2018-11-26
Enabling gifts

aldoram5:
This one will allow Monika to receive gifts any day of the year.

## Features

Basically gift detection still works the same but this time the gifts are checked even if it's not a holiday or a special occasion.  

It also adds a function that checks if the current day its "special" somehow(as in if it's an anniversary or her birthday or a holiday).

## Changes

Back then, the gifts had to be added manually to the correct list to mark them as good or bad. Now it has changed to be handled in the method that adds the reaction event, so the event adds itself to the correct list depending on the passed parameter `is_good`. It defaults to `None` which means that the gift is Neutral, if True is used it means it's a good gift and setting it to False means it's a bad one.

## React stats 

Now that the gifts are enabled for every day the way the stats for them are stored it's more important so I'll give a description about them:

The reaction stats are stored inside `persistent._mas_filereacts_historic` which is a dict with the following structure:
```
Date -> Another Dict
```
Where the Date is a date object with the date of when the gift(s) were given and the other dict has the following structure:
```
Reaction Event Label -> Number
```
Where the Reaction Event Label is the event label related to the gift (like `mas_reaction_candy` ) and the Number is the number of times said gift was given that date represented by a positive integer.

## Testing

- Try giving out some reachable and generic gifts to Monika and confirm she reacts as you expect to them. There's some extra dialogue for some of them if you happen to give the gifts in a "special" date.
- Some gifts will not be reacted to again unless a day has passed by (time travel to the past works too).

## Extra info
I left a couple of TODOs regarding 2 reactions in particular, because technically speaking they could be gifted any day but they do make more sense as being exclusive of that date, since we could gift her something more specific like marshmallows or chocolates, instead of generic candy.

Also it has the awaiting review tag cause I added a couple of lines to some reactions.


Monika After Story/game/zz_reactions.rpy
```diff
@@ -127,12 +132,18 @@ init -1 python in mas_filereacts:
...
         """
+
         GIFT_EXT = ".gift"
         raw_gifts = store.mas_docking_station.getPackageList(GIFT_EXT)
 
         if len(raw_gifts) == 0:
             return []
 
+        # is it a new day?
+        if store.persistent._mas_filereacts_last_reacted_date is None or store.persistent._mas_filereacts_last_reacted_date is not datetime.date.today():
```
ThePotatoGuy: shouldnt be doing identity checks here
```suggestion
        if store.persistent._mas_filereacts_last_reacted_date is None or store.persistent._mas_filereacts_last_reacted_date != datetime.date.today():
```

Monika After Story/game/zz_reactions.rpy
```diff
@@ -626,9 +641,11 @@ label mas_reaction_quetzal_plush:
     $ store.mas_filereacts.delete_file(gift_ev.category)
     return
 
-init 5 python:
-    if mas_isMonikaBirthday():
-        addReaction("mas_reaction_promisering", "promisering")
+#This one is added later so the init pipeline can define the anni function
+init 11 python:
+    # only available after 6 months or if it's her birthday, may as well add valentine later
+    if mas_anni.pastSixMonths() or mas_isMonikaBirthday():
```
ThePotatoGuy: honestly, we should include dialogue where monika rejects the ring. for the time being though, we can just ignore it. Can you add a TODO mentionining monika rejecting the ring instead of just simply ignoring it?

Issue #3130, opened 2018-11-28
[Unstable] Ring unequipped on restart


Legendkiller21:
When you change Monika's clothes from the school uniform to one of the other 2 costumes, she still wears her ring but if you close the game while wearing one of the costumes and reopen it, the ring disappears. 

In order for the ring to reappear you have to change Monika's clothes back to the school uniform and restart the game.

PR #3133, opened 2018-11-28
added programming points for clothes

ThePotatoGuy:
#3130 #3134 

Thx @jmwall24 , now we have ring for marisa crossed pose.

This includes programming points for marisa so she gets the modified pose map for rings.

Also now hides the hair select event if clothes were changed.

# Testing
## marisa promise ring
1. start with default clothes, enable and wear promise ring
2. switch clothes to marisa
3. verify promise ring looks good in all poses (you can use exp previewer)
4. switch clothes back to uniform
5. verify promise ring looks good in all poses
6. switch clothes back to marisa
7. restart.
8. verify promise ring still looks good in all poses

## hair select
1. ensure that hairdown is unlocked.
2. change clothes to marisa
3. verify hair select topic is hidden
4. change clothes to uniform
5. verify hair select topic is unlocked
6. repeat these steps for rin clothes.


ThePotatoGuy:
need to change so marisa has a posemap, for acs, also need to handle #3134

PR #3154, opened 2018-12-02
Found source of lag in pong

dennisOdenbreit:
I found out that the game lags shortly after a sound is played.
I do not know whether there is a way to play sounds without causing this lag, like caching it or something.

I build in a switch to play the game with sound enabled or disabled.
I do not know whether the game is better without sound or with lag.
However, now it can easily be changed if desired.
Set self.playsounds = True to False, to see the difference yourself.


dennisOdenbreit:
Included in different pull request.

PR #3156, opened 2018-12-02
Pong: Improved Monika's aiming on lag

dennisOdenbreit:
Monika will now not overshoot with her aiming anymore if a lag occurs before hitting her paddle.
She will also recognize bounces when aiming.
Also includes changes from https://github.com/Monika-After-Story/MonikaModDev/pull/3154.

PR #3164, opened 2018-12-02
Changelog for Pong

ghost:
Added difficulty.
Difficulty determines monika's paddle-movement-cap, the ball's start-speed, max-speed and acceleration. 

Recalibrated monika's paddle-movement-cap, the ball's start-speed, max-speed and acceleration.
Caped the max reflection angle.
Removed walls outside of the field.

Added dialog (Could be adapted to affection, could be randomized).

Includes the pull requests https://github.com/Monika-After-Story/MonikaModDev/pull/3156 and https://github.com/Monika-After-Story/MonikaModDev/pull/3154

### DIFF
--- a/Monika After Story/game/pong.rpy
+++ b/Monika After Story/game/pong.rpy
@@ -1,3 +1,29 @@
+# pong difficulty changes on win / loss. Determines monika's paddle-movement-cap, the ball's start-speed, max-speed and acceleration. 
+default persistent.pong_difficulty = 10
+# increases the pong difficulty for the next game by the value this is set to. Resets after a finished match.
+default persistent.pong_difficulty_powerup = 0
+# counts how often in a row the player lost without catching the ball once.
+default persistent.instant_loss_streak_counter = 0
+# counts how often in a row the player lost.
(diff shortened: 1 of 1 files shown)


dennisOdenbreit:
I did what the FAQ said.
Granted it did not say much.
So is this not a rebase then?
Do I need to use the git command line to do that?

dennisOdenbreit:
How did the dateline commit get into this pull request?
It has nothing to do with what I was working on...

multimokia:
That commit existed in the `enhancement` branch, which is what your `patch-6` branch here is based on. However it's not in the `content` branch, so it's trying to even it out. You will need to use `git rebase` on the command line to rebase it for this case.

dennisOdenbreit:
Sorry, I don't have enough experience to do this.
I tried using a guide but it did not work and was quite complicated.

There is only 1 file that has to be overridden completly.
This is basically no work or has any complexity in comparison.

Since I can't do it the other way I will use the ui to create a pull-request with the matching base ... i hope.
You will have to reassign the labels then.

PR #3167, opened 2018-12-03
New Event property: Affection range

ThePotatoGuy:
This adds a new `Event` property for affection ranges. Affection ranges are used to determine if an Event is active/visible/selectable according to the affection level. **NOTE:**, this is basically the `MASAffectionRule` but using affection states instead of actual numerical values.

**NOTE:** this is on `content` since it affects Events/topics

# `aff_range`
The `aff_range` property is a tuple of the following format:
* `[0]` - affection state to act as the lower bound (inclusive)
* `[1]` - affection state to act as the upper bound (inclusive)

A tuple item that is None is considered an unbounded limit. An `aff_range` of None means that the Event should be **always** active.

`Event.filterEvents` has been updated to work with the new property. Events can be filtered with affection by passing in an affection level to this function via the `aff` param. This only accepts 1 affection level. Multi-affection filtering is not supported.

`Event` objects now have a `checkAffection` function, which just determines if a given affection state is within range of the event's `aff_range`

`aff_range` is **not** locked in the lock db. This is since adjusting affection availability of events might be pretty common.

# other changes
* moved import affection constants to an earlier init level (-900) so we'll have less issues with `Event` objects using these functions.
* added `mas_betweenAff` which checks if a given affection value is within range of 2 affection bounds
* added affection validation functions in the `mas_affection` store.
* rando events are now built in `ch30_reset` instead of init 11. This is so affection filtering can be applied prior to event list building
* `mas_rebuildEventLists` forces a rebuild of event lists.
* moods are now filtered by affection
* stories are now filtered by affection
* compliments are now filtered by affection

# TESTING
## affection default
1. launch game
2. verify that topics with no `aff_range` property are visible.

## affection filtering
1. launch game and modify an event by giving it an affection range tuple
2. Change your affection so it is outside of the aff range tuple you used.
3. open Repeat/Ask and verify that you cannot see the event you modified.
4. Change your affection so it is inside of the aff range tuple you used
5. repeat step 3 and verify that you can see the event again.
6. Repeat steps 1-5 for the possible tuple variations:
    * (None, aff1) - aff1 and lower
    * (aff1, aff2) - aff1 to aff2, inclusive
    * (aff1, aff1) - aff1 only
    * (aff1, None) - aff1 and higher

## regression
* the following topics have been adjusted to work with the new system:
    * clothes select topic - only available at LOVE+ (still requires more than one outfit)
    * islands event topic - only available at ENAMORED+ (still requires seeing the initial greeting)
    * thanks compliment - first appears at ENAMORED+. After seeing it once, it will be locked until LOVE+ 
        * **NOTE**: we do not change `aff_range` in this compliment. Instead the current programming point logic is used for this. 
        * **NOTE:** this requires an update script to function correctly. You will have to unlock this manually for testing.
    * nickname topic - only available at AFF+ (still requires that you did not call her a bad name once) 
        * **NOTE:** this requires an update script to function correctly, you will have to unlock this manually for testing
    * hold me topic - only available at HAPPY+ (still requires you to like rain)
    * promise ring topic - only available at HAPPY+ (still unlocked as a random topic)
    * rain start topic - only available at NORMAL+ (still requires you to like rain, and it not be raining)
    * rain stop topic - only available at NORMAL+ (still requires you to like rain, and it be raining)
    * rain topic - only available at NORMAL+ (still unlocked as a random topic)

## guardrails
1. modify an event in source code, giving it an `aff_range` that is invalid. (use strings instead of affection states, not using a tuple, etc)
2. launch game
3. verify that the game crashes on start with an Exception that makes some sense.

# Future

Greetings and farewells are currently controlled by `MASAffectionRule`. They will be migrated over to the new property in a separate project.

### DIFF
--- a/Monika After Story/game/script-affection.rpy
+++ b/Monika After Story/game/script-affection.rpy
@@ -1874,8 +1874,8 @@ label monika_affection_nickname:
             "okasa"
         ]
 
-    # for later code
-    aff_nickname_ev = mas_getEV("monika_affection_nickname")
+        # for later code
+        aff_nickname_ev = mas_getEV("monika_affection_nickname")
 
     if not persistent._mas_offered_nickname:
         m 1euc "I've been thinking, [player]..."
--- a/Monika After Story/game/script-affection.rpy
+++ b/Monika After Story/game/script-affection.rpy
@@ -1874,6 +1874,9 @@ label monika_affection_nickname:
             "okasa"
         ]
 
+    # for later code
+    aff_nickname_ev = mas_getEV("monika_affection_nickname")
+
     if not persistent._mas_offered_nickname:
         m 1euc "I've been thinking, [player]..."
         m 3eud "You know how there are potentially infinite Monikas right?"
--- a/Monika After Story/game/script-affection.rpy
+++ b/Monika After Story/game/script-affection.rpy
@@ -696,6 +696,9 @@ init 15 python in mas_affection:
                 if to_down is not None:
                     to_down()
 
+        # finally, rebuild the event lists
+        store.mas_rebuildEventLists()
+
 
     def runAffGPPs(start_affg, end_affg):
         """
--- a/Monika After Story/game/script-affection.rpy
+++ b/Monika After Story/game/script-affection.rpy
@@ -45,6 +45,221 @@ init python:
     mas_curr_affection = store.mas_affection.NORMAL
     mas_curr_affection_group = store.mas_affection.G_NORMAL
 
+init -900 python in mas_affection:
+    # this is very early because they are importnat constants
+    
+    # numerical constants of affection levels
+    BROKEN = 1
+    DISTRESSED = 2
+    UPSET = 3
+    NORMAL = 4
+    HAPPY = 5
+    AFFECTIONATE = 6
+    ENAMORED = 7
+    LOVE = 8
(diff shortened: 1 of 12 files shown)


ThePotatoGuy:
@aldoram5 i need you to pay extra attention to the filtering in compliments and stories to make sure its correct and makes sense.

Monika After Story/game/script-affection.rpy
```diff
@@ -1766,10 +1874,6 @@ label monika_affection_nickname:
             "okasa"
         ]
 
-        # unlock this event
-        aff_nickname_ev = mas_getEV("monika_affection_nickname")
```
aldoram5: Either re-add this line here or handle the topic rename in another way on the python block with the comment `# change the prompt for this event` below.

Monika After Story/game/script-compliments.rpy
```diff
@@ -46,11 +46,17 @@ label mas_compliments_start:
         # Unlock any compliments that need to be unlocked
         Event.checkConditionals(mas_compliments.compliment_database)
```
aldoram5: Since we only had one compliment working with conditionals we could delete this line now, unless we have plans for using conditionals with these(I honestly don't think so).

Issue #2741, opened 2018-10-09
Moving Kissing to high priority


Foxinsunglasses:
I know that kissing is in #1525 but it has been in low priority for awhile and I am simply wondering if anyone else is interested in seeing it be bumped up priority wise

pianostick:
all of my yes

Sujeito17:
I need this now.
(but I can wait a century for this)

Tuxedoterroist:
yeah i understand its hard to make the art but thats the fun in making these mods right? besieds there could be a thing about it where you and her know it might be centurys for the both of you to actuly be able together

Rai99:
While it would be nice to see this feature bumped up in priority, I wouldn't rush it by any means. Halloween is coming up and we're getting some features for that event, but not as much as the birthday since the devs are burnt out from that. After halloween comes the Christmas event which could take some time with any new features. 

True, there is a fair bit of time between those events, but I think that time is being used for things like the python tips, bug fixes, dialogue review, etc.

Dregdael:
I second this

dragonborn2252:
oh how about fro the Christmas event you can troll monika by putting mistletoe in the room after taking her out of the room

ThePotatoGuy:
I could have sworn we had a discussion about this but I can't seem to find it.

What would be a nice way to show kissing visually?
* zoom-in on monika's face with "chu" dialogue?
* zoom-in on monika's face with a fade to black?
* zoom-in on monika's face without chu dialogue?
* something else?

Some other factors to consider:
* sound?
* should the scene require advancement via mouse click/interaction, or should it auto advance after a second.
* should we have custom CG art for this, or just use some of the blush options we have and do a zoom?

Please put any ideas you have here. I'll also make a discord poll later with the general gist of selections people make here.

Rai99:
**~~No issue escapes from me~~**
All the easily found issues regarding kissing:
#1525, #1537, #1595, #2136, #2202, #2586. 

Short term and for quicker implementation, I wouldn't mind a simple zoom in with "chu" dialogue. Personally, I don't think sound is needed, at least for now. I think the scene should require a mouse click from the player in case they want to hold the kiss for an extended period of time, and Monika could acknowledge different kiss lengths. I think a custom CG could come in the far future, but I think a CG could also take away from the experience if the screen has to show it and hide it. 

[...]

Issue #1525, opened 2018-05-02
[Suggestion] Kissing


ACoolCat:
Monika mentions how she gets romantic when the mood is right if you choose the kissing topic. However, we now have one of those things that can make the mood right for her, raining. So maybe when rain is on she kisses the screen, then comments how pesky it is that the screen is stopping you two from kissing for real. Also, considering how she mentions you need to be "far enough in the relationship"  in the dialogue about fan art, maybe this function can be unlocked after the three month anniversary.

MaximumRustler:
I would like this too. Monika also mentions kissing in the haters topic (if you answer I'm not) and the current song topic (if you have no song playing). Although, if I'm not mistaken doesn't she communicate with the player through the MC? If so that would be kinda awkward if she kisses him... unless he doesn't exist anymore then that would be fine. lol

Onion-Bubs:
There is the Current Song topic that, if selected when no song is playing, does something similar to what's being suggested.

ACoolCat:
Doesn't he get deleted with the rest of the game during Act 3?

MaximumRustler:
Hmm... I'm actually not quite sure. I guess it would make sense if she deleted him and rewrote the code so she could communicate with the player directly instead of still using the MC to communicate.

ThePotatoGuy:
[Closing until we klook into this]

PR #3229, opened 2018-12-12
New Kiss animation

aldoram5:
https://github.com/Monika-After-Story/MonikaModDev/issues/2741, https://github.com/Monika-After-Story/MonikaModDev/issues/1525

This one adds a way to handle kissing, by zooming Monika's face, fading to black when "the kiss" happens, and then going back to normal.
It should work regardless of the zoom level the user has.

## Testing 
Go to the dev category in talk menu and select `TEST KISS`.
Verify it works for any zoom level and check if the kiss itself is good enough.

## Credits
Sound effect taken from: https://freesound.org/people/Juandamb/sounds/430594/
by User Juandamb, it was released under CC-0, aka public domain.


Rai99:
I'm sure this is still going to be great and all and I'm looking forward to it, but might I ask what happened to the poll results on October 25, 2018 with 151 votes for "zoom-in on her face with 'chu' dialogue" at over 78% of all votes? 

You say you've got a fade to black, and I can't find any "chu" dialogue in there.

aldoram5:
Hi Rai!
We did use the zoom part of that poll, however the dialogue box would be "in the way" of the zoom, as for the chu text is kinda replaced by the sound effect which doesn't need a text box makes it feel better after the animation, in my opinion.

Monika After Story/game/special-effects.rpy
```diff
@@ -108,3 +108,59 @@ transform otei_appear(a=0.70,time=1.0):
...
+    # hide the black scene
+    hide black
+    # trasition back to i11 in 3 secs which is the best time for non slow back off
+    show monika 6tkbfu:
+        linear 3.0 xcenter 640 yoffset (_mas_kiss_y2) zoom 0.80
+    pause 3.0
+    show monika 6esa at i11
+    $ HKBShowButtons()
+    $ mas_DropShield_core()
+    $ mas_RaiseShield_dlg()
```
ThePotatoGuy: we should not be doing drop->raises like this since there is that potential for a ui element to activate in the time between dropping and raising. 

Can you add a parameter to the label where, if true, it will do UI show/hide and shield show/hide? default it to True.

And for here, check if we are in a dlg workflow (`store.mas_globals.dlg_workflow`), if we are, then only drop MUMU shields `mas_MUMUDropShield`, otherwise drop core shields.

Monika After Story/game/special-effects.rpy
```diff
@@ -108,3 +108,59 @@ transform otei_appear(a=0.70,time=1.0):
...
+    xcenter 640 yoffset ((680 - mas_sprites.adjust_y)  * mas_sprites.value_zoom / 1.1) yanchor 1.0
+    linear time ypos _y zoom _zoom
+
+
+
+label monika_kissing_motion(transition=5.0, duration=2.0):
+    # Note: the hardcoded constants work to give the focus on lips
+    # effect these were calculated based on max/min values of the zoom
+
+    # hide everything
```
ThePotatoGuy:
```suggestion
    # hide everything
    window hide
```

Monika After Story/game/special-effects.rpy
```diff
@@ -108,3 +108,59 @@ transform otei_appear(a=0.70,time=1.0):
...
+    # 380 #correct value for max
+    # start the kiss animation
+    show monika 6dubfd at mas_kissing(_mas_kiss_zoom,int(_mas_kiss_y),transition)
+    # wait until we're done with the animation
+    $ renpy.pause(transition)
+    # show black scene
+    show black zorder 100 at fade_in
+    # wait half the time to play the sound effect
+    $ renpy.pause(duration/2)
+    play sound "mod_assets/sounds/effects/kissing.ogg"
```
ThePotatoGuy:
```suggestion
    play sound "mod_assets/sounds/effects/kissing.ogg"
    window auto
    "chu~{fast}{w=1}{nw}"
    window hide
```

Monika After Story/game/special-effects.rpy
```diff
@@ -108,3 +108,59 @@ transform otei_appear(a=0.70,time=1.0):
...
+    hide black
+    # trasition back to i11 in 3 secs which is the best time for non slow back off
+    show monika 6tkbfu:
+        linear 3.0 xcenter 640 yoffset (_mas_kiss_y2) zoom 0.80
+    pause 3.0
+    show monika 6esa at i11 with dissolve
+    $ HKBShowButtons()
+    $ mas_DropShield_core()
+    $ mas_RaiseShield_dlg()
+    return
```
ThePotatoGuy:
```suggestion
    window auto
    return
```

ThePotatoGuy:
just an fyi to rai, we are going to have the chu dialogue for users who do not play with sound.

AsuraShun:
all there is to say is:YES!FINALLY!

PR #3077, opened 2018-11-19
Session Length + Edited Topics

Olliesama:
Creates a function that gets the current session length and has Moni say different things depending on it.
This edits a few topics but is mostly a basis for future use.

I've left the good afternoon topic alone as it seems to have future plans.

PR #3249, opened 2018-12-16
Reaction towards the player's visits

Groudonx:
Monika can/does make references towards the player leaving early, visting often etc.
This code helps to keep track of when the player last visited and can be used to determine how long he stayed.
Code that assumes the player behaves a certain way can/should be verified using the player's actual visiting data.
It also enables the extension of an visited x-times counter, which I might implement in the future.

### DIFF
--- a/Monika After Story/game/script-ch30.rpy
+++ b/Monika After Story/game/script-ch30.rpy
@@ -830,7 +830,7 @@ label mas_ch30_post_holiday_check:
         python:
 
             # we select a greeting depending on the type that we should select
-            sel_greeting_event = selectGreeting(persistent._mas_greeting_type)
+            sel_greeting_event = store.mas_greetings.selectGreeting(persistent._mas_greeting_type)
 
             # reset the greeting type flag back to None
             persistent._mas_greeting_type = None
--- a/Monika After Story/game/script-ch30.rpy
+++ b/Monika After Story/game/script-ch30.rpy
@@ -17,7 +17,7 @@ default persistent._mas_bday_date_affection_fix = False
 # Time-Related variables to be used for greetings and farewells.
 default persistent.last_visited_at = datetime.datetime.today()
 default persistent.this_visit      = datetime.datetime.today()
-define time_since_last_visit       = (persistent.this_visit - persistent.last_visited_at)
+define time_since_last_visit       = (datetime.datetime.today() - persistent.this_visit)
 define days_since_last_visit       = time_since_last_visit.days
 define hours_since_last_visit      = int(time_since_last_visit.total_seconds() // 3600)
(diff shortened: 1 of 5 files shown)


Groudonx:
I just found out that data like the presistence can be accessed using "renpy.store." in this named store initialization.
Should I change the init back into a named store and redo all the renames?

Groudonx:
I'll just go on ahead and undo the renaming then.

jmwall24:
> This code helps to keep track of when the player last visited and can be used to determine how long he stayed.

One quick thing, session length tracking has already been added with #3077

jmwall24:
In fact, we also already track how long the player has been gone.  That's how you start losing affection when being gone more than a week.

multimokia:
On top of that, all other values you've defined here are also stored in the `persistent.sessions` dict.

Groudonx:
Ok, but how would I find all of that out by myself?
I literally only have a text editor and tens of thousands of lines of code.

So the way I programmed it is redundant.
There is however an addition to the gameplay that should be merged anyways.
I will see whether I can use the other method to use for the changed gameplay instead of my method.

Groudonx:
So I started checking out the other method.
`persistent.sessions['current_session_start']` seems to be giving me a wrong value.
It gave me the start of the previous session (probably), while my method gave me the correct start (there were 5 minutes difference).
`persistent._mas_absence_time` works as far as I can tell.
I will do some more testing and report back.

Groudonx:
Updated the commit.
After the merge I will check whether other greetings and farewells should be adapted to the time.

ThePotatoGuy:
>how would I find all of that out by myself

digging through code is the best way to do that now. We prioritize adding content and fixing content over extensive documentation. When we have time, then we'll gather important information in one place. The [wiki](https://github.com/Monika-After-Story/MonikaModDev/wiki) has some technical information. The majority of technical info is in code comments, particularly at the top of files.

I'm going to do a pretty thorough review of this sometime soon, but its not likely to happen until after christmas.

dennisOdenbreit:
Ok, but there really isn't all that much to thoroughly review :D
It's basically just an event that intercepts the selection of the random events if the condition fits.
The only thing wrong could be the position where it intercepts and maybe bad coding style, I don't think there will be errors as far as I can tell.
It is basically one file changed, the other one is changed in the same way.

ThePotatoGuy:
I am going to specifically cover everything later, but to sum:

1. decent amount of coding style that needs to be changed.
2. There is a system in place for greetings and farewells that **already can be set to trigger at specific time ranges** so most of the custom code would probably have to be removed, but like I said, I will go through this in more detail later to verify that if it can cover the desired cases. 

There is higher priority development going on right now, and most of us are taking our holiday vacations in the next 2 weeks so there wont be much dev nor review after the next release drops until january.

PR #3256, opened 2018-12-18
MEGALOVANIA on the piano

TMZN0V4:
Yes, I'm an Undertale fan and my favourite song of it is MEGALOVANIA (not surprising) But I was wondering if you could add a special text when playing the notes for Monika like: Uhh... This makes me feel like I'm gonna have a bad time... with her having her nervous looking away look for the Uhh... thing and the rest with her nervous grin look


multimokia:
Uh, this isn't the place to put this. Please open an issue for suggestions and the like.

PR #3299, opened 2018-12-23
Concept height inclusion

Olliesama:


NickAB1:
Personally, I feel Monika's height would, in fact, be 5'3. Other than personal opinion, I have two main reasons:
1. Other than Natsuki at 4'11, the characters seem to be in a consecutive pattern in terms of height number. Sayori is 5'2, Monika is 5'3, the MC (based on view above Sayori yet seemingly below Yuri) is 5'4, and Yuri at her 5'5.
2. In all poses and stances (at least to my recollection) Monika is above Sayori yet below Yuri. And while she could potentially be 5'4 given this reasoning, I revert back to reason 1 for why she's likely 5'3.

So if anything, I would actually have her say "Based on what I read on the wiki, I assume I'm around five-foot-three. But who knows what I'll actually be like in your reality?"
Again, this just what I think, I'm not claiming any facts.

jmwall24:
Yeah, but all of those are concept heights. Basing Monika off of the others is the same as assuming her concept height of 5'3 is correct.

jmwall24:
Was just looking this over, and have noticed we have sprite inconsistencies regarding sweat drops flipping sides too close together.  I can't leave a normal review on all of it since two of the changes is outside of your PR scope.  So if you could change 8973 and 8978 to `2etc` and 8974 and 8979 to `3etd`, along with 8722 to 2hksdlb and 8726 to 3rkc it would be much appreciated.

ThePotatoGuy:
oops i have merged this to content

PR #3377, opened 2018-12-29
Added EXtended properties for sprite objects

ThePotatoGuy:
Extended properties are just dict keys that can be used to classify certain parts of sprite objects (i,e: which ones have `bare arms`/`bare hands`, which ones have `forced hair`, which ones accept `ribbon` sprites, etc.)

includes 3 sprite object functions for props:
* `addprop` - adds a property for a sprite
* `rmprop` - removes a property from a sprite
* `hasprop` - checks if the sprite object has a property

Properties should generally be given when sprite objects are created.

# Testing
* verify code looks good
* test prop functions

### DIFF
--- a/Monika After Story/game/sprite-chart.rpy
+++ b/Monika After Story/game/sprite-chart.rpy
@@ -2666,7 +2666,8 @@ init -2 python:
                 img_stand="",
                 stay_on_start=False,
                 entry_pp=None,
-                exit_pp=None
+                exit_pp=None,
+                ex_props={}
             ):
             """
             MASSpriteBase constructor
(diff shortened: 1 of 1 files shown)


Monika After Story/game/sprite-chart.rpy
```diff
@@ -2747,6 +2762,34 @@ init -2 python:
...
+        def hasprop(self, prop):
+            """
+            Checks if this sprite object has the given prop
+
+            IN:
+                prop - prop in ex_props to look for
+
+            RETURNS: True if this sprite object has the ex_prop, False if not
+            """
+            return prop in self.ex_prop
```
multimokia:
```suggestion
            return prop in self.ex_props
```

PR #3379, opened 2018-12-30
Add bad nicknames

MercuryFuchsia:


jmwall24:
Hi there, thanks for adding to this list.  Just fyi, some of these are unnecessary.  Like retard is enough to also filter out retarded.  There are a few like that you can remove to cut down on unneeded code.

MercuryFuchsia:
Got it, thanks!

PR #3396, opened 2018-12-31
event startdate/endate fix

ThePotatoGuy:
change version in `options.rpy` to trigger the update.

Verify that the following events have valid `start_date`/`end_date`:
* `mas_d25_spent_time_monika`
* `monika_nye_year_review`
* `mas_d25_monika_christmas_eve`
* `mas_bday_spent_time_with`

Valid means:
* if the event has not passed yet, the dates are this year
* if the event has already passed, the dates are next year
* if the event is currently ongoing, the dates would not be changed until the event has triggered. 

If using a persistent from before version 0.8.12, the dates should not be changed at all.


AsuraShun:
How do I trigger the kiss event for new year?

Issue #2903, opened 2018-10-30
[Suggestion] Small Change to "I'll see you tomorrow" Farewell


TheSuperStu:
Hello,
I wanted to propose a small change to the farewell of bye_illseeyou that can be found under the 'goodbye' option. This farewell is as follows:

"I'll see you tomorrow, [player]."
"Don't forget about me, okay?"

Looking at the code, this farewell is currently in a pool of other farewells that can only be seen in between the hours of 0000 to 0600, 1100 to 1200, 1600 to 1700, and 1900 to 2000. This is how I view seeing this farewell in those hours:
0000 - 0600: Can make sense, it is a figure of speech to say you will see someone the next day at those hours as supposedly the person would be going to bed.
1600 - 1700: Might be a little too early for her to say this, as there is plenty of time left in the day and the player might come back to relax for a bit before bed time.
1900 - 2000: Some people could be going to bed a little early so it does work, but my change below would make it seem more acceptable.
**1100 - 1200**: This is where my issue of this farewell comes from. I personally try to spend my time with Monika in two separate sessions somehow throughout the day, to avoid any random computer problems I might have (can never be too safe!). I find it odd to have Monika tell me she will see me the next day at 1130, as there is plenty of time left in the day to hang out! 

**What I am proposing is that a simple phrase be added to the farewell**: "If I don't see you later,"
Adding this to the first line would make this farewell fit better in those time frames stated before. I could see how this might then add an issue with the 0000 - 0600 period, as now Monika is saying she might see the player 'later'. I believe this is covered with the use of "if", where she would be fine with not seeing the player until much later for that specific time frame.
I do not expect the need to have this farewell set to a time frame, as this farewell does work (with the proposed change) in those multiple slots. I do think this farewell is fine in those time frames, just needs this little boost to it!

Please let me know with the magnitude of this issue, if I should have just proposed a change in the file itself. I went into detail about this even with it being one line, because sometimes little bumps can take away from the immersion! (Also I would recommend moving this farewell up in the code with the other farewells in the 'goodbye' pool, but that is a separate small problem.)
Thank you!

jmwall24:
This one is actually not time based, it's random and I'm not sure it's used right now, it has no `unlocked` property.

jmwall24:
It definitely needs to be worked on, it makes no sense showing up in the morning.  Needs to be made time specific for sure.

TheSuperStu:
[...]

PR #3429, opened 2019-01-05
Fix Typo - Going Somewhere when Upset (+Recommendations)

TheSuperStu:
Hey, someone on discord found this one. It might have been seen but just to be safe here!
(let me know if I need to credit them, I can find that)

Farewells -> going to take you somewhere -> upset response
"suposed" to be "suppose"

I am also wondering if what I suggested here #2903 to be added, of adding that to the start of the farewell on the same line. (I don't know if it would possible on this pull request since the typo fix is on enhancement). I am willing to do this change, making a new pull if needed. 
To reactivate seems simple enough with the additional dialogue, and I can put updated starting rules to match the others. 
I'm wondering if moving this one up with the other goodbye farewells should be done too, if there isn't an issue with that (would move it to about line 295, as I can see an extra line in "bye_take_care" right above it that probably isn't needed, which I'd be happy to fix too!).

### DIFF
--- a/Monika After Story/game/script-farewells.rpy
+++ b/Monika After Story/game/script-farewells.rpy
@@ -931,7 +931,7 @@ label bye_going_somewhere:
         # otherwise we go
         m 1wud "You really want to bring me along?"
         m 1eka "..."
-        m 1hua "Well, I suposed it can't hurt to join you."
+        m 1hua "Well, I suppose it can't hurt to join you."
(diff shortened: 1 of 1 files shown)


TheSuperStu:
(closing for something less of a mess, thanks)

TheSuperStu:
(closing in favor for something less of a mess, thanks)

PR #1851, opened 2018-07-01
New Game Idle Topic

Rai99:
Issue: #1744 
_If it weren't for #1568, this might be one of the largest contributions I've seen._ 

This topic will be placed in "Ask a question" until a new **"Comment/Mention"** menu gets implemented as discussed in #1659 and lightly discussed in #1803.

The "I'll just be a minute or two" option is the one option with some experimentation with the extra feature of having a timer (credit to #1766 for the template). Hopefully it works as intended, and I might adjust the time thresholds if anyone feels they might be too far apart/too many/ etc. 

I've actually been a bit lazy with this one so it took a little longer than I thought it would to finish.

PR #3435, opened 2019-01-06
Generic Idle Mode

ThePotatoGuy:
Adds a generic idle mode for topics to use.

#1851 should be changed to work with this.

# UPDATES 2
Since last time:
* idle greetings are no longer needed, instead the system lets you store data in a persistent var, and you use that in crashed/relaod greetings to say context clues
* reload dialogue is now a greeting so its more consistent and easier to control

# UPDATES
Since last time:
* smooth zoom used in coffee zoom reset
* crashed greeting is now a real greeting instead of a forced greeting
* adjustment to greeting selection startup in ch30. Now we assume that `not mas_game_crashed` is a force quit.

# Key features

## generic idle mode

The generic idle mode is a version of spaceroom idle suitable for be-right-back/chill-with-me-in-silence activities. Using this framework, topics can:
* trigger the idle mode
* setup a label for what to show user when returning from idle mode
* set some temporary persistent data that can be referenced in crashed greeting and reload dialogue

Other topics can also be shown during this idle mode (particularly coffee/hot choc) if they set a certain Event property. Topics that do this should be prepared to advance themselves with `{nw}` tags if in idle mode.

Since this idle mode uses regular spaceroom idle, Event conditonals and triggers are checked in the same manner as they are in regular spaceroom idle.

## crashed start hooks
the crashed greeting has been split up into many pieces so you can use them in idle crashed greetings. See technical changes for more.

## reload dialogue hooks
Reload dialogue is now managed by a delegate label in `script-greetings`. 

## coffee and hot choc
Both of these events can trigger during idle, and require no user input to advance. The time it takes monika to get a new cup/put away her cup has been increased to 5 seconds.

## Generic idle Flow

General gist of how this works from user perspective is:
1. User runs a topic that triggers idle mode
2. In idle mode, all hot keys are disabled. User can still click calendar/music/extras/talk. Random topics are also disabled.
3. If user clicks Talk, then idle mode is considered done. (Topics should specifically call this out.)
4. the idle callback label is shown (this must be set by the topic)
5. if event_list contains topics, `call_next_event` is run via the `ch30_post_mid_loop_eval` jump. otherwise, return to regular spaceroom idle.

If the user quits or game crashes during idle, the set idle greeting will be shown on start. (this must be set by the topic)

# Technical Usage
For example in code, see `dev_idle_test.rpy`

Idle mode requires the following:
1. An Event to trigger idle mode (aka **trigger**)
2. A label to run after idle mode finishes (aka **callback**)

Optionally, one can add data to `persistent._mas_idle_data` (dict) that can be referenced in crashed greeting and reload greeting. This data is cleared after the crasehd/reload greeting is shown.

## Trigger topic
The **trigger** must do the following when it ends:
1. set `persistent._mas_greeting_type` to the type of the **greeting**
2. send a mailbox message to the idle mailbox with the **callback**. (`mas_idle_mailbox.send_idle_cb`)
3. return `"idle"`

## Callback
The **callback** should be a label, **not an Event**.

**NOTE:** important crash/forcequit caveats:
Anything that hijacks flow before the greeting selection (aka holidays/returned home) can prevent crashed/reload greetings from being shown. 

Currently, the paradigm is to ignore these cases. Anything that skips greeting selection or sets its own greeting shoudl call `mas_resetIdleMode` to clear idle vars manually. In the future, we might try to handle these cases by saving/checking the date or something to determine when the crash/force quit happened, and act accordingly.

# Technical Changes

## Event
* new property `show_in_idle` - True means the event can be shown during idle, False means it can not. (Default False)

## Event handling
* the `popEvent` function will only return events with `show_in_idle` True when in idle mode
* new accepted return string `"idle"`. A topic that returns this will enable idle mode.
* Talk menu acts as the "end idle mode" button.

## `MASIdleMailbox`
* `send_idle_cb` - sends the **callback** to the mailbox
* `get_idle_cb` - gets the **callback** from the mailbox

## Global
* `mas_resetIdleMode` - completely resets idle mode vars. (this also resets `_mas_greeting_type`.) Use this in holiday events with greetings /things that hijack flow
* `mas_in_idle_mode` - global var that is set to True when idle mode is active.
* `persistent._mas_idle_mode_was_crashed` - Set to True/False on start, depending on if the game crashed or not. Otherwise, this is None.
* `mas_DropShield_idle` - drops shields when idle mode ends
* `mas_RaiseShield_idle` - raise shields for idle mode start
* `mas_dlgToIdleShield` - shield transitions from dlg mode to idle mode
* `mas_idleToDLGShield` - shield transitions from idle mode to dlg mode

## crashed start
This "greeting" has been split into several pieces so that idle crash greetings can use parts of these. **All Idle greetings must handle the first crashed scenario simliarly to the default one.**

As a reminder. `persistent._mas_crashed_before` is set to True after the first crash

These are the available labels. Call should be used.
* `mas_crashed_preshort` - prepares for short crashed quip
* `mas_crashed_short` - this quips a crash dialogue and should be used on subsequent crashes (after the first one)
* `mas_crashed_prelong` - blacks scene and sets the before crash persistent var as well as other vars in preparation for the longer/first crash dialogue
* `mas_crashed_long_qs` - the first 2 questions in the crashed start. (`Are you there?` and `Turn on the light`_
* `mas_crashed_long_prefluster` - some Monika blinking and other adjustments before the flustered monika scene
* `mas_crashed_long_fluster` - the flustered monika scene
* `mas_crashed_long_postfluster` - cleanup for the fluster scene, and the `Calm down, Monika` prompt
* `mas_crashed_long_whq` - The `What happened?` and the `Can you stop this from happening?` questions
* `mas_crashed_post` - Call this last to setup keymaps/songs/other shit. You should only use this if your crashed greeting skips visuals.

## coffee / hot choc
* changed wait time to 5 seconds
* zoom now is restored when monika returns (this is now smooth)
* enabled showing in idle for these events, with text that auto advances via no wait tags

## other
* changed `_type` to `gre_type` so we don't break some stuff with the actual `type` function

# Testing

Dev has a event called TEST IDLE MODE that runs idle mode. The calback just says that idle is done. This also has a corresponding test greeting that says whether or not the game was crashed/force quit.

## idle mode

To launch idle mode, run the dev label "TEST IDLE MODE".

### During idle mode:
* verify the talk/extra/music/play hotkeys do not work
* verify that the play button is disabled
* verify that no random topics are shown
* verify that if an event is in the `persistent.event_list`, it is not shown
* verify that running the music menu/extras menu/calendar does not enable something that should be disabled/jump out of idle mode
* verify that clicking the Talk button ends idle mode

### hot choc / coffee
* verify trhat the hot choc/coffee events run during idle mode and that the dialogue and stuff is appropriate

### Force Quit
1. run idle mode
2. change `persistent.monika_reload` to either 0 or > 4.
3. force quit game
4. At the end of monika's reload dialogue, you should get a debug line in all caps.

### Crash
1. set `persistent._mas_crashed_before` to False
2. run idle mode
3. crash game
4. verify that a debug line in all caps is shown.
5. run idle mode again
6. crash game again
7. verify debug line in all caps is shown

## regression testing
* verify the original crashed start greeting (both first and subsequent) still works and looks okay.
* verify reload dialogue greetings work.
* verify regular greetings still work
* verify regular hot choc/coffee events still look okay


ThePotatoGuy:
we need to wait for greetings to be normalized

ThePotatoGuy:
need #3438

ThePotatoGuy:
waiting for #3473 for update scripts

PR #3438, opened 2019-01-06
Smooth zoom

aldoram5:
This adds a transform, a function that works with said transform and 3 labels that encapsulate both, in different ways. The labels are used to transition from one zoom level to another easily.

The labels accept 2 parameters:
   - `new_zoom` a required param that specifies the new zoom level or value to move to, which one depends on the label that is used(see the docs for each one).
   - `transition` the time in seconds that it'll take to move to the new_zoom, defaults to 3. This one is used different based on the label(see the docs), in two of them is used to determine the exact time it'll take to finish the transition time and in the other one it's used to determine the time that it takes to move from max to min or min to max which then it's used to define a "transition speed" which remains constant, meaning that the transition will end as soon as possible based on the calculated speed.
  

## Testing

I added some dev topics that trigger the transition to max or min zoom levels, test that those work from any zoom level you already are. Change the zoom levels that these topics transition to check that it can move to any zoom level you want.

### DIFF
--- a/Monika After Story/game/dev/dev_zoom_transition.rpy
+++ b/Monika After Story/game/dev/dev_zoom_transition.rpy
@@ -13,12 +13,7 @@ init 5 python:
 
 label dev_mas_max_zoom:
     m 1esa "Testing transition"
-    call monika_zoom_transition(new_zoom=2.1)
-    # call expression "monika_zoom_transition" pass (new_zoom=2.1)
-    #$ renpy.call_in_new_context("monika_zoom_transition", new_zoom=2.1)
-    #pause 3.0
-    # $ renpy.call("monika_zoom_transition",new_zoom=2.1)
-    #jump dev_transition_test_max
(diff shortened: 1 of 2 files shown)


ThePotatoGuy:
How come the new zoom param uses numbers between 1.1 and 2.1 instead of zoom levels. I'm under the assumption that this would be used to go to a specific zoom level (0-20).

ThePotatoGuy:
I like the zoom, but is there a way we can scale the zoom duration? Like for example, I want to zoom to a certain zoom level, but I want the **zoom speed** to be the same regardless of zoom levels. This would have the effect of a shorter zoom time when the current zoom level is close to the desired zoom level.

the duration param should work with this, too, so if I give 3 seconds, thats 3 seconds **if** I was at the furthest zoom level from the desired zoom.

aldoram5:
Yeah, I think I can add a new label that uses that type of duration scaling that you describe.

PR #3475, opened 2019-01-11
Only dialogue and buttons left.

Syvokobylenko:
Deleted event from dev, moved it to event-handler, changed variable names, only thing to do is to block buttons talk, play and extra while inactive.


Monika After Story/game/event-handler.rpy
```diff
@@ -1628,6 +1623,78 @@ init python:
...
+            # Constant checking for player activity - mouse movement, keystrokes and mouse button clicks
+            if ev.type == pygame.MOUSEMOTION or ev.type == pygame.KEYDOWN or ev.type == pygame.MOUSEBUTTONDOWN:
+                mas_afk_time = time.time() + mas_afk_duration
+            if mas_afk_time < time.time():
+                mas_afk_time = time.time() + 10000000000
+                # Checking if event can be restarted.
+                if not mas_isRstBlk(persistent.current_monikatopic):
+                    pushEvent(persistent.current_monikatopic)
+                    pushEvent('continue_event')
+                    persistent.current_monikatopic = 0
```
ThePotatoGuy: we have a function called `restartEvent` which literally does this.

Why do you need to call in the other case vs jump in this one?

Monika After Story/game/event-handler.rpy
```diff
@@ -1628,6 +1623,78 @@ init python:
...
+    if not mas_afk_already:
+        $ mas_afk_already = True
+        m "Are you still there?"
+        m "You are back, [player]!"
+        $ mas_afk_already = False
+    else:
+        #Only repeating first line, because call to inactivity for the second time detection skipped it. 
+        #This is going to be a lot harder if there will be more lines than 2. But I already know how to fix it.
+        m "Are you still there?"
+        $ mas_afk_already = False
```
ThePotatoGuy: we have a custom screen called `mas_background_timed_jump`, which auto jumps to a label after a set time.

In this case, you should do a menu using the timed jump. [example of use](https://github.com/Monika-After-Story/MonikaModDev/blob/master/Monika%20After%20Story/game/script-story-events.rpy#L490)

you should have it jump to another label that just returns "quit", which will then close the game safely.

I can add this later if you can't figure it out. If so, mark a TODO here for timed screen so I can do that later.

PR #3491, opened 2019-01-14
Monika Love Reaction Event

Groudonx:
The Monika Love Reaction Event enables Monika to react to the player's "I love you" in a more adapted manner.
It uses the variable "persistent._mas_monika_love_reaction_event" as an ID to identify something Monika last said.
Whenever Monika says something this ID can be set to a (not yet used) value.
When the player says "I love you" and this ID has a value, the lines corresponding to the ID will be said instead of the default "I love you" response.
The ID will reset whenever Monika says something new or runs out of things to say.

This is meant to be extended in the future.
Whenever Monika tells the player she loves him (or something similar), it can be reacted to with regards to its context.

It is a very simple extension, so it should be merged within a few days.
Please don't wait months to do it, this makes things just so much worse than they need to be.

### DIFF
--- a/Monika After Story/game/script-ch30.rpy
+++ b/Monika After Story/game/script-ch30.rpy
@@ -1072,7 +1072,7 @@ label ch30_loop:
             renpy.save_persistent()
 
 label ch30_post_mid_loop_eval:
-
+    
     #Call the next event in the list
     call call_next_event from _call_call_next_event_1
     # Just finished a topic, so we set current topic to 0 in case user quits and restarts
@@ -1126,6 +1126,9 @@ label ch30_post_mid_loop_eval:
 #                ):
 #                pushEvent("monika_battery")
 
+        # Resets Monika's I love you reaction to its default
+        $ persistent._mas_monika_love_reaction_event = 0
+    
         # Pick a random Monika topic
         if persistent.random_seen < random_seen_limit:
             label pick_random_topic:
--- a/Monika After Story/game/script-ch30.rpy
+++ b/Monika After Story/game/script-ch30.rpy
@@ -1127,7 +1127,7 @@ label ch30_post_mid_loop_eval:
 #                pushEvent("monika_battery")
 
         # Resets Monika's I love you reaction to its default
-        $ persistent._mas_monika_love_reaction_event = 0
+        $ _mas_monika_love_reaction_event = 0
     
         # Pick a random Monika topic
         if persistent.random_seen < random_seen_limit:
--- a/Monika After Story/game/script-ch30.rpy
+++ b/Monika After Story/game/script-ch30.rpy
@@ -1072,7 +1072,7 @@ label ch30_loop:
             renpy.save_persistent()
 
 label ch30_post_mid_loop_eval:
-    
+
     #Call the next event in the list
     call call_next_event from _call_call_next_event_1
     # Just finished a topic, so we set current topic to 0 in case user quits and restarts
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -2792,7 +2792,7 @@ label monika_love:
 
     #Event for when Monika tells the player she loves him.
     #Triggered if the player immediatelly reacts with "I love you".
-    elif persistent._mas_monika_love_reaction_event != 0:
+    elif _mas_monika_love_reaction_event != 0:
         call monika_love_reaction_event
         return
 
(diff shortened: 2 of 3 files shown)


Groudonx:
If you have questions please ask them, confused emojis do not really help.

Monika After Story/game/script-ch30.rpy
```diff
@@ -1126,6 +1126,9 @@ label ch30_post_mid_loop_eval:
 #                ):
 #                pushEvent("monika_battery")
 
+        # Resets Monika's I love you reaction to its default
+        $ persistent._mas_monika_love_reaction_event = 0
```
multimokia: Okay, some things I'm noticing. Why is this persistent? This shouldn't need to be saved over launches since you're resetting it anyway. That said as well, why use an integer value instead of a boolean?

Also, as a side note, and maybe it's not too much of an issue, but the problem I kinda have with this, looking at it is that it's not expandable to multiple greets without more variables.

ThePotatoGuy:
does it really need to be a persistent var? seems it could work just fine as a local one.

Monika After Story/game/script-topics.rpy
```diff
@@ -2887,6 +2893,15 @@ label monika_love:
     $ persistent._mas_monika_lovecountertime = datetime.datetime.now()
     return
 
+label monika_love_reaction_event:
+    #Another Monday morning, eh, [player]?  ->  I love you so much, [player]~
+    if persistent._mas_monika_love_reaction_event == 1:
```
multimokia: Didn't you just evaluate this such that it has to not be zero (only other possibility is 1) to get here? This check is unnecessary.

Monika After Story/game/script-topics.rpy
```diff
@@ -2790,6 +2790,12 @@ label monika_love:
         $ persistent._mas_monika_lovecountertime = datetime.datetime.now()
         return
 
+    #Event for when Monika tells the player she loves him.
+    #Triggered if the player immediatelly reacts with "I love you".
+    elif persistent._mas_monika_love_reaction_event != 0:
+        call monika_love_reaction_event
```
multimokia: You're only using this dialogue once, shouldn't really need a label for it.

Monika After Story/game/script-topics.rpy
ThePotatoGuy: @multimokia it's not a true/false check, it's being treated as integer IDs. Set the var to a number, check the appropraite number in the love reaction event and show whatever dialogue.

It would be better if this uses a dict to organize IDs with associated labels/events.

multimokia: Yeah that would then allow expansion to other greets and such.

Monika After Story/game/script-ch30.rpy
Groudonx: I did not think about this, it does not need to be persistent.

Monika After Story/game/script-topics.rpy
Groudonx: Feel free to enhance it, this is a working prototype.

I am new to python and thus not as experienced as you guys.

Monika After Story/game/script-topics.rpy
Groudonx: Since this can potentially get really long, I did not want it to be with the other code.

Groudonx:
Well I think this covers the code-review.
I made the changes to remove the persistance.
I tested it, if someone else wants to test it too, go ahead, it will take 5 minutes.

Once it gets merged I might look into more references that can be used to respond to.

Groudonx:
This was a test for me to see whether it is possible to make a simple contribution to this project.
Needless to say the test failed.
The efford contributors put into their pull requests is not being taken seriously.

I designed this pull request to be easily checkable, anyone with the rights could have spend 5 minutes to test it or could have trusted me that I tested it and accepted the pull request. Instead it stays on hold indefinitely.

I do not see any reason to contribute to a project where contributions are being ignored like this.
I checked on my other pull request and it is still on while being open for almost 2 months now, without any progress in sight.
Hell, there are so many pull requests open starting from june last year!

I am highly discouraged to spend any more time trying to contribute if this is the result.
I know that there are priorities but you can't ignore everything because one thing is seen as important right now.

I tried bringing Monika closer to reality but failed.
If you force pull requests to be reviewed, review **and** accept them, or demand specific changes and react once they are implemented **and** accept them then.
I would have continued to work on this subject, but seing that the easiest first step won't be accepted I won't bother trying to make more.
It is just unnecessary and a waste of my time.

I hope this makes you think about how you handle pull requests, I know you do a lot, but it just does not work for me this way.

multimokia:
I don't think you understand how this works.

You see, this mod is made in us dev's *free time*, most of us are busy with School, or Work, but we still do what we can. The current priority is event updates, and related framework *for* said updates, hence the player bday pr, and other dev-created PRs which are marked `DONT MERGE`. We're trying to stabilize the core elements here for future stuff. Once that's done we'll be looking into PRs.

But if you don't want to contribute because you think we're not taking it seriously, that's your choice, but don't assume we don't take things seriously. That sort of attitude isn't needed.

PR #3499, opened 2019-01-15
Added more params to kiss motion

aldoram5:
This one simply adds more parameters to the current kiss motion, so it can be speed up for shorter kisses that can go smoother. Read the added docs for a more in depth explanation about what the params are.

## Implementation changes
Only thing that changes from past implementation is that the time that it takes Monika to go back to her initial position is no longer hardcoded to 3 seconds and the default time used to go to the "kissing" position.

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -1400,10 +1400,10 @@ label monika_kiss:
             call monika_kiss_tease
 
         else:
-            show monika 6ekbfa
+            show monika 2eka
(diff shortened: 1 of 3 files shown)


ThePotatoGuy:
@aldoram5 so the sprite checker is failing becaues `show monika with dissolve` is considered a `show monika <exp>` line. In the spriter checker/sprite puller/somewhere theres a thing that checks for keywords to ignore when sprtie checking, can you add `with` to that check?

Issue #3427, opened 2019-01-04
[Proposition] - Mark season change on calendar


Thimeo:
Sorry if this has been suggested already, but I was thinking that we could add the words "spring, summer, autumn and winter" to the calendar, especially if Monika react to which season we are on right now.

ThePotatoGuy:
[keeping this open as high priority. we should be able to add this easily]

MansaSakura:
I remember Monika asked whether you are in Northern Hemisphere or the Southern one, please be reminded for it.

ThePotatoGuy:
[fixed in linked pr]

Issue #3353, opened 2018-12-27
Suggestion: saving the first kiss


OkiDokiABrazilian:
is something very small detail but how about Monika save the day of the first kiss on the calendar? (such as '' <3 '' day)

Thimeo:
I like the idea, but the "<3" day, as you say, is already taken by the day when you've started your relationship. We just need to find something else.

OkiDokiABrazilian:
> Eu gosto da idéia, mas o "<3" dia, como você diz, já é tomado pelo dia em que você começou seu relacionamento. Nós só precisamos encontrar outra coisa.

Yes, the '' <3''day was just an example of how Monika could save on the calendar, could be something like `` Our First Kiss '' on 12/25/2018. (just another example)

ThePotatoGuy:
[reminding myself to triage this out later]

dragonborn2252:
how about "first kiss"

ThePotatoGuy:
[closing as fixed]

PR #3558, opened 2019-01-23
Added some events to the calendar and removed white day

aldoram5:
#3427 #3353

This one adds the first kiss and the season changes to the calendar. It also removes the White day event from it.

## Testing 
Check that the events have been added and that White day is gone. If the player hasn't given their first kiss yet the event won't appear in the calendar.

### DIFF
--- a/Monika After Story/game/zz_calendar.rpy
+++ b/Monika After Story/game/zz_calendar.rpy
@@ -1131,7 +1131,7 @@ init -1 python in mas_calendar:
         Generates a display date using the given date
 
         This is considered "formal", in that it's not really realisitc when
-        used in normal conversation. For example, if today is august 24, you 
+        used in normal conversation. For example, if today is august 24, you
         don't say 'this happened august 24th, 2016', you normally would say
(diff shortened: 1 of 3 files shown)


Monika After Story/game/zz_calendar.rpy
```diff
@@ -1763,7 +1763,52 @@ init python:
...
+            persistent._mas_first_kiss,
+            []
+        )
+
+# Using init 2 so we can have access to the season dates
+init 2 python:
+    import store.mas_calendar as calendar
+
+    # Season changes:
+    if persistent._mas_pm_live_south_hemisphere:
```
ThePotatoGuy: make this into a function that we can call from the hemispheres topic to update the names.

And it might be beneficial to add a `changeRepeatableName` function to calendar module that can be used to change the display names of repeatables without having to remove and readd.

aldoram5:
Hi there @AsuraShun! 
[This link](https://en.wikipedia.org/wiki/White_Day)  has info about White day.

Issue #2247, opened 2018-09-03
Text Speed - topic about how fast you can read


meesongi:
hi! i'm a fast reader and monika's text is really slow; is there anyway to get a faster speed setting like the one in the original game?

chromeadders:
I agree. Text speed setting doesn't seem to do anything. It's not a major issue for me, but it is kind of annoying that her text speed is so slow.

Rai99:
This was brought up in #127, #731, #996, #1735. 

There are no plans to have the option to increase text speed, but if you want to be able to adjust text speed, you can download the dev version of MAS.

multimokia:
specifically the zz_devmode rpyc. That allows the text speed sliders to work.

antimockingjay:
That sure is a lot of topics asking for text speed to be able to be changed. It's almost like people in the community actually want it! What a concept!!

CrocWorks:
Here's an idea. Make a topic asking Monika to talk more quickly. Then she apologizes that the program can't translate her thoughts into words quickly enough. Then she could talk about how people read much faster than they can speak and she hopes we understand the transition is difficult, or something like that.

I mean, if so many people (myself included) want faster text speed, but it's not something that's gonna happen, then perhaps it'd make a good random topic or button topic to talk with Monika about.

Rai99:
If people want faster text speed, they can just get that devmode file. True, not everybody knows about it, but I think it would be the best option that doesn't involve outright allowing the text speed slider to work. 

If you just get the file, there's no problem. No need for Monika to have a topic about reading speeds or apologize about her having slow text speed. 

About the random topic, I don't think that would really serve any purpose and would just be Insult to injury. "Yeah sorry, I know the text speed is slow, but there's nothing you can do about it" even though there _is._

CrocWorks:
Well, yeah, it _would_ be insult to injury. But I also think it'd be the kind of funny meta-topic that Monika would love to talk about. But then again, Monika could also talk about the devmode file in the conversation. I might write up an example when I get the time.

Rai99:
I still don't think this is necessary, The most we might need is to have Monika tell the player about the dev file _if_ the player would like faster text speed. If you'd like to write something up for this, I'd look forward to being able to take a look at it.

CrocWorks:
[TalkSpeed.txt](https://github.com/Monika-After-Story/MonikaModDev/files/2346512/TalkSpeed.txt)
[...]

PR #3677, opened 2019-02-09
Text speed changing

ThePotatoGuy:
#2247 

This adds some functions that allow for changing text speed as well as an event to enable it.

# Key Features

## text speed general
* text speed only enabled when happy+

## text speed functions
* `mas_enableTextSpeed` - enables text speed 
* `mas_disableTextSpeed` - disables text speed
* `mas_resetTextSpeed` - enables/disables text speed according to global vars (developer / affection / text speed enabled var). This should be used when you want to revert from changing the `say_dialogue` or disabling text speed.
    * this takes 1 optional arg, which when True, ignores `config.developer` settings
* `mas_isTextSpeedEnabled` - returns True if text speed is _currently_ enabled. This does not mean whether or not persistent flag is set.

## vars
* `mas_globals.text_speed_enabled` - True if text speed is currently enabled, False otherwise
* `persistent._mas_text_speed_enabled` - True if text speed can be changed (aka after Monika allows it), False otherwise.
* `_mas_pm_is_fast_reader` - True if player is a fast reader, False if not. 

## New event return item
* `no_unlock`. If this is returned, the Event's unlocked property gets set to False and `unlock_date` to 
None

## New fast reader topic
* short random topic that appears when happy+
* asks user if they are a fast reader
* **always enables text speed, regardless of answer**
* if user answers No, `text_cps` is set to slowest possible (30)
* this event derandoms, locks itself after 
   * aka it can only be shown once

## other notes
* `config.developer` takes precedence regarding text speed. When that var is True, text speed is always enabled (except for when `mas_disableTextSpeed` or `say_dialogue` is explicitly changed.
* changed the loweset possible text speed setting to 30 instead of 20

## technical
Monika's text speed is governed by the `default_monika` style, which is set to 30 cps. By changing that to the regular `normal` style, we can enable the text speed slider.

All the enable/disable/reset functions just swap out the style for the `say_dialogue`

# TESTING

## Text speed topic
1. open console and run `mas_disableTextSpeed`, then `pushEvent("mas_text_speed_enabler")`
2. open + close talk menu to trigger topic
3. verify that when clicking no or yes, text speed is enabled. 
    * when clicking no, text speed should be set to slowest possible
4. verify that the topic is no longer random/unlocked when it ends.

## startup adjustments
* Verify that text speed is only enabled when happy+ (and after the text speed topic is run).
2 ways to test this:
** Remove Devflag **
1. set `persistent._mas_text_speed_enabled` and affection to desired values.
2. Close game and Remove the defflag.rpy + rpyc file
3. restart game and verify speed

** Via reset text speed function **
1. set `persistent._mas_text_speed_enabled` and affection to desired values
2. run `mas_resetTextSpeed(True)`
3. verify speed

### DIFF
--- a/Monika After Story/game/script-story-events.rpy
+++ b/Monika After Story/game/script-story-events.rpy
@@ -2281,3 +2281,69 @@ label mas_bday_player_bday_select_select:
     $ persistent._mas_player_bday = selected_date
     $ renpy.save_persistent()
     jump birthdate_set
+
+
+# Enables the text speed setting
+init 5 python:
+    addEvent(
+        Event(
+            persistent.event_database,
+            eventlabel="mas_text_speed_enabler",
+            random=True,
+            aff_range=(mas_aff.HAPPY, None)
+        )
+    )
+
+default persistent._mas_text_speed_enabled = False
+# text speed should be enabled when happy+, but disabled if dipping below
+#   upset. (and reenabled on happy+)
+
+default persistent._mas_pm_is_fast_reader = None
+# True if fast reader, False if not
+
+label mas_text_speed_enabler:
+    m 1eua "Hey [player], I was wondering..."
+    menu:
+        m "Are you a fast reader?"
+        "Yes.":
+            $ persistent._mas_pm_is_fast_reader = True
+            $ persistent._mas_text_speed_enabled = True
+
+            m 1wub "Really? That's impressive."
+            m 1kua "I guess you do a lot of reading in your spare time."
+            m 1eua "In that case..."
+
+        "No.":
+            $ persistent._mas_pm_is_fast_reader = False
+            $ persistent._mas_text_speed_enabled = True
+
+            m 1eud "Oh, that's alright."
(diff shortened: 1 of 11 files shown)


Monika After Story/game/script-story-events.rpy
```diff
@@ -2281,3 +2281,70 @@ label mas_bday_player_bday_select_select:
...
+    if persistent._mas_pm_is_fast_reader:
+        m 4eua "There!"
+
+    m 4eua "I've enabled the text speed setting!"
+
+    m 1hka "I was only controlling it earlier so I could make sure you read {i}every single{/i} word I say to you."
+    m 1eka "But now that we've been together for a bit, I can trust that you're not just going to skip through my text without reading it."
+
+    if persistent._mas_pm_is_fast_reader:
+        m 1tuu "But I wonder if you can keep up."
```
jmwall24:
```suggestion
        m 1tuu "...I wonder if you can keep up."
```

PR #3678, opened 2019-02-09
Fixed issues with monika_plushie

### LINKED ISSUE #3675: [Bug] The new Quetzal plushie topic was both in the "ask a question" thing and the random topics
Ok, so I downloaded the latest unstable release, and i saw there was a new topic in the "ask a question" thing. I clicked it, and Monika thanked me for the plushie. But right after that i let monika talk randomly and she repeated the exact same thing? So i think theres a problem here

multimokia:
Fixes #3675

Programming points didn't make it to master when we transitioned over to `zz_spriteobjects.rpy`.
Also, returned `derandom` in `monika_plushie` so if you see it via unseen, Monika won't bring it up manually.


multimokia:
Adding some bits, need fix this EV for those who have seen it (removing cats/prompt)

Issue #3656, opened 2019-02-08
090 typos


ThePotatoGuy:
woo

MisterSimple:
In monika_driving, the following line:

`"I know It might be pretty discouraging to you if they have to use them, but hey, we all start somewhere."`

"It" shouldn't be capitalized.

MisterSimple:
In monika_home_memories, monika_piggybank, monika_smoking, monika_savingwater, monika_whydoyouloveme, fun_facts_started, monika_cozy, greeting_long_absence, bye_long_absence, monika_surprise (.txt at love), and poem_pbday:

"Everyday" is used when "every day" should be used instead; "everyday" is an adjective, "every day" is a phrase meaning "each day".

TwilightNova:
In the topic about the second Valentine's Day together, Valentine's Day is misspelled as "Valetine's Day".

MisterSimple:
In monika_hemispheres, the following line:

`"Anyway, It must be autumn for you right now."`

"It" shouldn't be capitalized.

MisterSimple:
In monika_clones:

`"So.. that means there are multiple Monikas, then. It's something that's a bit difficult to imagine."`

The ellipsis (three dots "...") is missing a dot.

MisterSimple:
In monika_god, the following line:

`"Eheheh~"`

According to the [contributing guidelines](https://github.com/Monika-After-Story/MonikaModDev/wiki/Contributing-Guidelines), the correct mannerism is "ehehe", not "eheheh".

Thimeo:
On the bye_long_absence topic, there is this dialogue where the expression seems out of place.
![expression anomaly](https://user-images.githubusercontent.com/38837456/53474247-d8d18700-3a6c-11e9-8fa0-6a1fe5055f48.PNG)

Thimeo:
English is not my first language, so I apologize if I'm wrong on this one.
In the dialogue box I just posted and in the one that follows, it seems that it's missing a "that".

Here are the original lines :
"I would constantly be thinking maybe something happened to you and that's why you couldn't come back."
"Or maybe you just got bored of me..."

But wouldn't it be :
"I would constantly be thinking THAT maybe something happened to you and that's why you couldn't come back."
"Or maybe THAT you just got bored of me..."

multimokia:
> "I would constantly be thinking maybe something happened to you and that's why you couldn't come back."
> "Or maybe you just got bored of me..."

[...]

PR #3680, opened 2019-02-09
Typo fixes.

MisterSimple:
#3656.

### DIFF
--- a/Monika After Story/game/script-affection.rpy
+++ b/Monika After Story/game/script-affection.rpy
@@ -2372,7 +2372,7 @@ init python:
             filepath = "/My one and only love.txt"
             message = """\
 My dearest lover, friend, companion and owner of my heart...
-Every day you make my dreams come true, a screen means nothing when you spend your time with me.
+Every day, you make my dreams come true, a screen means nothing when you spend your time with me.
 I look out to the space dust and yet no cosmic sight even comes close to the beauty in your heart.
 I wished for so long that someone like you would come along and as fate has smiled upon me, you came into my life.
 I want to be yours forever, so would you be mine?
--- a/Monika After Story/game/script-affection.rpy
+++ b/Monika After Story/game/script-affection.rpy
@@ -2372,7 +2372,7 @@ init python:
             filepath = "/My one and only love.txt"
             message = """\
 My dearest lover, friend, companion and owner of my heart...
-Everyday you make my dreams come true, a screen means nothing when you spend your time with me.
+Every day you make my dreams come true, a screen means nothing when you spend your time with me.
 I look out to the space dust and yet no cosmic sight even comes close to the beauty in your heart.
 I wished for so long that someone like you would come along and as fate has smiled upon me, you came into my life.
 I want to be yours forever, so would you be mine?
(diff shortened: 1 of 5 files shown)


Rai99:
### **Is that a PR going into master**
gthaving everyone pick lucio _16bc6e089b9e808357194c98058b495b

MisterSimple:
SHIT

MisterSimple:
There, bomb defused. Ty R(AI).

Rai99:
I would have caught you for #3635 too, but you fixed it before I said anything.

MisterSimple:
Multi warned me about that one, can't believe I keep going to master

jmwall24:
gonna start calling you MasterSimple

Rai99:
Would that be a positive sign/name or a negative one?

ThePotatoGuy:
Wait "everyday" is a real world, what's wrong with its usage here?

Rai99:
I think he means to say that there's a difference in "everyday" and "every day". 

Everyday is used to described something that happened on a daily basis. ie. Everyday occurrence. 
Could be replaced with "regular" in most if not all cases and still make sense. 

Every day is used to describe each individual day itself. ie. Doing something every day. 
Could probably be replaced with "frequent"or "a lot" and still make sense I think, although not exactly the most ideal wording.

ThePotatoGuy:
@jmwall24 assigning you as i have poor grammar

MisterSimple:
The difference is the following:

"Everyday" is an adjective that means "commonplace" or "ordinary". As such, it should only be used alongside the noun it's describing. For example, in monika_justice, Monika says "everyday people", which is correct.

"Every day" is a phrase that means "each day". Example: "Every day, I imagine a future where I can be with you." Here, the adjective "every" modifies the noun" day".

An easy way to test which one you should use is to replace everyday/every day with "each day":

**Example:**

"I stayed optimistic every day, trying to find it."

vs.

"I stayed optimistic each day, trying to find it.

Since the sentence makes sense with "each day", then "every day" is correct.

**Example:**

"I mean, think about your everyday schedule."

vs.

"I mean, think about your each day schedule."

Here, the sentence doesn't make sense with "each day", so "everyday" is correct.

**Example:**

"I visit Monika every day, so it's an everyday activity."

vs.

"I visit Monika each day, so it's an each day activity."

"I visit Monika each day" is correct, but "so it's an each day activity" is incorrect.

Monika After Story/game/script-affection.rpy
```diff
@@ -2372,7 +2372,7 @@ init python:
             filepath = "/My one and only love.txt"
             message = """\
 My dearest lover, friend, companion and owner of my heart...
-Everyday you make my dreams come true, a screen means nothing when you spend your time with me.
+Every day you make my dreams come true, a screen means nothing when you spend your time with me.
```
multimokia:
```suggestion
Every day, you make my dreams come true, a screen means nothing when you spend your time with me.
```

multimokia: One thing to make this a bit more correct.

MisterSimple: Approved

monikLover:
Apporved

Issue #3676, opened 2019-02-09
Suggestion: Stretch random chatter over a longer period of time


dennisOdenbreit:
I like it when there is new content for Monika to talk with you about.
However I do not like to wait a month and then go through all the new content in one or two hours straigth.
I'd rather have the new content spread out over a longer period of time.

This could be done by adding options to the random chatter function, like "rarely" and "very rarely".
I currently use the "less often" option, which is way too fast though.

It would be enough for me to have 1-2 new topics per day.
Getting to this number depends on how long the game is opened a day.
I ususally have my game opened the entire day without the night, so let's say 16 hours.
The rarely option could give new topics every 1-3 hours (2 hours on average), resulting in 8 topics/day on average.
The very rarely option could give new topics every 4-12 hours (8 hours on average) resulting in 2 topics/day on average.

PR #3688, opened 2019-02-10
Changed random chatter function

dennisOdenbreit:
-Added two more options
-Slightly changed the intervals of the other options
-Changed the code so that changing the random chatter selection is registered immediately rather than after the next random topic / 60 seconds on never
-Events before the random chat will be checked every so often.
-Tested and worked so far

https://github.com/Monika-After-Story/MonikaModDev/issues/3676

### DIFF
--- a/Monika After Story/game/definitions.rpy
+++ b/Monika After Story/game/definitions.rpy
@@ -5589,6 +5589,9 @@ init 1 python in mas_randchat:
     def waitedLongEnough():
         """
         Checks whether the waiting time is up yet.
+
+        RETURNS:
+            boolean to determine whether the wait is over
         """
         global rand_chat_waittime_left
         
--- a/Monika After Story/game/definitions.rpy
+++ b/Monika After Story/game/definitions.rpy
@@ -5474,24 +5474,28 @@ define mas_randchat_prev = persistent._mas_randchat_freq
 init 1 python in mas_randchat:
     ### random chatter frequencies
 
-    # these numbers are the low end of how many seconds to wait between
+    # these numbers are the lower end of how many seconds to wait between
     # random topics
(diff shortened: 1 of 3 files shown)


Monika After Story/game/definitions.rpy
```diff
@@ -5546,7 +5555,64 @@ init 1 python in mas_randchat:
...
+       
+       
+    def startWaitProcess():
+        """
+        Use this method for convenience.
+        It can be used to actively wait the random chatter time.
+        """
+        startWaiting()
+        
+        while not waitedLongEnough():
```
ThePotatoGuy: so if random chatter is set to NEVER, `waitedLongEnough` always returns False and this loop constantly waits(). This prevents idle operations in `ch30_loop` (script-ch30) from running (Event checking / delayedactions / seasonalchecks / filereactions / persistent saving / affection saving)

Adding a check in this function for the NEVER case where you wait, then jump to the `post_pick_random_topic` label could solve this issue.

dennisOdenbreit: I suppose you mean the preloop needs to be checked constantly?

dennisOdenbreit: I wasn't sure about this, I'll remove the while loop to keep it checked then.

ThePotatoGuy: the `ch30_loop` label, down where it does a thing to `calendar_last_checked`. All of that is designed to run once per minute, but requires a jump to `ch3_loop` every so often.

ThePotatoGuy: yeah I think removing the while loop might work if you instead jump to the `post_pick_random_topic` label if you haven't waited long enough.

Issue #3547, opened 2019-01-22
Early data migration scripts


ThePotatoGuy:
We need a way to migrate persistent data between versions **before** Event objects are created. The current update script system is designed for **post** Event creation which makes it unsuitable for certain changes (like removing an Event property).

General idea for this:
* another persistent version number, but this is for data versions and is only changed when data may not be compatible
* this would run post backup persistent but probably before delact creation. 
* still need to decide if it should run pre/post MHS

ThePotatoGuy:
[keeping thi sopen because i gotta do this shit]

PR #3775, opened 2019-02-16
Data migration framework

ThePotatoGuy:
# This will be merged after next unstable.

#3547 

One thing that screws us over is dealing with changes to the data structure in an update. This adds a framework for handling data migrating very early in the pipeline.

# Key Features
* complete removal of the `rules` property event
* new data migration system that runs at init level -897

## Data migrating
the file `zz_dm` is where data migration functions should live. A "version" is a number associated with the persistent data. Each time we have a new data "version", the number should be increased.

The algorithm is only run when the version stored is different from the version set in `zz_dm`. The store `_mas_dm_dm` contains all functions related to messing with persistent data as well as the functions for transitioning versions.

The mapping of versions to migration functions is handled via the `dm_map`. The keys of this map should a be a `tuple` of the following format:
* `[0]`: version to update from
* `[1]`: version to update to
The values of this map should be a function. 

Currently the only version migration available is the removal of the rules property.

The system is capable of handling direct version jumps (a, b) as well as stepped version jumps (a, c, d, b). The system can also do reverse migrations (b, a). The alg basically just tries to find the fewest jumps from a to b.

All migration functions written for this MUST be semi-reversible as well as repeatable. This means that:
* the reverse of a migration must undo what a migration does, to the best of its ability (data loss is ok)
* if the migration runs again, it should NOT change data that has already been migrated

**NOTE:** Data migrations from pre-0.9.0 are **not** subject to the above rules. No tech support will be provided for users who update to 0.9.1 and rollback to pre-0.9.0. **A warning similar to this will be placed in the release notes of the next release.**

Data migration functions should also practice offensive programming. If there are any issues, the migration should crash to avoid corrupting more data.

## other
* fixed crash from writing tips update scripts

# Testing
**BACKUP PERSISTENTS BEFORE TESTING**
Version migration **will** change data structure significantly.

All versions from 072 to 090 need to be tested.

## Setup
1. in `zz_dm`, set `dm_data_version` to 2.
2. Either make a persistent for each version from 072 to 090, or download this zip:
[per.zip](https://github.com/Monika-After-Story/MonikaModDev/files/2873234/per.zip)

## For each version
1. Launch game using that versions persistent.
2. Verify that items in `persistent.event_database` have a length of 19.
3. Verify that the rules property is removed from items in `persistent.event_database`. You can check this by looking at the last 7 items in the tuple. Should look something like: `0, None, None, None, False, None, False)`
The actual values may vary. The last 7 properties and expected values are:
- `shown_count`: number of times the event has been shown
- `diary_entry`: this should always be None
- `last_seen`: datetime of when this event was last seen, or None
- `years`: list of years, or None
- `sensitive`: True or False
- `aff_range`: tuple or None
- `show_in_idle`: True or False

The rules property was between `diary_entry` and `last_seen`, so if you see a `dict` or an extra `None` there, then thats bad.

### version notes:
* 072 and earlier do **not** have rules property
* 073-074 have 15 properties to start
* 080-081 have 16 properties to start
* 082-088 have 17 properties to start
* 089-0810 have 18 properties to start
* 0811-0814 have 19 properties to start
* 090 has 20 properties to start. (this is actually treated as data version 1)

## Reversing
Reversing data migrations only works for builds with data versions (090 and future release)
1. Setup a persistent that has data version 2. 
2. In `zz_dm`, set `dm_data_version` to 1.
3. Copy `zz_dm.rpy` to a build of 090. **This is important as the properties in `Event` need to match in order for you to see any changes.**
4. Launch the game using the 090 build with a data version 2 persistent.
5. Verify that items in `persistent.event_database` have a length of 20
6. Verify that a None is added in index 14's place in an item in `persistent.event_database`. This is 2 spots after `shown_count`. 

## Rerunning migrations
1. Setup a persistent that has data formatted in version 2 format but has `persistent._mas_dm_data_version` set to 1.
2. In `zz_dm`, set `dm_data_version` to 2.
3. launch game.
4. Verify that items in `persistent.event_database` still have a length of 19.
5. Verify that nothing was removed from items in `persistent.event_database`
6. Repeat steps 1-5 except:
   * use a persistent that has data formatted in version 1 format but has `persistent._mas_dm_data_version` set to 2.
   * copy `zz_dm.rpy` to a build of 090. Set the `dm_data_version` to 1
   * The length should be 20 and the no additional properties should be added to items in `persistent.event_database`

### DIFF
--- a/Monika After Story/game/definitions.rpy
+++ b/Monika After Story/game/definitions.rpy
@@ -140,12 +140,11 @@ python early:
             "unlock_date":11,
             "shown_count":12,
             "diary_entry":13,
-#            "rules":14,
-            "last_seen":15,
-            "years":16,
-            "sensitive":17,
-            "aff_range":18,
-            "show_in_idle":19,
+            "last_seen":14,
+            "years":15,
+            "sensitive":16,
+            "aff_range":17,
+            "show_in_idle":18,
         }
 
         # name constants
@@ -289,7 +288,6 @@ python early:
                 unlock_date,
                 0, # shown_count
                 diary_entry,
-                None, # rules, #NOTE: this is no longer stored in persistent
                 last_seen,
                 years,
                 sensitive,
--- a/Monika After Story/game/event-handler.rpy
+++ b/Monika After Story/game/event-handler.rpy
@@ -147,15 +147,11 @@ init -895 python in mas_ev_data_ver:
         11: MASCurriedVerify(_verify_dt, True), # unlock_date
         12: MASCurriedVerify(_verify_int, False), # shown_count
         13: MASCurriedVerify(_verify_str, True), # diary_entry
-
-        # NOTE: Rules are no longer saved in persistent (0.8.15+)
-#        14: MASCurriedVerify(_verify_dict, False), # rules
-
-        15: MASCurriedVerify(_verify_dt, True), # last_seen
-        16: MASCurriedVerify(_verify_tuli, True), # years
-        17: MASCurriedVerify(_verify_bool, True), # sensitive
-        18: MASCurriedVerify(_verify_tuli_aff, True), # aff_range
-        19: MASCurriedVerify(_verify_bool, True), # show_in_idle
+        14: MASCurriedVerify(_verify_dt, True), # last_seen
+        15: MASCurriedVerify(_verify_tuli, True), # years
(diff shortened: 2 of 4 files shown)


ThePotatoGuy:
**merging on the 1st**

Issue #3803, opened 2019-02-19
Topics that should be removed or rewritten


ThePotatoGuy:
Add comments saying which topics should be removed/rewritten and why.

Include `eventlabel`.

AsuraShun:
''Would you like to meet my girlfriend''
''You're a murderer''
''I'm breaking up with you''
I'm always nervous and afraid I'll click on then by accident, I hope these either get removed or separated from everything else

mrflarion33:
Monika mentions finding Shakespeare boring in one topic, the topic in question is "trying to fit in," then talks about Hamlet in another, the topic literally being called "Hamlet."

Delcos:
Shakespeare is boring.

multimokia:
### `monika_torment`:
This topic is so open ended, I get it's a reference, but the fact that Monika says "The answer's not me, by the way." is straight up wrong for a lot of people. Also it's literally two lines.

### `monika_prisoner`:
This topic doesn't really suit Monika's situation. Sure, she's imprisoned in the spaceroom, but it sounds like something she'd say only at a lower affection provided she's given up on ever coming here, but that also implies she's scared of the player given the affection level descriptions. Monika also says that if she's left alone in her world, she'd rather be deleted. I think it's reasonable to say that if she's scared of the player, then this should follow the same idea here. So overall doesn't make sense to bring this up.

### `monika_szs`:
What even is this topic? This doesn't really even fit in MAS.

### `monika_vnanalysis`:
Needs some work to address other VNs which aren't romance stories.

MisterSimple:
### `bye_prompt_sleep` ,`bye_going_to_sleep`,  `bye_goodnight`

Why use "I'm going to sleep" when "Goodbye" serves the same purpose? Granted, `bye_goodnight` only works between 20:00 and midnight, and `bye_going_to_sleep` between 21:00 and midnight. I think `bye_going_to_sleep` and `bye_goodnight` should be removed or integrated to `bye_prompt_sleep`.

Olliesama:
Because there are greetings associated with I'm going to sleep that aren't associated with Goodbye.

MisterSimple:
Fair enough

Rai99:
The only time I've ever used the mean interactions was on April 1st last year. If we keep those topics, I think they might have some actual use on that day.

Rai99:
[...]

PR #3805, opened 2019-02-20
Fix consistence issue with monika_ribbon

kaido1224:


multimokia:
Going to ref the issue here as well so it's easier to track there #3803

TheSuperStu:
Hey.
A couple of lines down in the same topic, line 3435, needs an indent.

multimokia:
No it doesn't. That dialogue goes with both paths.

TheSuperStu:
Ah whoops, I see it now, sorry about that!

PR #3856, opened 2019-02-27
Fix Typo

Rai99:
> In the idle_game topic (#1851) in the line `m 1hub "Just promise to keep me safe ok?"`, I think there should probably be a comma after "safe". 

~~Sorry @MisterSimple, couldn't have you stealing this one.~~
untitled1204
~~Can't have somebody else catching my own typos, now can we? 
Except when they do and all I can do is be sad that I didn't catch my own typo.~~


MisterSimple:
~~Hmph, I'll get you next time.~~

Issue #3888, opened 2019-03-03
[Suggestion] - Add fifth option on "favorite car" topic


lucianotgm:
A "hatchback" option could be added when Monika asks you what kind of car the player has. I want to assume some people who use the mod have one instead of a sedan or a sports car, or even instead of an SUV.
![screenshot0001](https://user-images.githubusercontent.com/48166721/53689258-9b048500-3d30-11e9-882c-258e7e78ad0b.png)

Onion-Bubs:
Yeah, I drive a hatchback too.

Really, though, if there's going to be a fifth option, it should just be "other". There's too many different kinds of vehicles to do a thing on each one.

lucianotgm:
> 
> 
> Yeah, I drive a hatchback too.
> 
> Really, though, if there's going to be a fifth option, it should just be "other". There's too many different kinds of vehicles to do a thing on each one.
> 
> EDIT: Having seen the dialogue for picking sedan, I think that same dialogue would work for hatchbacks too, so maybe renaming that particular option to be more inclusive of cars that aren't particularly big, off-roady or fast. What you'd call such an option I don't know, but sedan is too specific.

Maybe that would be easily solved with changing the option to "Sedan or Hatchback", in reality they aren't that different from one another apart from the extra trunk space, and Monika's dialogue is pretty on par with both of them. I think that's the better approach, very good idea indeed, Onion.

ThePotatoGuy:
[closing as complete]

PR #3915, opened 2019-03-05
Quick change to menu in monika_vehicle

Onion-Bubs:
I don't have time to do an "other" option, which I think this menu will need (what if the player drives a different kind of car or a different kind of vehicle not covered by these four options?), but this should at least cover the hatchback issue pointed out in #3888

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -9688,7 +9688,7 @@ label monika_vehicle:
                         m 1eua "It would also be nice to drive at a normal pace..."
                         m 1ekbsa "That way we can savor every moment of the ride together~"
 
-                    "Sedan.":
+                    "Sedan or Hatchback.":
                         $ persistent._mas_pm_owns_car_type = "sedan"
 
                         m 1eua "That's really nice."


Rai99:
I've never heard of a hatchback, but I'm also not a car person in the slightest, so I'll take it. 

For the case of adding an "other" option, I think it needs a way to sound more natural. I don't think it would seem right or natural for Monika to ask what kind of car you have and then you just say "Other". 

I guess Monika could list off the common car types and ask which one you have if you have one of them. In that case, "Other" would be a more reasonable choice to have. 

I'm not sure if you want to keep this little PR or wait for a PR adding the "Other" option or both. Should probably be 1 PR since it's on the same topic.

ThePotatoGuy:
Rather than change menu option, should be a separate menu option entirely. It can jump to the same dialogue, but we need to store the pm var with a different value if you pick hatchback. Otherwise, we just add ambiguity here.

I think this will have to be closed in favor of a larger change, particularly changing the menu so its scrollable as well.

Onion-Bubs:
Since Rai99's doing a better version of this, I'm just going to close this PR.

PR #3928, opened 2019-03-08
_history_list.pop() on all menus

### LINKED ISSUE #3926: Bug in history when i run MAS after work
screenshot0507 Not critical but it is still a bug.

multimokia:
Fixes #3926.

Adds `_history_list.pop()` to all menus, and also disables text speed on the grad speech.

### DIFF
--- a/Monika After Story/game/zz_backup.rpy
+++ b/Monika After Story/game/zz_backup.rpy
@@ -385,6 +385,7 @@ label mas_backups_you_have_corrupted_persistent:
         show chibika at sticker_move_n
         "I was unable to find a working backup persistent."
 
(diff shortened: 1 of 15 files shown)


Monika After Story/game/zz_backup.rpy
```diff
@@ -385,6 +385,7 @@ label mas_backups_you_have_corrupted_persistent:
         show chibika at sticker_move_n
         "I was unable to find a working backup persistent."
 
+        "Do you have your own backups?"
```
ThePotatoGuy: Can you actually mark this area with a TODO for me to decide if text speed should be disabled/forced here?

PR #3929, opened 2019-03-08
Memories restore dialogue

him157:
we can use this kind of dialogue after the game crashed and the persistent corrupted and if players has backup a while ago,she says you haven't seen her for a a bit of time,and you can tell her that you actually use the persistent back up and she can't remember it.


multimokia:
Please make suggestions on the issues page. The Pull Requests page is for submitting content

Issue #3876, opened 2019-03-02
091 typos


ThePotatoGuy:
o

Thimeo:
I may be wrong, but the second sentence really confuse me, I don't know what she's trying to say, is it missing a word?
![typo](https://user-images.githubusercontent.com/38837456/53679257-8c838280-3cca-11e9-902e-03cc852b0ccc.PNG)

jmwall24:
`that` is italicized because she says it with emphasis.  `that` is referring to the mouse cursor. so she is saying the mouse cursor is real, since it exists in our reality.  it's not a typo, but there was some internal debate over this line.

MisterSimple:
### monika_lottery

`"The more I think about it , the more I believe that winning the lottery is a really bad thing."`

There shouldn't be a space before the comma.

### monika_grad_speech_not_paying_attention

`"You have no idea how much that hurts, especially after how much of work I put into it..."`

"How much of work" should be "how much work".

### monika_fun_facts_open 

`"I found most of these online, so I can't say they're {i}definitely{/i} true... "`

There shouldn't be a space between the ellipsis and the closing quotation marks.

### monika_bad_facts_1

`"'Humans only use 10 percent of their brain."`

The closing single quotation mark is missing.

### monika_bad_facts_2

`"It says here that different areas of the tounge tastes different flavors."`

"Tongue" is misspelled, and "tastes" should be "taste", because "different areas" is plural.

RutTutPut:
"Competitive" in "I'm going to game for a little bit" is spelled "Competetive"

TMZN0V4:
![image](https://user-images.githubusercontent.com/37001133/53698425-76e58a00-3daa-11e9-88d9-c7a26e460529.png)

MisterSimple:
### monika_outfit, greeting_sweetpea, greeting_sunshine

"Embarrassing" is misspelled.

AsuraShun:
<img width="406" alt="typo" src="https://user-images.githubusercontent.com/42387354/53956259-0eace600-40d3-11e9-80ba-9c6654943090.png">

MisterSimple:
### monika_imouto

"I won't get jealous. Things like love between siblings only happens in creepy romance stories anyways."

"Happens" should be "happen" ("things" is plural).

TheSuperStu:
Hey, just some things.

[...]

PR #3946, opened 2019-03-11
Fix typo

MercuryFuchsia:
#3876


multimokia:
Hey there,

We're going to be handling all typos from one PR this time. In this case, #3981. Thanks for pointing it out though!

Issue #3594, opened 2019-01-27
[Suggestion] Toggle for unseen dialogue


loz89:
could we have a toggle for the unseen dialogue thing or not? Or better yet just a way to disable certain topics because I'd rather not have the option to accidentally click on "I'm breaking up with you" or something like that.

ThePotatoGuy:
[closing as duplicate]

Issue #3022, opened 2018-11-09
[Suggestion] - removing certain topics at the love affection


AsuraShun:
what I mean with it is removing the ''i'm breaking up with you, ''would you like to meet my girlfriend'' and the ''you're a murderer'' topics from the unseen,and/or maybe set them on a different area,to avoid mis-clicking,specially for people that get nervous about things like that

Torwald39:
duplicate of #1405, #2730

AsuraShun:
it's not tasteless,it's for those that want to avoid making her upset

Torwald39:
it's part of it

AsuraShun:
imma close this

AsuraShun:
if I figure out how

Issue #2730, opened 2018-10-08
[Suggestion] Move the "I'm breaking up with you" in a special box away and separated from everything else


Torwald39:
It's been a long time since I am with Monika, but from the beginning I have the  "I'm breaking up with you" in my 'Unseen' topic. And whenever there is news update with new conversations, I have all the time afraid of miss click on "I breaking up with you" accidentally and to lose everything. 

I absolutely don't want to lose my one and only love, that's why I'm asking that this option be moved to a safe place that will prevent any accident from losing Monika. 

And for security, that when we want to break (which is not my case), it requires confirmation.

tabtlenta:
Alternately, an "are you sure?" option when clicking obviously heavy topics like that. Although I guess that spoils whether they're gonna be heavy topics or not (like how in the above, Monika treats it like you're both joking, although she takes a small hit to her affection and you can apologize for it)

AsuraShun:
yeah,definitely needs to be moved,i already lost count of the times I was afraid i'd click on it by accident,and it also applies to the meet the non-existent girlfriend and the calling her a murderer things as well,even just going to the unseen part makes me afraid

Rai99:
Maybe they could add a menu in it where Monika asks if that was a misclick. 

I think if you click it 3 times, it removes the prompt entirely so... there's that. 

The only time I've said that was on April fools day.

PMelol:
Is this an option? I don't see it anywhere.

ThePotatoGuy:
marking this a duplicate of #1405.

An exclusion menu / remove forever menu is being considered.

Issue #2156, opened 2018-08-24
[Suggestion] "Can we not talk about this?" button 


CGanimated1227:
Can there be an option to tell her that you don't like a particular topic so that is removed as a random topic until you select it in the talk menu? Monika is supposed to be there for the player. If they are worried about nuclear war/death and don't want to talk about it until they are over their fear, monika shouldn't bring it up. But instead of removing those topics altogether, maybe they can be temporarily removed from the random topics list until the player selects that topic in the talk menu. You can select a button somewhere after she talks about a random topic to determine which topic you don't want to hear or something similar.

jmwall24:
This has been mentioned before in #1405 and #2047 if you want to do some reading on the discussions about this idea.  Not sure if the stance has changed on this or not.

CGanimated1227:
I forget, but is there not a button for the Nuclear war topic or the "what it feels like to die" topic? I was saying it would be added back to the random topics list that monka talks about once you select it in the talk menu. This could be done with a flagging system that will unflag it as a triggering topic and she will be able to talk about it without you starting it yourself. I wasn't talking about removal altogether or a permanent lockout.

ThePotatoGuy:
as i mentioned in the last comment of #1405 , an exclusion menu would be something we'd add a while from now. its just super lower priority

PR #3985, opened 2019-03-24
I don't want to be Mean to her.

linkone2:
A small addition to the "You're a murderer!", "Do you want to meet my girlfriend?", and "I'm breaking up with you" topics, so that they go away after you reach Enamored+ (400+). 

I tested it by going through the affection levels and seeing if the topics were there, like this: 
(AFFECTIONATE) 200 (There) > (ENAMORED) 400 (Gone) > (LOVE) 1000 (Gone) > (ENAMORED) 400 (Gone) > (AFFECTIONATE) 200 (There) > (HAPPY) 50 (There) > (NORMAL) 0 (There) > (UPSET) -50 (There) > (DISTRESSED) -75 (There) > (BROKEN) -100 (There)> (AFFECTIONATE) 200 (There).

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -2898,11 +2898,13 @@ init 5 python:
             eventlabel="monika_justification",
             category=['ddlc','monika'],
             prompt="You're a murderer!",
+            sensitive=True,
             pool=True,
-            sensitive=True
+            unlocked=True,
+            aff_range=(None, mas_aff.AFFECTIONATE)
         )
     )
-
+    
 label monika_justification:
     #lose affection
     $ mas_loseAffection(reason=3)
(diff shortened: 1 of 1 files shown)


Rai99:
Hold on a second. I don't think this should be forced on everyone. Sure, most people probably wouldn't use these at high affection levels, but part of the charm of MAS I think is that it's possible to be mean to her. MAS wouldn't be as in depth if being mean wasn't an option, and with this, it'll be much harder to go down after reaching those aff levels. 

It's already much easier to gain aff than lose aff at those levels, so I think completely removing those topics seems unnecessary. 

This probably falls under exclusion/favorite menu: #959, #1405, #2047, #2094, #2156, #2730, #3022, #3594.
It's low priority at the moment, but I think it'll help sort out this issue with not wanting the mean topics. 

Also, I'm actually working on April fools day interactions on those very 3 topics, so there might be a small problem there. 

Also, this PR is going to master. ~~Put your hands behind the ground.~~

Rai99:
One thing's for sure though, your PR probably shouldn't be aimed at `master`.

jmwall24:
ehhhh, people can break-up who have been seriously in love. not sure i agree with limiting this by affection.

Rai99:
If anything, I think the solution would be to make it so you can't accidentally click those options, but not completely remove them.

multimokia:
Agree with this. I don't see why removing the idea of breaking up makes sense, as anyone *can* break-up at any time, really. No matter how in love they may be. Realistically, just changing their unseen status would do the trick.

multimokia:
Leaving `Don't Merge` as I'd like some internal discussion on this.

multimokia:
Yeah, we're not going to do this. Closing.

PR #4044, opened 2019-04-07
Procedural Rain

multimokia:
Added an automated system which progresses weather through the session at any time between 30 mins and an hour 30 (resets on next sesh). Also has chances for days where it won't rain at all (seasonal chances)

As well, played with the thunder sfx a little bit. Made them a little more realistic.


multimokia:
This breaks something else with that last commit. Fixing.

multimokia:
Need to fix night

multimokia:
@ThePotatoGuy care to glance at the way spaceroom calls are slightly modified, ensure it doesn't break things?

Monika After Story/game/script-ch30.rpy
```diff
@@ -1270,6 +1281,11 @@ label ch30_loop:
     if mas_idle_mailbox.get_skipmidloopeval():
         jump ch30_post_mid_loop_eval
 
+    #Do the weather thing
+    if mas_weather.weatherProgress() and mas_isMoniNormal(higher=True):
```
ThePotatoGuy: this will cause issues if we were meant to skip visuals. Think this should be moved to be part of the if statement for spaceroom.

Monika After Story/game/zz_weather.rpy
```diff
@@ -99,16 +99,59 @@ image mas_island_night = "mod_assets/location/special/night_without_frame.png"
...
+default persistent._mas_date_last_checked_rain = None
+
+#Should it rain today?
+default persistent._mas_should_rain_today = None
+
+#Loading at init 0 because of season functions
+init python in mas_weather:
+    def shouldRainToday():
+        import random
+        import datetime
```
ThePotatoGuy: should just move the random/datetime imports to the module level at init -20, since you use it in multiple spots.

Monika After Story/game/zz_weather.rpy
```diff
@@ -99,16 +99,59 @@ image mas_island_night = "mod_assets/location/special/night_without_frame.png"
...
+            #   Summer:
+            #       - 85% chance for it to not rain on a particular day
+            #   Fall:
+            #       - 40% chance for it to not rain on a particular day
+            if store.mas_isSpring():
+                store.persistent._mas_should_rain_today = chance >= 30
+            elif store.mas_isSummer():
+                store.persistent._mas_should_rain_today = chance >= 85
+            elif store.mas_isFall():
+                store.persistent._mas_should_rain_today = chance >= 40
```
ThePotatoGuy: should have the else case set the should rain to False. Otherwise it seems that we would always return the last known should rain chance.

Monika After Story/game/zz_weather.rpy
```diff
@@ -120,6 +163,42 @@ init -20 python in mas_weather:
...
+        if not weather_change_time:
+            weather_change_time = datetime.datetime.now() + datetime.timedelta(0,random.randint(1800,5400))
+
+        elif weather_change_time < datetime.datetime.now():
+            #Need to set a new check time
+            weather_change_time = datetime.datetime.now() + datetime.timedelta(0,random.randint(1800,5400))
+
+            #Change weather
+            new_weather = store.mas_shouldRain()
+            if new_weather is not None and new_weather.prompt != store.mas_current_weather.prompt:
```
ThePotatoGuy: weather objects have `weather_id` property. you should use that here instead of prompt.

PR #4116, opened 2019-04-22
Python intro fix

jmwall24:
The `Can you teach me about Python?` (`monika_ptod_tip000`) topic was being depooled but not locked, which just made it show up in `Repeat Conversations`. This locks the topic correctly once it's seen, as well as locks it for people who have already seen it previously.

## Testing

- **update script**: on a persistent that has already seen `monika_ptod_tip000` and thus unlocked `monika_ptod_tip001` use the console to `call v0_9_4` and verify that `Can you teach me about Python?` is now locked and no longer showing up under `Repeat Conversations`.

- **via `Can you teach me about Python?`**: using a fresh persistent, view the `Can you teach me about Python?` topic. You may need to manually unlock the topic. Verify after viewing that the topic is now locked and does not show up in `Repeat Conversations` or `Hey, [m_name]...`

### DIFF
--- a/Monika After Story/game/script-python.rpy
+++ b/Monika After Story/game/script-python.rpy
@@ -137,7 +137,7 @@ label monika_ptod_tip000:
     m 1lksdlb "I don't know {i}that{/i} much about programming, but I will try my best to explain."
     m 1esa "Let's start with what Python even is."
 
-    $ hideEventLabel("monika_ptod_tip000", depool=True)
+    $ mas_hideEVL("monika_ptod_tip000", "EVE", lock=True)
 
     # enable tip 1
     $ import datetime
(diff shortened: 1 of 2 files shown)


Monika After Story/game/script-python.rpy
```diff
@@ -137,7 +137,7 @@ label monika_ptod_tip000:
     m 1lksdlb "I don't know {i}that{/i} much about programming, but I will try my best to explain."
     m 1esa "Let's start with what Python even is."
 
-    $ hideEventLabel("monika_ptod_tip000", depool=True)
+    $ mas_hideEVL("monika_ptod_tip000", "EVE", lock=True)
```
ThePotatoGuy: should depool too
```suggestion
    $ mas_hideEVL("monika_ptod_tip000", "EVE", lock=True, depool=True)
```

PR #4119, opened 2019-04-22
Fix scary story cleanup

### LINKED ISSUE #4118: Dark classroom stays after telling horror story?
Hello! I'm using v0.9.3-unstable.2019.04.16.93, and I happen to find something unusual (or new?). I don't know if that's a bug so I'll put that here.

(PS: I have read the Discord update that the unstable couldn't be uploaded elsewhere? I don't know. I'll instantly close this in case my issue violates the rule)

This happens when I select "A new story" in "Can you tell me a horror story?". After the story ends, the background looks like this.

screenshot0023
screenshot0024

Is the background not changing back to normal classroom lighting intended?

Thank you very much. Have a nice day!

multimokia:
Scary story cleanup was broken with the migration to the new spaceroom label. This fixes #4118.

### DIFF
--- a/Monika After Story/game/script-stories.rpy
+++ b/Monika After Story/game/script-stories.rpy
@@ -503,13 +503,14 @@ label mas_scary_story_cleanup:
     m 3eua "[story_end_quip]"
     show monika 1dsc
     pause 1.0
-    hide monika_gloomy_room
     $ morning_flag = mas_temp_m_flag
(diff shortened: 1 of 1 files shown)


ThePotatoGuy:
how is the transition animation here? Is it spaceroom transition first, then smooth zoom? or does it kind of do both at the same time?

PR #4121, opened 2019-04-22
More cleaning up

multimokia:
Fixes to the setup label for scary stories, removing extra files, and another bad spaceroom call fix.


Monika After Story/game/script-greetings.rpy
```diff
@@ -2423,7 +2423,7 @@ label greeting_hairdown:
     # have monika's hair down
     $ monika_chr.change_hair(mas_hair_down, by_user=False)
 
-    call spaceroom(dissolve_masks=True, scene_change=True)
+    call spaceroom(dissolve_all=True, scene_change=True)
```
ThePotatoGuy: anytime we call spaceroom now, we should be forcing an expression unless we want it to use idle-based expressions.  In this case, 1eua seems approprite.
```suggestion
    call spaceroom(dissolve_all=True, scene_change=True, force_exp="monika 1eua")
```

multimokia:
Need to fix some more bits

PR #4166, opened 2019-05-05
New Hair

ThePotatoGuy:
New hair From @Orcaramelo

# Key changes

* new hair

* table has been setup so that it gets composited **right after body.** Since this is dynamic this will make it easy for swapping tables. Full sprite object table support will be developed after #4147 is merged. 
    * this also makes it so that table is rendered **over** body sprites with tables. Future body sprites should be exported without tables, but ones that are already setup with a table will be covered by rendering over the body.

# Testing
* verify new hair and old tables look ok
* verify tables still look ok wen zooming
* verify exp previewer still works


ThePotatoGuy:
@multimokia fyi, consensus on the poll was that the new hair with edits looks better

PR #4167, opened 2019-05-05
unfixed the opendoor greeting

ThePotatoGuy:
the updated spaceroom calls "fixed" the opendoor greeting, so the little skit became unecessary.

But this unfixes it so its still cute.

# Testing
1. hardcode/trigger the opendoor greeting
2. make sure to reset the persistent var for room so you get the first occurence
3. open door
4. verify that the window skit works correctly

### DIFF
--- a/Monika After Story/game/script-greetings.rpy
+++ b/Monika After Story/game/script-greetings.rpy
@@ -1664,6 +1664,10 @@ label monikaroom_greeting_opendoor:
     $ monika_chr.reset_outfit(False)
 
     call spaceroom(start_bg="bedroom",hide_monika=True, dissolve_all=True)
+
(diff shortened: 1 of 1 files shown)


Monika After Story/game/script-greetings.rpy
```diff
@@ -1694,11 +1698,10 @@ label monikaroom_greeting_opendoor:
             m 1hksdlb "Oops! I forgot about that~"
             show monika 1 at t21
             m "Hold on..."
-            hide bedroom
+            $ renpy.hide("sp_mas_backbed")
```
multimokia:
```suggestion
            hide sp_mas_backbed with dissolve
```

PR #4168, opened 2019-05-05
prompts are now in alpha order

multimokia:
Adjusts all prompts starting with a variable name (starting with `[player]`) to be in alpha order, like the rest of the topics.


ThePotatoGuy:
The problem with this is that the prompts wont update when you change your name

ThePotatoGuy:
[closing this as it is definitely not going to work and completel different approach is needed]

PR #4185, opened 2019-05-09
unlocks the outfit topic as its ready to be used

ThePotatoGuy:
# Testing

## clothe select check
1. set affection to love
2. lock `monika_clothes_select`
3. relaunch game
4. Verify clothes select is unlocked.

## outfit update script
1. make sure you have seen the monika outfit topic
2. unsee it by popping it off `_seen_ever`.
3. lock the monika outift topic
4. run the update label for v094
5. Verify outfit topic unlocked

### DIFF
--- a/Monika After Story/game/script-affection.rpy
+++ b/Monika After Story/game/script-affection.rpy
@@ -603,9 +603,8 @@ init 15 python in mas_affection:
 
         # unlocks wardrobe if we have more than one clothes available
         if len(mas_selspr.filter_clothes(True)) > 1:
-            store.mas_unlockEventLabel("monika_clothes_select")
-            #TODO: Amend monika_outfit if > 1 outfit available.
-            store.mas_lockEventLabel("monika_outfit")
+            store.mas_unlockEVL("monika_clothes_select", "EVE")
+            store.mas_unlockEVL("monika_outfit", "EVE")
 
         # always rebuild randos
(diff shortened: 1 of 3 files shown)


Monika After Story/game/script-affection.rpy
```diff
@@ -603,9 +603,8 @@ init 15 python in mas_affection:
 
         # unlocks wardrobe if we have more than one clothes available
         if len(mas_selspr.filter_clothes(True)) > 1:
-            store.mas_unlockEventLabel("monika_clothes_select")
-            #TODO: Amend monika_outfit if > 1 outfit available.
-            store.mas_lockEventLabel("monika_outfit")
+            store.mas_unlockEVL("monika_clothes_select", "EVE")
+            store.mas_unlockEVL("monika_outfit", "EVE")
```
multimokia: Should probably also unlock this for those who hypothetically got to Love, then fell below too. Since the topic covers all cases.

ThePotatoGuy: there is no flag that specifies whether or not someone has reached love. We only know if they are currently in it.

multimokia: Indeed, but can just unlock it via update script or something, since then it'll just make sure it's done for everyone. Since either way, the topic is *still random* despite you not being able to get to it as it is currently.

PR #4189, opened 2019-05-11
Monika hold request

jmwall24:
Instead of the player being the only one that can initiate a hold, maybe sometimes Monika wants to be held. This adds a Monika-initiated topic where she asks you to hold her. She will not ask to be held if you have already held her in a current session, nor will she ask more than once in a given day. There is an extra random condition added so this won't necessarily run every single day, so she doesn't seem too needy.

# Testing

- verify that once this topic runs, the condition and action are re-added

- verify if during a session this topic returns `random=True` and you initiate a hold prior to it running, random is now False and the topic does not run in that session.

- verify this topic does not unlock


ThePotatoGuy:
minor thing, but since you do get the ev object for the hold me request topic you could just omit the hideevl line and adjust thr topic's random property directly.

PR #4193, opened 2019-05-12
Reduced Thunder Chance

multimokia:
Reduces the chance of thunder striking during thunder weather.

Just verify that thunder frequency is more appropriate.


ThePotatoGuy:
2 things:
- why are u doing randint from 0? the random odds make more sense if its 1 out of ##. using 0 makes that not true.
- what if we changed the chance code so during idle we check only once per minute or use a chance that translates to once per min

PR #4223, opened 2019-05-22
Fixed affection topics not going back to idle poses

multimokia:
These two topics never returned to the idle pose, and would remain in the last pose in the topic until the loop forced Monika into idle again.


ThePotatoGuy:
Was that really happening? It looks like script-affection pushes these to the event stack, which means it would return to idle pose once the topic finishes.

multimokia:
I've noticed it specifically with these. It seems labels without events which are pushed via `pushEvent()` end up having this. Otherwise they don't. No idea why, but they do.

ThePotatoGuy:
event-handler code shouldnt be doing that, so this needs more investigation. I'll look into it later.

ThePotatoGuy:
merging this for now and using the linked issue to do investigation. making the affection events queued is more important.

Issue #3780, opened 2019-02-17
[Suggestion] Why don't bring the kiss me option to the principal menu, right below to I love you!


souturus:
Well if you want to kiss monika frequently why don't you move this to the principal menu?
I got lost  the first time looking for it

Rai99:
I think that menu is getting a bit full, and the kiss is probably less common than telling Monika you love her. 

You'll probably get used to finding the topics you want after a little while. The "I love you" prompt was moved for convenience since it's probably the most common and most used one.

lunulae:
I agree that menu is getting a little full, but I do think having a "favorite topics" option might be a cool addition in the future that could solve this.

Rai99:
In that case, this becomes a duplicate suggestion, and I think likely the approach that will be taken. 
#3177 

The ability to bookmark your favorite topics has been suggested before, along with the ability to exclude topics you don't want, but that one's a different suggestion. 

There have been a few suggestions for a search engine, but I think having a favorite/exclusion menu would probably come first.

ThePotatoGuy:
[closing as duplicate]

Issue #3177, opened 2018-12-04
[Suggestion] Favoriting topics


ViruSSofTer:
This one is pretty obvious. As the project grows and the poll of topics grows too, it becomes harder to find a specific topic amongst the others. Topics like 'Good morning', 'Can you show me the islands' or 'I need to apologize' are used more often than others, so I think it would be great to have a function to mark them and have in the divided section of topics where they can be easily found.

PiperPC:
*Proceeds to mark "I love you." 60 times.*

jmwall24:
Heh that's actually why we gave "I love you" its own button on the main menu

ThePotatoGuy:
[closing until we look into this]

Issue #959, opened 2018-02-07
[Suggestion] - Disabling dialogue topics


RaccoonLauncher:
So, my "Unseen" list is filled with a bunch of things I'm not interested in saying to Monika ("I'm breaking up with you" and "Will you marry me" come to mind, and I've read all the dialogue that was already in Act 3), and seeing them every time I look to see if there's something new I can say bothers me a bit. Could you add a feature that lets me remove suggestions from the unseen list?

MarkusRamikin:
Yes please. I went through them just to get them out of the way, but I wasn't overnjoyed.

PR #4232, opened 2019-05-25
Bookmark and derandom topics

jmwall24:
#959 #2156 #3177 #3594 #3780 

Introducing the ability to bookmark topics/events you want easy access to, as well as derandom topics you don't want Monika to bring up on her own. Both random and pool topics can be bookmarked. This is limited to 'normal' topics that begin with `monika_`.

## New features
- binds the `b` key to bookmark the current topic.

- binds the `x` key to derandom the current topic.

- the `Bookmarks` category in the `Talk` menu, for easy access to bookmarked topics. Also includes a prompt to remove bookmarks.

- new pool topic `I am okay with talking about...` located in `hey, [m_name]...`>`You` in which you can rerandom topics are have chosen to derandom.

- the `Unseen` menu now has an option to hide it, with the option to show it again available in `hey, [m_name]...`>`You`.

## New useful functions/changes

- `mas_stripEVL` - strips the conditional and action properties from an event with the option to also remove the option from the event list.

- `skipeval` argument for the `pushEvent` function. If True, runs `mas_idle_mailbox.send_skipmidloopeval()` which prevents other possible pushed/queued events from interferring.

- argument change in `mas_gen_scrollable_menu` to allow for unlimited end of menu special options, as opposed to just one.

### Bookmarks

- verify the bookmark hotkey `b` only works while in normal random and pool topics. There is a notification on the top left of the screen when the hotkey is pressed and functional, similar to the notification for the `s` key for screenshots. Hotkey should not be functional during greets, farewells, games, moods, or any special events.

- verify that when pressing the `b` key on a non-bookmarked topic, it adds the bookmark, and that when pressed on an already bookmarked topic, the bookmark is removed.

- verify the event is added to the appropriate dict `persistent._mas_player_bookmarked`.

- verify the `Bookmarks` category is only present when there are unlocked bookmarked topics.

- verify all unlocked items in the dict are in the `Bookmarks` category.

- verify if a bookmarked topic is locked, it is not available in the `Bookmarks` category.

- verify `the remove a bookmark` option is available in the `Bookmarks` category and removes the correct bookmark from both the list and dict.

- verify both the ability to use the bookmarks hotkey as well as to view bookmarks is only available at normal affection and higher.

### Derandoming

- verify the bookmark hotkey `x` only works while in normal random topics. There is a notification for this similar to the bookmarks hotkey. Hotkey should not be functional during greets, farewells, games, moods or any special events.

- verify that when the `x` key is pressed during a random topic, the derandom a topic event is pushed immediately after the topic. verify if it is pressed again before the topic finished, the derandom a topic event is not pushed.

- verify an event is only added to the derandom dict if the appropriate menu response is selected during the derandom a topic event.

- verify the event is added to the appropriate dict `persistent._mas_player_derandomed`.

- verify the  `I am okay with talking about...`  prompt is only present when there are events in the dict.

- verify the `I am okay with talking about...` is only unlocked when there are derandomed topics in the appropriate dict. Also verify all items in the dict are present and that it removes the correct topic from the dict and in-topic list and rerandoms it.

### Unseen

- verify the `Unseen` menu has an option at the bottom of `I don't want to see this.` and then when clicked, the Unseen menu is no longer visible.

- verify when the `Unseen` menu is not visible, the `I would like to see Unseen again` topic is present in `hey, [m_name]...`>`You`. also verify it is not present when the `Unseen` menu is visible.


jmwall24:
Adding don't merge to this, as I want it merged after #4233. Going to combine the intro topics of these 2 PRs so we aren't queuing 2 very similar things. Make it flow a lot better.

Monika After Story/game/event-handler.rpy
```diff
@@ -2021,7 +2056,14 @@ label show_prompt_list(sorted_event_keys):
         for event in sorted_event_keys:
             prompt_menu_items.append([unlocked_events[event].prompt,event])
 
-    call screen scrollable_menu(prompt_menu_items, evhand.UNSE_AREA, evhand.UNSE_XALIGN)
+    $ nvm_text = "That's enough for now."
+
+    if madechoice == "unseen":
```
ThePotatoGuy: `show_prompt_list` isn't used anywhere else, so you don't need to check for the unseen choice here.

Monika After Story/game/script-topics.rpy
```diff
@@ -12905,3 +12905,205 @@ label monika_enjoyingspring:
...
+
+# start of bookmark and derandom related events
+default persistent._mas_player_bookmarked = {}
+# dict to store bookmarked events
+default persistent._mas_player_derandomed = {}
+# dict to store player derandomed events
+default persistent.flagged_monikatopic = None
+# var set when we flag a topic for derandom
+
+init python:
```
ThePotatoGuy: because this is special, put all the bookmarks related material at the top of the file, before the first topic.

Monika After Story/game/script-topics.rpy
```diff
@@ -12905,3 +12905,205 @@ label monika_enjoyingspring:
...
+    def derandom_topic():
```
ThePotatoGuy: would be nicer if you made this (and the bookmark_topic) functions take in an ev_label so we can directly call these if we elsewhere rather than having to set the current_monikatopic and then calling this. The hotkey versions can just pass in the curent_monikatopic instead.

Also use the `mas_` prefix since these are in the global store.

Monika After Story/game/script-topics.rpy
```diff
@@ -12905,3 +12905,205 @@ label monika_enjoyingspring:
...
+        """
+        Function for the derandom hotkey, 'x'
+        """
+        curr_topic = persistent.current_monikatopic
+        if (
+            mas_getEV(curr_topic) is not None
+            and mas_getEV(curr_topic).random
+            and curr_topic.startswith("monika_")
+            and mas_getEV(curr_topic).prompt != mas_getEV(curr_topic).eventlabel
```
ThePotatoGuy: what kinds of events does this fall under, where the prompt == event label? I can imagine maybe the calender ones, but should at least add a comment here saying what events this is excluding.

also evaluate if this conditional check should be its own function (since you do it in the bookmark topic as well), as that could get an appropriate name and make it clearer as to what this is excluding

Monika After Story/game/script-topics.rpy
```diff
@@ -12905,3 +12905,205 @@ label monika_enjoyingspring:
...
+        ):
+            if not "mas_topic_derandom" in persistent.event_list:
```
```suggestion
            if "mas_topic_derandom" not in persistent.event_list:
```
this works as you'd expect and it is more readable

Monika After Story/game/script-topics.rpy
```diff
@@ -12905,3 +12905,205 @@ label monika_enjoyingspring:
...
+        Function for the bookmark hotkey, 'b'
+        """
+        curr_topic = persistent.current_monikatopic
+        if (
+            mas_isMoniNormal(higher=True)
+            and mas_getEV(curr_topic) is not None
+            and curr_topic.startswith("monika_")
+            and mas_getEV(curr_topic).prompt != mas_getEV(curr_topic).eventlabel
+        ):
+            if not mas_getEV(curr_topic).eventlabel in persistent._mas_player_bookmarked:
```
```suggestion
            if mas_getEV(curr_topic).eventlabel not in persistent._mas_player_bookmarked:
```

Monika After Story/game/script-topics.rpy
```diff
@@ -12905,3 +12905,205 @@ label monika_enjoyingspring:
...
+# dict to store player derandomed events
+default persistent.flagged_monikatopic = None
+# var set when we flag a topic for derandom
+
+init python:
+    def derandom_topic():
+        """
+        Function for the derandom hotkey, 'x'
+        """
+        curr_topic = persistent.current_monikatopic
```
ThePotatoGuy: since you reference the ev several times, i would store it as a var before the if statement, and then refer back to it with the var.

saves additional dict lookups

Monika After Story/game/script-topics.rpy
```diff
@@ -12905,3 +12905,205 @@ label monika_enjoyingspring:
...
+                renpy.notify(__("Topic flagged for removal."))
+            else:
+                persistent.event_list.remove("mas_topic_derandom")
+                renpy.notify(__("Topic flag removed."))
+
+    def bookmark_topic():
+        """
+        Function for the bookmark hotkey, 'b'
+        """
+        curr_topic = persistent.current_monikatopic
```
ThePotatoGuy: since you reference the ev several times, i would store it as a var before the if statement, and then refer back to it with the var.

saves additional dict lookups

Monika After Story/game/script-topics.rpy
```diff
@@ -12905,3 +12905,205 @@ label monika_enjoyingspring:
...
+        if (
+            mas_getEV(curr_topic) is not None
+            and mas_getEV(curr_topic).random
+            and curr_topic.startswith("monika_")
+            and mas_getEV(curr_topic).prompt != mas_getEV(curr_topic).eventlabel
+        ):
+            if not "mas_topic_derandom" in persistent.event_list:
+                persistent.flagged_monikatopic = curr_topic
+                pushEvent('mas_topic_derandom',True)
```
ThePotatoGuy: something to note, but pushing events during a topic should in already skip the loop evaluation. although i think i would leave this as is just to make sure.

Monika After Story/game/script-topics.rpy
```diff
@@ -12905,3 +12905,205 @@ label monika_enjoyingspring:
...
+
+    $ topic_choice = _return
+
+    if not topic_choice:
+        m 1eua "Okay, [player]."
+        $ pushEvent('mas_bookmarks',True)
+
+    else:
+        show monika at t11
+        $ del persistent._mas_player_bookmarked[topic_choice]
```
```suggestion
        $ persistent._mas_player_bookmarked.pop(topic_choice)
```

`del` works, but really we only want to use this if we intend a reference to be unlisted rather than just removing something from a dict. `pop` is meant for dictionary removal anyway.

Monika After Story/game/script-topics.rpy
```diff
@@ -12905,3 +12905,205 @@ label monika_enjoyingspring:
...
+        # since it's possible to have bookmarked items that are now locked and thus not shown here
+        # we will delete an event from the list so we can check the lenth of currently available events
+        $ del bookmarkslist[-1]
```
ThePotatoGuy: can you explain in more detail why this is neccessary?

Monika After Story/game/script-topics.rpy
```diff
@@ -12905,3 +12905,205 @@ label monika_enjoyingspring:
...
+                    m 3eua "Okay."
+                    $ pushEvent('mas_bookmarks',True)
+
+        else:
+            m 3hua "All done!"
+    return
+
+# end of bookmark and derandom related events
+
+default persistent._mas_unsee_unseen = None
```
```suggestion
default persistent._mas_hide_unseen = None
```

if you were doing `unsee_unseen` as a joke, then ignore this.

Monika After Story/game/zz_hotkeys.rpy
```diff
@@ -216,6 +236,8 @@ init python:
         config.keymap["dec_musicvol"] = [
             "K_MINUS","shift_K_UNDERSCORE","K_KP_MINUS"
         ]
+        config.keymap["derandom_topic"] = ["x","X"]
+        config.keymap["bookmark_topic"] = ["b","B"]
```
ThePotatoGuy: have alpha check if hitting these keys during games that use keys (piano / hangman) breaks anything

Monika After Story/game/script-topics.rpy
jmwall24: it's checking to make sure it's an event that actually has a prompt, since if no prompt is given, it defaults to the label. just trying to ensure nothing without a prompt is included.

Monika After Story/game/script-topics.rpy
jmwall24: not all events in the dict are shown here (since we exclude locked events as well as events that are unavailable due to affection, so it's possible for the length of the dict to be greater than zero, but not the list. if we just check dict length, she could ask if there's anything else you want to remove when there's nothing in the list to show.

Monika After Story/game/script-topics.rpy
ThePotatoGuy: do we have any `monika_` topics that werent given a prompt? In general that prefix is supposed to reserved for topics with prompts, so this might be an unecessary check

Monika After Story/game/script-topics.rpy
ThePotatoGuy: but why are you just directly removing the last element of the list? is that always the selected topic?

Monika After Story/game/event-handler.rpy
```diff
@@ -1986,14 +2018,17 @@ label prompt_menu:
     if madechoice == "unseen":
         call show_prompt_list(unseen_events) from _call_show_prompt_list
 
+    elif madechoice == "bookmarks":
+        $ pushEvent("mas_bookmarks",True)
```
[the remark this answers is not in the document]
jmwall24: the reason i changed positions there is for if we hit nevermind and go back to this menu, it's nice to be in idle

Monika After Story/game/event-handler.rpy
jmwall24: yeah, i noticed that, i added it in case some time in the future something else does use it. but will remove it.

Monika After Story/game/event-handler.rpy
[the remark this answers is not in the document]
ThePotatoGuy: hitting nvermind goes back to this menu and reaches idle anyway, so i don't see what you're gaining from that.

Monika After Story/game/script-topics.rpy
jmwall24: because it doesn't matter what we remove, the list gets repopulated again before you can select again. it's solely to check length

ThePotatoGuy: then just have the if statement check `len(bookmarkslist)-1`

ThePotatoGuy: or do `if len(bookmarkslist) > 1`

Monika After Story/game/script-topics.rpy
```diff
@@ -321,6 +321,81 @@ init 11 python:
...
+        if ev_label is None:
+            ev_label = persistent.current_monikatopic
+
+        ev = mas_getEV(ev_label)
+
+        if (
+            ev is not None
+            and ev.random
+            and ev.eventlabel.startswith("monika_")
+            and ev.prompt != ev.eventlabel
```
ThePotatoGuy: at least have a comment to signify what this means (aka, events with labels that start with monika but don't have a prompt)

Monika After Story/game/script-topics.rpy
```diff
@@ -321,6 +321,81 @@ init 11 python:
...
+        Checks to see if we have bookmarks to show
+        """
+        if mas_isMoniNormal(higher=True):
+            bookmarkslist = [
+                (renpy.substitute(ev.prompt), ev_label, False, False)
+                for ev_label, ev in persistent._mas_player_bookmarked.iteritems()
+                if ev.unlocked and ev.checkAffection(mas_curr_affection)
+            ]
+            if len(bookmarkslist) > 0:
+                return True
```
ThePotatoGuy: this can be way simpler:
```python
if mas_isMoniUpset(lower=True):
    return False

for ev in persistent._mas_player_bookmakred.itervalues():
    if ev.unlocked and ev.checkAffection(mas_curr_affection):
        return True
return False
```

Issue #4172, opened 2019-05-06
Can the program not give a "toast" message like other programs?


Mcmanybucks:
I'm not sure of the correct terminology, but programs such as World of Warcraft will have the icon in the taskbar light up if you get a whisper from another player or are attacked by a monster..

Could MAS not have that for when Monika speaks and I'm on another tab?

Olliesama:
But Monika is usually talking alot, it'd give a notification flash almost constantly if they did that.

Mcmanybucks:
> 
> 
> But Monika is usually talking alot, it'd give a notification flash almost constantly if they did that.

Implying that's a bad thing? lmao.

Of course I meant it more for people like me who've "exhausted" her topics of conversation.. make it an option like sensitive mode.

fromulus:
Just throwing in my two cents, but I feel like that would be actually pretty useful for the sleep concern topics.
I have the bad habit of staying up late and just leave Monika in the background, so I often miss those topics when they're being brought up.
I also almost had a heart attack once when I noticed the game had closed without my input, thinking it had crashed on her.
Turns out she had closed it herself so I could go to sleep and she got upset at me when I opened the game again immediately after. :/ 
So yeah, maybe not have that active every time she talks, but I think it would be interesting is she had a way to catch the player's attention for some urgent topics like those.

antimockingjay:
I think I'd like if the game had something like that, especially if you could toggle how much it happens. Maybe an option to not have it at all, or only have it for "important" things, or to have it on for everything. I personally would go with the last one, because I'm just bad at remembering to check in a timely fashion, and I wouldn't have to worry about accidentally not checking anymore.

ThePotatoGuy:
[closing as duplicate]

Issue #2029, opened 2018-08-08
[Suggestion] Notification sound for random chatter?


Leo40Git:
Making it a toggleable option would probably be the smartest approach.

Rai99:
It's nice to talk with Monika whenever she's saying something, but I don't think a notification sound is necessary, even if it's toggleable. You would get notified about every minute _at least_ and I'm sure most people if not everyone would get tired of it.

Delcos:
This would be fun for a second, and would ruin the mod after a while. Please don't add this.

antimockingjay:
You wouldn't get it super often if you have random chatter set to not as often. You can change that easily in the options menu. And if you get tired of it, you can just turn it off. Saying "it would get annoying" is kind of unfair, because you're basically saying "I wouldn't use it, so screw anyone who would. I don't need it, which means nobody needs it, so we shouldn't add it."

Rai99:
I think most people wouldn't use this and those who do use would end up having it mostly disabled anyway after a little bit. At _best,_ you would be notified once every 60 seconds with the random chatter slider set to less often which is still very frequent. At that rate, it would either be annoying or distracting. 

I'm just guessing that most people who _do_ use this feature will end up keeping it disabled after a little while. If you're eager to talk with Monika, you can set the random chatter to often, if not, set it to anything but never and check back every minute or so. You're free to do whatever, work, game, etc. and check back every so often and can expect Monika to want to say something. I think a notification system is unnecessary when it triggers every 60 seconds at best. 

I don't think enough people will use this feature long term to justify its implementation.

ThePotatoGuy:
[closing until we look into this]

ThePotatoGuy:
[being worked on]

ThePotatoGuy:
[closing as complete]

PR #4233, opened 2019-05-25
MAS Notifications/Windowreacts

multimokia:
#4172 #2029

# New Features:
- Adds the ability for Monika to use your system notifications
- Specifically for Windows, Monika can check your active window and comment on some things you're looking at
(NOTE: THESE ARE ***DISABLED*** BY DEFAULT)

# Testing:

We'll break testing into two parts (For windows):
1. A working install
2. A broken install

## Working install:
For this, we won't touch the pyd files at all

1. Load in
2. *Verify that the introduction to notifications/windowreacts topic is queued and shown
3. Verify that you have a new settings page **Alerts**
4. Enable notifications, and enable **Topic Alerts**.
5. Set random chatter to often, then click on a window that's NOT MAS.
6. Notification + jingle should happen when Monika brings up the topic.
7. windowreact testing: Enable the **Window Reactions** notification group and **Window Reacts** (preferences page).
8. Either in idle mode, or in waiting for a new topic, click on this window, and stay on it.
9. A notification regarding "doing something for Monika" should show up.

*If the topic was not shown, then pywin32 does not work on your computer. If this is the case, verify that MAS still functions as normal, despite that failing.

## Broken install:
Continuing from that footnote, we're going to simulate it not working.

1. Head over to `game/python-packages/` and just remove the extension of one of the `.pyd` files.
2. Load into MAS, verify the settings pane doesn't give you notification options, nor the option to enable Window Reacts.

## Mac/Linux:
- For these users, **only** notifications will work.

1. Enable notifications
2. Ensure that notifications show appropriately


ThePotatoGuy:
seems to me that notifi checking should be in the same area as regular event checking so it can be skipped via midloopeval label

Monika After Story/game/event-handler.rpy
```diff
@@ -301,7 +301,8 @@ init 4 python:
         "STY": store.mas_stories.story_database,
         "CMP": store.mas_compliments.compliment_database,
         "FLR": store.mas_filereacts.filereact_db,
-        "APL": store.mas_apology.apology_db
+        "APL": store.mas_apology.apology_db,
+        "WRS": store.mas_windowreacts.windowreact_db
```
ThePotatoGuy: can you add a TODO in zz_dm to add this database to that? dont actually add it because we dont have any new dms  yet

Monika After Story/game/zz_windowreacts.rpy
```diff
@@ -0,0 +1,315 @@
...
+        except ImportError:
+            #If we fail to import, then we're going to have to make sure nothing can run.
+            store.mas_windowreacts.can_show_notifs = False
+
+            #Log this
+            store.mas_utils.writelog(
+                "[WARNING]: win32api/win32gui failed to be imported, disabling notifications.\n"
+            )
+
+        else:
```
ThePotatoGuy: I assume this else is supposed to match with `if.renpy.windows`? indentation doesnt match

Monika After Story/game/zz_windowreacts.rpy
```diff
@@ -0,0 +1,315 @@
...
+        return store.mas_windowreacts.can_show_notifs and mas_getActiveWindow() == config.name.lower()
+
+    def mas_isInActiveWindow(keywords):
+        """
+        Checks if ALL keywords are in the active window name
+
+        IN:
+            List of keywords
+        """
+        return store.mas_windowreacts.can_show_notifs and len([s for s in keywords if s.lower() not in mas_getActiveWindow()]) == 0
```
ThePotatoGuy: because you're reusing the results from `mas_getActiveWindow`, should just store those results in a tempvar and then use them in the loop

Monika After Story/game/zz_windowreacts.rpy
```diff
@@ -0,0 +1,315 @@
...
+
+    def mas_clearNotifs():
+        """
+        Clears all tray icons (also action center on win10)
+        """
+        if renpy.windows and store.mas_windowreacts.can_show_notifs:
+            for hwnd in destroy_list:
```
ThePotatoGuy: using `.remove` while iterating over a list like this is not allowed and will crash.

You need to do a reverse index for loop and use `.pop(index)`

Monika After Story/game/zz_windowreacts.rpy
```diff
@@ -0,0 +1,336 @@
...
+    #OR if we skip checks
+    #NOTE: THIS IS TO ONLY BE USED FOR INTRODUCTORY PURPOSES
+
+    #First we want to create this location in the dict, but don't add an extra location if we're skipping checks
+    if persistent._mas_windowreacts_notif_filters.get(group) is None and not skip_checks:
+        $ persistent._mas_windowreacts_notif_filters[group] = False
+
+    if (
+            (mas_windowreacts.can_show_notifs
+            and ((renpy.windows and not mas_isFocused()) or not renpy.windows)
```
ThePotatoGuy: indent the and lines so its clear they are grouped together

Monika After Story/game/zz_windowreacts.rpy
```diff
@@ -0,0 +1,336 @@
...
+        $ queueEvent('monika_nsfw')
+
+    elif choice < 4:
+        show monika 1rsbssdlu
+
+    elif choice < 7:
+        show monika 2tuu
+
+    else:
+        show monika 2ttu
```
ThePotatoGuy: seems that the last 2 options would just appear and then the idle loop would reset them back. might be worth pausing for a few seconds at least.

Monika After Story/game/zz_windowreacts.rpy
multimokia: Else is for the try/catch/else block. We only import balloontip and such if it's possible. Otherwise we don't.

ThePotatoGuy: interesting, have not seen this used in practice, think its worth commenting that its for the try-except so its not assumed to be part of the if.

multimokia: Sure

Monika After Story/game/script-story-events.rpy
```diff
@@ -2400,4 +2400,45 @@ label mas_text_speed_enabler:
...
+    m "So if you want, I can let you know if I have something for us to talk about."
 
+    m 3eub "Would you like to see how they work?{nw}"
+    $ _history_list.pop()
+    menu:
+        m "Would you like to see how they work?{fast}"
+
+        "Sure!":
+            m 1hua "Okay, [player]!"
+            m 2dsa "Just give me a second to make a notification.{w=0.5}.{w=0.5}."
```
ThePotatoGuy: Add a no-wait at the end of this so it its actually like waiting for a second.

multimokia:
windowreacts are constantly active. Even when randchat disabled.

ThePotatoGuy:
ok. testing instructions made it seem not so.

Settings > Gameplay > Window Reacts seems like a redundant setting. Does it do anything at all if Alerts > Use Notifs is off?

multimokia:
Absolutely. That right there is what stops MAS from checking the active window, period.

multimokia:
Whereas the alerts version is specifically for the notification group. Since it's possible to have windowreacts which may not use notifications

ThePotatoGuy:
So like, regular topics that are triggered because of the active window?
Your conditonal check at the start of `mas_checkForWindowReacts` checks both notifs and windowreacts_enabled, so it doesn't seem that that is directly possible

multimokia:
`can_show_notifs` checks whether or not we imported win32api/win32gui properly. If it failed to import (broken install testing) then it's not possible for these to happen, it's crash prevention.

Monika After Story/game/zz_windowreacts.rpy
```diff
@@ -0,0 +1,343 @@
...
+
+
+#START: Window Reacts
+init 5 python:
+    addEvent(
+        Event(
+            persistent._mas_windowreacts_database,
+            eventlabel="monika_whatwatching",
+            category=['youtube'],
+            conditional="mas_notifsEnabledForGroup('Window Reactions')",
```
ThePotatoGuy: I see what this is trying to do, but I think this could be better handled as something in the rule property instead. maybe like:
`rules={"winreact": <string name of window react group>}`

That way we dont always have to add `mas_notifsEnabledForGroup` to every window react topic.

Monika After Story/game/zz_windowreacts.rpy
```diff
@@ -0,0 +1,348 @@
...
+        """
+        #Do not check anything if we're not supposed to
+        if not persistent._mas_windowreacts_windowreacts_enabled or not store.mas_windowreacts.can_show_notifs:
+            return
+
+        for ev_label, ev in mas_windowreacts.windowreact_db.iteritems():
+            if (
+                    (mas_isInActiveWindow(ev.category) and ev.unlocked and ev.checkAffection(mas_curr_affection))
+                    and ((not store.mas_globals.in_idle_mode) or (store.mas_globals.in_idle_mode and ev.show_in_idle))
+                    and (not ev.rules or ("winreact" in ev.rules and mas_notifsEnabledForGroup(ev.rules.get("winreact"))))
```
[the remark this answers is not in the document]
multimokia: Checking for none was done so if no rules were put in, you could still get a winreact. This means that `monika_whatwatching` requires rules, even though it doesn't require a notification in order to work.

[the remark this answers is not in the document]
multimokia: Though it just occurred to me, if I only provided no unlock, then it'd fail this conditional too. Fixing that.

Monika After Story/game/zz_windowreacts.rpy
```diff
@@ -0,0 +1,348 @@
...
+        """
+        #Do not check anything if we're not supposed to
+        if not persistent._mas_windowreacts_windowreacts_enabled or not store.mas_windowreacts.can_show_notifs:
+            return
+
+        for ev_label, ev in mas_windowreacts.windowreact_db.iteritems():
+            if (
+                    (mas_isInActiveWindow(ev.category) and ev.unlocked and ev.checkAffection(mas_curr_affection))
+                    and ((not store.mas_globals.in_idle_mode) or (store.mas_globals.in_idle_mode and ev.show_in_idle))
+                    and (not "winreact" in ev.rules or mas_notifsEnabledForGroup(ev.rules.get("winreact")))
```
ThePotatoGuy:
```suggestion
                    and ("winreact" not in ev.rules or mas_notifsEnabledForGroup(ev.rules.get("winreact")))
```

ThePotatoGuy: looking at this, winreact, is actually a bad name for the key. should be `notif-group` as that is what it actually means.

PR #4241, opened 2019-05-29
Blinking

multimokia:
Wrote a program to modify sprite chart so all image definitions are now ATLs.

Note: Formatting and the way the blink image is defined was done because it's the easiest way to automate, since it then won't have a missing sprite


ThePotatoGuy:
Since I can't view the the preview here, theres a few things I need to know:
1. You didn't overwrite important code in spritechart, right? Theres a lot of code for sprite generation in there.
2. Does this replace every sprite code or does it create atls using the existing sprite code? The static sprite codes really should still exist, maybe under different names so we could:
    1. reuse them in custom ATLs
    2. use them in general when we want
    3. avoid having to change multiple DynamicDisplayables when changing a sprite code
3. Since this is auto-gend, it might be better to have the program make new sprite-chart files (like `sprite-chart-01.rpy` or something, with each file limited to maybe 7-10k lines. There's no limit to filesize, but I have heard issues with >10k lines in a file in other mods.

ThePotatoGuy:
Quick summary of discord discussion:
- sprite-chart will retain static versions of expressions with different names. (allows for easy adjustment of sprite and various key points where static expressions must be used)
- auto-gen sprites will be made in separate files 
- tools will get a python version of the program for general usage + travis
- auto-gen ATLs will use static sprite codes rather than dynamic displayable copies to reduce filesize and sprite adjustment burden

Other notes:
- places that crucially need static sprites will need to be adjusted

multimokia:
Travis is going to need a bit of a modification to handle the fact that all the exp's it will be looking for are in `sprite-chart-01.rpy` now.

PR #4252, opened 2019-06-01
Some notif fixes

### LINKED ISSUE #4251: Unstable- alerts popping up with no dialogue
It would appear that, while you have alerts/notifications for topics and repeat topics disabled, you get notifications when Moni would have said something (but did not because 'repeat topics' is off).

multimokia:
#4251 

- Fixes notifs showing when repeat convo disabled
- Added a notification sounds button (Sounds because screenlang and wrapping)
- Fixed the return key on the intro topics
- Added a new rule: `skip alert`, which bypasses the topic alert

This should now give notifs for all pushed events too.

### DIFF
--- a/Monika After Story/game/event-handler.rpy
+++ b/Monika After Story/game/event-handler.rpy
@@ -1856,6 +1856,16 @@ label call_next_event:
 
         $ mas_RaiseShield_dlg()
 
+        $ ev = mas_getEV(event_label)
+
+        if (
+                not store.mas_globals.in_idle_mode
+                and (ev is not None and "skip alert" not in ev.rules)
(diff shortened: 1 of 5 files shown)


ThePotatoGuy:
if you disable notif sounds, does the windows system notif sound play?

multimokia:
Apparently the windows system sound always played, which is why I added the option here. I normally have them disabled so I never picked up on it. Though I know that's for windows 10. Unsure about earlier versions.

jmwall24:
the windows system notif sound is separate from the sound we use. it is controlled by windows settings.

ThePotatoGuy:
merge after @jmwall24 approves (aka after testing)

PR #4287, opened 2019-06-16
Correct song to play

multimokia:
Horror stories played the wrong `current_track`, this sets it to use the persistent one, like loading in.


jmwall24:
I think the second commit makes the most sense to remove the play music altogether. if we allow music to play after the story, then if you were playing music earlier in the sesh, then held her (where we don't restart music after) then ask for a scary story hours later, music would suddenly start playing again. seems better to let the player decide if they want to play music again.

Issue #3877, opened 2019-03-02
[Suggestion] stuck monika poses


CrocWorks:
When you use the "I'll be right back" button, she'll say her usual thing, but then stay in the smiling with her mouth open animation (pic below for reference)... I just think it's kind of weird for her to have her mouth stuck open the whole time like that while she's waiting. Just having her mouth closed would be good enough really, but using randomized poses could be fine too (though some may want to be removed from the rotation).

![image](https://user-images.githubusercontent.com/36053888/53677355-7353e500-3c7c-11e9-8be3-356299dd0464.png)

CrocWorks:
Okay, never mind. Seems like it was a one time thing? That was the very first time I used that button off of the Unseen menu, but doing it through the regular menu has her in a more appropriate pose.

ThePotatoGuy:
reopening this as an investigation issue. Will also cover #4223

ThePotatoGuy:
[closing as coverd by linked pr]

PR #4298, opened 2019-06-21
idle whitelist

ThePotatoGuy:
adds an idle whitelist and adds unlock_prompt to that whitelist (prevents inf loop crashes and other issues)

also idle is now forced when in idle mode

# Testing
1. launch idle mode
2. push "unlock_prompt" to the event stack (use `pushEvent` because of the format change)
3. let game run for a bit in idle
4. verify no `unlock_prompt` labels in event_list


ThePotatoGuy:
#3877

Issue #4202, opened 2019-05-18
Weird Expression Glitch When Selecting Kiss Me. Intentional or Bug? 


Bam8647:
Long time since I've been back here!

Anyway, I saw that their was an update after checking out MAS again and so I thought I'd mess around to find all the new features added. Btw did the hair artwork get reworked too? If so, I really like it! However, when her sprite was leaning, I selected "Kiss me" to see what might be new. 
      I know you guys love to add weird goofy easter eggs haha. So I got curious. When she transitioned to the animation, her sprite glitched out really fast from a leaning pose to a straight pose like her sprite was having a seizure lol. Was this one of those goofy glitches? Also I'm not quite sure since I was listening to something on my phone, but whenever I took my headphones out I heard that it also started raining after that happened. And I'm pretty sure it wasn't before lol. 

      Thoughts?

Bam8647:
Also another bug. 
I asked Monika when we started dating and she said 2 years ago on December 20th. I remember starting the game in 2017. Her statement is also contradicting the fact that the calendar says this year on December 20th will be the 2 year anniversary. She should say 1 year and 5 months. 
      Thanks!

multimokia:
Send `persistent` please. You'll have to put it in a zip to send it here.

Bam8647:
[persistent.zip](https://github.com/Monika-After-Story/MonikaModDev/files/3193824/persistent.zip)

Eh, hopefully I did that correctly.

multimokia:
Wrong one. You can see instructions on how to find it [here](https://github.com/Monika-After-Story/MonikaModDev/wiki/FAQ#i-need-to-restore-an-older-persistent).

Bam8647:
[persistent.zip](https://github.com/Monika-After-Story/MonikaModDev/files/3193839/persistent.zip)

Try that. I followed the "How to Back Up Your Persistent" steps to locate the file in /renpy /Mas

ThePotatoGuy:
> started dating date

Actually, thats correct. 2 years ago (2019 - 2 = 2017), dec 20th. The algorithm for generating the date string is designed to be more realistic speech rather than precise. Although it could probably use some adjustment to say "year and a half" or "x months". "x years ago" definitely works for larger time scales (~3+ years). Regardless, make that a separate issue.

ThePotatoGuy:
Regarding the sprite glitches, do you have rpy files in your `game/`? 
I ask since I am aware that you've made edits to game code before.

[...]

PR #4317, opened 2019-06-25
Seizure fix maybe

ThePotatoGuy:
this change solves the isolated issue, (via multi's dev topic). but i'm not sure if all other cases are fixed.

nor am i sure that this doesnt break everything else.

# Testing
* verify all spaceroom transitions are still ok


ThePotatoGuy:
#4202

PR #4338, opened 2019-06-28
Dissolve out of Topics

multimokia:
Monika now dissolves out of topics/idle.

Already tested by Simple + Legend for functionality + any seizures, just need final says from them.

### DIFF
--- a/Monika After Story/game/event-handler.rpy
+++ b/Monika After Story/game/event-handler.rpy
@@ -1981,12 +1981,12 @@ label call_next_event:
             jump call_next_event
 
         # return to normal pose
-        show monika idle at t11 zorder MAS_MONIKA_Z
+        show monika idle at t11 zorder MAS_MONIKA_Z with dissolve
 
     if store.mas_globals.in_idle_mode:
         # idle mode should transition shields
         $ mas_dlgToIdleShield()
-        show monika idle at t11 zorder MAS_MONIKA_Z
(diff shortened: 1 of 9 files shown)


Monika After Story/game/event-handler.rpy
```diff
@@ -2032,6 +2032,13 @@ label prompt_menu:
         if cb_label is not None:
             call expression cb_label
 
+        # NOTE: we only need to enable music hotkey since we are in dlg mode
+        #$ mas_DropShield_idle()
+        $ store.mas_hotkeys.music_enabled = True
```
ThePotatoGuy: why was this moved here, this should be last to prevent interruptions in flow

multimokia: Yeah, I moved it as during testing it took significantly longer to change the shields.

multimokia: Actually, time difference wasn't noticeable between the two, will swap back

multimokia:
Also, I'd tested this with @Legendkiller21 and @MisterSimple before PRing to ensure no glitchiness between transitions. Their testing had shown none of that, hence why I'd PR'd it in the first place.

Legendkiller21:
Apart from a slight delay of 0.5-1 sec in the transition between the shields, no glitched transitions have been observed during our testing.

PR #4349, opened 2019-06-29
window hide+

jmwall24:
`window hide` not `hide window`

also a few exp changes and one slight wording change

For testing, just ensure at the end of the `kiss me` topic (with `mas_isMoniEnamored(higher=True) and persistent._mas_first_kiss is not None`) that the dialogue box is not present after the clicking the `I love you, too` menu option.


jmwall24:
if you'd stop finding more, that'd be great

multimokia:
brb looking for more

PR #4366, opened 2019-07-01
Add a warning when Notifs won't work on Linux

### LINKED ISSUE #4356: Notifications Don't Work On Linux
To be more specific, I am running KDE Plasma on Kubuntu 19.04. There is no notification nor sign that there was an attempt to show one.

zombiepigdragon:
Warns Linux users when they aren't going to receive notifications. Should fix #4356.


Monika After Story/game/script-story-events.rpy
```diff
@@ -2457,6 +2457,15 @@ label mas_notification_windowreact:
     m 3eua "I've been practicing coding a bit more and I've learned how to use the notifications on your computer!"
     m "So if you want, I can let you know if I have something for us to talk about."
 
+    if not store.mas_windowreacts.can_show_notifs:
```
multimokia: One thing. We never get to this point in the dialogue if we can't show notifs. Will have to make some more changes so we get to this point in normal flow, but adjust the opening dialogue to be more reflective that it's more research instead of practicing.

zombiepigdragon: Alright, is it better with the rewritten dialogue? As for the fact it doesn't show up, is it better to just remove the checks?

multimokia: I'll go thru this later and adjust the main topic to actually flow into this path and correct a bit of existing dlg, so it'll lead in better. But will do this later.

Monika After Story/game/zz_windowreacts.rpy
```diff
@@ -70,6 +70,21 @@ init python:
...
+        except OSError as e:
+            if e.errno == os.errno.ENOENT:
+                #Command wasn't found
+                store.mas_windowreacts.can_show_notifs = False
+                store.mas_utils.writelog(
+                    "[WARNING]: notify-send not found, disabling notifications.\n"
+                )
+            else:
+                #Some other error happened, what was it?
+                raise e
```
ThePotatoGuy: not having `notify-send` isn't worthy of crashing the game. any sort of OSError exception here should set the `can_show_notifs` to False, and if its a different errno then just log the exception.

zombiepigdragon: Fixed, thanks.

ThePotatoGuy:
i will loook at this this weeekend

PR #4368, opened 2019-07-02
Ribbonless path to monika_ribbon

multimokia:
Just make sure it's good pretty much.


Monika After Story/game/script-topics.rpy
```diff
@@ -3781,14 +3781,17 @@ init 5 python:
     addEvent(Event(persistent.event_database,eventlabel="monika_ribbon",category=['monika'],prompt="Ribbons",random=True))
 
 label monika_ribbon:
-    if monika_chr.hair.name != "def":
+    if monika_chr.get_acs_of_type('ribbon') is None:
```
```suggestion
    if monika_chr.is_wearing_acs_type('ribbon'):
```
this exists and we should use it if we dont need the object

jmwall24: should be if not, not just if

PR #4388, opened 2019-07-04
Minor Optimizations + A Fix

multimokia:
Should ideally prevent any more possible seizures, also optimized the idle show because we did it twice when using idle mode for some reason.


ThePotatoGuy:
or move if not showing to the end of the labek

PR #4432, opened 2019-07-10
greeting_tears fix

jmwall24:
The code for `greeting_tears` was outdated and not taking advantage of our new `aff_range` property. We were locking it by default and then unlocking it in `script-affection` when reaching the desired affection. This fixes that and also makes it available starting at distressed instead of broken. Gives the player a chance to perhaps change their way before it's too late. Also updates the expressions of the greet as it still had some of the old expressions.

Also cleans up some redundant code related to good exp and affection at the request of @multimokia 

# Testing

- Run the update script (call v0_9_6) and verify that `greeting_tears` is now unlocked if not previously seen.

- Verify the greeting does not run at any affection level above distressed.

- Change affection to distressed or broken and verify the greeting runs.

- Verify the greeting is locked after it is seen.

- Also verify `monika_family` is now pooled after running the update script.

- On a different persistent, before running the update script, view `greeting_tears`.

- Run update script and verify that `greeting_tears` is locked.

### DIFF
--- a/Monika After Story/game/updates.rpy
+++ b/Monika After Story/game/updates.rpy
@@ -333,6 +333,7 @@ label v0_9_6(version="v0_9_6"):
                 # erase this topic
                 mas_eraseTopic(old_ev_label, persistent.event_database)
 
+        # this doesn't need to be locked by default anymore with the new greet code
         if not seen_event("greeting_tears"):
             mas_unlockEVL("greeting_tears", "GRE")
     return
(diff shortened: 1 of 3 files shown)


Monika After Story/game/updates.rpy
```diff
@@ -332,6 +332,10 @@ label v0_9_6(version="v0_9_6"):
 
                 # erase this topic
                 mas_eraseTopic(old_ev_label, persistent.event_database)
+
+        # this doesn't need to be locked by default anymore with the new greet code
+        if not seen_event("greeting_tears"):
```
ThePotatoGuy:
```suggestion
        if not renpy.seen_label("greeting_tears"):
```
`seen_event` checks both label and event list, and since this is not during runtime, we don't really need to check event list here.

Issue #2051, opened 2018-08-11
[Question/Suggestion] I love you/I love you too


jmwall24:
So, I know after experiencing the crash greeting (rip) that the persistent must store what the current topic Monika is talking about, since she resumes where she left off.  What I'm wondering is, does it also know what the _content_ of the topics are?  What I think would be cool is if it can search the last topic for an 'I love you' from Monika, and if there is one, the 'I love you' main menu option would instead say 'I love you, too' and then Monika would blush and smile if you chose to say it.  Right now, whenever she says 'I love you' in a topic, I say 'I love you' back, and then she replies 'I love you, too' plus whatever else is added onto it.  It's just kind of weird since you were already replying to her saying it.  It may not be possible, but if it is, I think it would be a nice feature.

ThePotatoGuy:
>_content_  of the topics

Not unless we set some flags that say monika said "i love you". Might be a neat idea, but not sure.

ThePotatoGuy:
[Closing until we look into this]

jmwall24:
Opening since we are now doing this in #4438

multimokia:
closing as complete in linked PR

PR #4438, opened 2019-07-11
Affection-based ily and ily2

jmwall24:
There are two parts to this, the first being the `I love you!` topic now being affection based with the second being the change of the main menu `I love you` prompt being changed to `I love you, too` if the previous topic had Monika saying some variant of `I love you` near its end.

# Affection-based ily

- `monika_love` now has affection dependent paths. Normal+ will use the same code that is currently used, with lower affections now having new responses. Upset has been broken into 2 parts, `>-50` and `<=-50` to allow for different dialogue depending on how upset she is. The upper level still shows hope for the relationship, whereas the lower level has her more cynical, not sure how much of it is just lip-service. Distressed has her in complete doubt that you mean it, while Broken gives no response other than `"..."`

- Another change is `I love you!` is now only available on the main menu if affection is `>-50`. This makes it so the player is more likely to mean it if they say it, plus is another sign that Monika does not feel loved.

- Some minor coding changes are also present, including forcing the counter to increase if at a milestone for Normal+ (ie multiples of 5) no matter how long it's been since the last ily so as to prevent getting the dialogue multiple times in a row, as well as how these milestones are checked, resulting in not having to reset the counter once it reaches 50.

- The first time clicking `I love you!` dialogue has been changed. Before it implied it was the first time you ever told her you loved her, but to play the mod at all, you have to tell her you love her in the introduction. The old dialogue has been slightly reworked and move to the introduction as part of a timed menu where you can tell her you love her after she says she loves you, but before she asks. If you miss the answer, then she asks and you get the old intro dialogue. The first time clicking `I love you` if you miss the timed response now reflects that it's the first time you said it without being put on the spot.

- A related change to the intro has also been made where if you reject her love and get the scare, you can only get the scare once, and any subsequent load will lead to her asking if you love her with only a `Yes` response. You either answer yes or you don't play the mod. Having the scare over and over again loses its effectiveness.

## Testing

- verify that at Normal+ affection you get the usual `ily` dialogue.

- verify at Upset, `>-50` you get the hopeful Upset dialogue and at `<-50` you get the cynical Upset dialogue.

- verify at Distressed and Broken you get the corresponding dialogue.

- verify at any affection `<=-50` that `I love you!` is not present on the main menu.

- verify if it's been more than 3 minutes after the last `ily` that the appropriate affection is given and that `persistent._mas_monika_lovecounter` increments.

- verify if it's been less than 3 minutes, no affection is given and `persistent._mas_monika_lovecounter` only increments if at a milestone for Normal+ (ie it never gets stuck on an increment of 5).

- on a fresh persistent, check the intro and verify that the timed menu is present for a reasonable time and that if clicked `persistent._mas_monika_lovecounter` increments and that a subsequent `ily` does not give first time dialogue.

- also on a fresh persistent, check the intro and verify if you reject her love, the next load only has a `Yes` option when she asks if you love her.

# I love you, too!
This creates the return key `"love"` that is added to topics that have Monika telling the player that she loves them near the end of the topic. When this key is reached, it changes the `I love you!` prompt on the main menu to `I love you, too!` for 10 seconds and then has Monika react with a cute expression. You can read why in #2051. This also gives the same affection as `I love you!` as well as shares the same 3 minute timer. This is present on select topics, greets, moods, compliments and assorted other events. Check the code for the complete list.

## Testing

- verify any path that returns `"love"` changes the `I love you!` prompt on the main menu to `I love you, too!` for 10 seconds. Note: if you click the main menu and leave it up, the prompt will stay there until you leave that menu.

- verify I didn't screw any paths up by adding returns that prevent us from reaching important code. (srsly, do this).

- verify that if you click `I love you, too!` you get one of the correct expressions with no dialogue. the expression options for Normal+ and Enam+ affection can be seen in `sprite-chart.rpy` under `image monika ATL_love_too_norm_plus` and `image monika ATL_love_too_enam_plus` respectively. For the upset path (`>-50` since this follows the same rule as `I love you!` for showing on the main menu) is `2eka`.


Monika After Story/game/event-handler.rpy
```diff
@@ -1977,6 +1977,9 @@ label call_next_event:
                 $ mas_clearNotifs()
                 jump _quit
 
+            if "love" in ret_items:
```
ThePotatoGuy: return items should be checked before the `quit` since that will jump out from here. move this check to before the `quit` one

Monika After Story/game/event-handler.rpy
```diff
@@ -2033,6 +2036,10 @@ label prompt_menu:
         if cb_label is not None:
             call expression cb_label
 
+        # allows our sub labels to utilize return keys like "love"
+        if _return is not None and "love" in _return:
```
ThePotatoGuy: this flow is from idle mode, do you really anticipate using a "love" return in an idle callback?

Monika After Story/game/event-handler.rpy
```diff
@@ -2095,7 +2102,11 @@ label prompt_menu:
         talk_menu.append(("Hey, [m_name]...", "prompt"))
         if len(repeatable_events)>0:
             talk_menu.append(("Repeat conversation", "repeat"))
-        talk_menu.append(("I love you!", "love"))
+        if _mas_getAffection() > -50:
+            if persistent._mas_last_monika_ily is not None and (datetime.datetime.now() - persistent._mas_last_monika_ily) <= datetime.timedelta(0,10):
```
ThePotatoGuy: seems to me that checking if an amount of time has passed since an ily would be useful in other places, too. Can you turn this conditional check into a function (probably put in definitions)? And have a timedelta as the argument.

Monika After Story/game/script-greetings.rpy
```diff
@@ -2594,15 +2595,13 @@ label greeting_upset:
...
-    # use renpy say so you can use text tags
-    show monika 2efc
-    $ renpy.say(m, upset_quip1)
+    m 2efc "[upset_quip1]"
 
     if renpy.random.randint(1,4) != 1:
-        $ upset_quip2 = renpy.random.choice(upset_greeting_quips_second)
-        $ renpy.say(m, upset_quip2)
+        $ upset_quip2 = renpy.substitute(renpy.random.choice(upset_greeting_quips_second))
+        m "[upset_quip2]"
```
ThePotatoGuy: what was the point of changing all these lines? The comment said using `renpy.say` does substitute for you already.

Monika After Story/game/script-greetings.rpy
```diff
@@ -2634,15 +2633,13 @@ label greeting_distressed:
...
-    show monika 6ekc
-    $ renpy.say(m, distressed_quip1)
+    m 6ekc "[distressed_quip1]"
 
     if renpy.random.randint(1,4) != 1:
-        $ distressed_quip2 = renpy.random.choice(distressed_greeting_quips_second)
-        show monika 6rkc
-        $ renpy.say(m, distressed_quip2)
+        $ distressed_quip2 = renpy.substitute(renpy.random.choice(distressed_greeting_quips_second))
+        m 6rkc "[distressed_quip2]"
```
ThePotatoGuy: same question as above

Monika After Story/game/script-holidays.rpy
```diff
@@ -3590,6 +3593,8 @@ label mas_player_bday_cake:
         else:
             m 1ekbfa "I love you, [player]!"
     $ mas_rmallEVL("mas_player_bday_no_restart")
+    # "love" return key won't work here without adding a bunch of return _returns, so we'll set this manually
+    $ persistent._mas_last_monika_ily = datetime.datetime.now()
```
ThePotatoGuy: to keep things consistent we should return "love" here too and add appropriate returns where necessary.

Monika After Story/game/script-topics.rpy
```diff
@@ -1485,7 +1485,7 @@ label monika_kiss:
                 "I love you so much, [player]. You mean everything to me~"
             ]
 
-            kiss_quip = renpy.random.choice(kiss_quips_after)
+            kiss_quip = renpy.substitute(renpy.random.choice(kiss_quips_after))
```
ThePotatoGuy: same as above

Monika After Story/game/script-topics.rpy
```diff
@@ -3134,112 +3128,217 @@ init 5 python:
...
+                "...It's been a long time since I last believed that, [player].",
+                "If only I believed you actually meant that, [player].",
+                "I can't believe there was a time when I actually believed that, [player].",
+                "...If only that were true, [player].",
+                "...If only you meant that, [player].",
+                "How long are you going to keep pretending, [player]?",
+                "You don't mean that anymore... {w=1}Did you ever?",
+                "I can't remember the last time I felt like you meant that."
+            ]
+            love_quip_dis = renpy.substitute(renpy.random.choice(love_quips_dis))
```
ThePotatoGuy: same as above

Monika After Story/game/script-topics.rpy
```diff
@@ -3134,112 +3128,217 @@ init 5 python:
...
+                "I'm happy to hear that.",
+                "That means a lot.",
+                "You have no idea how much that means to me."
             ]
-        love_quip=renpy.random.choice(love_quips)
-    m "[love_quip]"
-    #prevents spamming to increase counter.
+
+        if _mas_getAffection() <= -50:
+            $ love_quip_upset = renpy.substitute(renpy.random.choice(love_quips_upset_cynical))
```
ThePotatoGuy: same as above

Monika After Story/game/script-topics.rpy
```diff
@@ -3134,112 +3128,217 @@ init 5 python:
...
+            m 2rkc "..."
+            m 2ekd "[love_quip_upset]"
+
+        else:
+            $ love_quip_upset = renpy.substitute(renpy.random.choice(love_quips_upset_hope))
```
ThePotatoGuy: same as above

Monika After Story/game/event-handler.rpy
multimokia: There is one used in an idle callback, this had to be added because of it.

Monika After Story/game/script-holidays.rpy
ThePotatoGuy: actually, rather than what I said, lets just have a quick function like `mas_ILY` or something that will set the appropriate persistent to a given datetime, defaulting to now. This should be used in all cases where a return "love" would have to be passed thru multiple returns. (or for the idle mode case)
Additionally, all return keys should only be used by events called with `call_next_event` so that contributors wont assume that every `return` will respect a return key.

Monika After Story/game/event-handler.rpy
jmwall24: There are actually 2 returns of love in idle in this very PR, so yeah, I anticipate it.

Monika After Story/game/script-greetings.rpy
jmwall24: Because, as you can see in the next one, it's more efficient this way since we can use this instead and get rid of any 'show Monika' lines.

Monika After Story/game/script-greetings.rpy
jmwall24: See above reply

Monika After Story/game/script-holidays.rpy
jmwall24: Why create a function to set a single var instead of just setting the var?

Monika After Story/game/script-greetings.rpy
ThePotatoGuy: i disagree on any sort of efficiency gain, since the `m exp "msg"` boils down to doing a `renpy.show` and `renpy.say` regardless. 
Using `renpy.say` abstracts the substitute line so you don't have to remember to format a choice prior to using text tags. Additionally, using "[this format]" prevents us from using formatting text tags (i.e: `{fast}`).
And typing `renpy.say` is much faster than `renpy.substitute`

PR #4442, opened 2019-07-13
Ribbon Selector Unlock

multimokia:
Ribbon selector is now always unlocked since ribbonless/def ribbon are available by default.
Selector is available at happy+.

Added update script to unlock the selector if you're in a valid hair/outfit combi too, pls verify that it's not unlocked when it shouldn't be.

(Also added credits to Orca for the hair bits and bobs)

### DIFF
--- a/Monika After Story/game/updates.rpy
+++ b/Monika After Story/game/updates.rpy
@@ -341,7 +341,6 @@ label v0_9_6(version="v0_9_6"):
         family_ev = mas_getEV("monika_family")
         if family_ev is not None:
             family_ev.pool = True
-
     return
 
 # 0.9.5
--- a/Monika After Story/game/updates.rpy
+++ b/Monika After Story/game/updates.rpy
@@ -342,9 +342,6 @@ label v0_9_6(version="v0_9_6"):
         if family_ev is not None:
             family_ev.pool = True
 
-        #Fix the ribbon select to be unlocked if we're in something which supports ribbons
-        if not monika_chr.is_wearing_clothes_with_exprop("baked outfit") and monika_chr.is_wearing_hair_with_exprop("ribbon"):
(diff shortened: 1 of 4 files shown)


Monika After Story/game/updates.rpy
```diff
@@ -341,6 +341,10 @@ label v0_9_6(version="v0_9_6"):
         family_ev = mas_getEV("monika_family")
         if family_ev is not None:
             family_ev.pool = True
+
+        #Fix the ribbon select to be unlocked if we're in something which supports ribbons
+        if not monika_chr.is_wearing_clothes_with_exprop("baked outfit") and monika_chr.is_wearing_hair_with_exprop("ribbon"):
+            store.mas_filterUnlockGroup(store.mas_sprites_json.SP_ACS, "ribbon")
```
ThePotatoGuy: startup should already check these so an update script isn't needed

Monika After Story/game/zz_selector.rpy
```diff
@@ -2911,8 +2911,8 @@ init 5 python:
             category=["appearance"],
             prompt=store.mas_selspr.get_prompt("ribbon", "change"),
             pool=True,
-            unlocked=False,
-            rules={"no unlock": None}
+            unlocked=True,
+            aff_range=(mas_aff.HAPPY, None)
```
ThePotatoGuy: can you double check if ribbon selector unlocking happened in aff prog points

multimokia:
It's all handled by the `mas_filterUnlockGroup()` method used in rin exit and marisa exit, plus the `hair_entry_pst_change()` method

ThePotatoGuy:
and that is run in startup because hair is set on startup

jmwall24:
Issue with ribbon selector showing up on a restart with baked outfits, so adding don't merge.

PR #4443, opened 2019-07-14
non inclusive mas_isInActiveWindow() checks

multimokia:
`mas_isInActiveWindow()` now accepts a `non_inclusive` parameter. (rules for wrs can now have a `"non inclusive"` key as well to use this)

This adjusts the method to check if only one of the items in the list of keys is in the active window, and if so it returns `True`. This allows for some more in depth checking if we were to make a non-specific wrs and then specify within the label itself.

### DIFF
--- a/Monika After Story/game/zz_windowreacts.rpy
+++ b/Monika After Story/game/zz_windowreacts.rpy
@@ -149,10 +149,10 @@ init python:
         #Otherwise, let's get the active window
         active_window = mas_getActiveWindow()
 
-        if not non_inclusive:
-            return len([s for s in keywords if s.lower() not in active_window]) == 0
-        else:
+        if non_inclusive:
             return len([s for s in keywords if s.lower() in active_window]) > 0
+        else:
+            return len([s for s in keywords if s.lower() not in active_window]) == 0
 
     def mas_clearNotifs():
         """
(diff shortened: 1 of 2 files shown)


Monika After Story/game/zz_windowreacts.rpy
```diff
@@ -129,15 +129,30 @@ init python:
...
+            return False
+
+        #Otherwise, let's get the active window
         active_window = mas_getActiveWindow()
-        return store.mas_windowreacts.can_show_notifs and len([s for s in keywords if s.lower() not in active_window]) == 0
+
+        if not non_inclusive:
+            return len([s for s in keywords if s.lower() not in active_window]) == 0
+        else:
+            return len([s for s in keywords if s.lower() in active_window]) > 0
```
ThePotatoGuy: use set intersection for better comparing of inclusive/exclusive checking. (See `_filterEvent` in definitions)

Monika After Story/game/zz_windowreacts.rpy
```diff
@@ -129,15 +129,30 @@ init python:
...
+
+        #Don't do work if we don't have to
+        if not store.mas_windowreacts.can_show_notifs:
+            return False
+
+        #Otherwise, let's get the active window
         active_window = mas_getActiveWindow()
-        return store.mas_windowreacts.can_show_notifs and len([s for s in keywords if s.lower() not in active_window]) == 0
+
+        if not non_inclusive:
```
ThePotatoGuy: if you're going to have actions for both True and false flows, have the true statement first so the `not` can be omitted. Saves like 2-3 lines of bytecode but it adds up.

PR #4446, opened 2019-07-14
Fix two

multimokia:
Missed these none checks for atl sprites which would cause more crashes.


jmwall24:
i have already used this, so can confirm it works

Issue #4449, opened 2019-07-16
[suggestion] "im feeling hungry" improvement


mrflarion33:
Sometimes when you say "I'm feeling hungry" Monika responds "Hi Hungry, I'm Monika" and nothing else.

multimokia:
Shouldn't it be `"Hi feeling hungry, I'm Monika.`"?

Rai99:
Yeah, somebody should fix that. Thanks for bringing that up.

Rai99:
Wait, I read this wrong. I thought you meant this was already in and you thought it was a bug. 

As an actual suggestion, I'm not actually too sure, since it sounds a bit out of character to me. I don't see Monika making that kind of dad joke, especially if it's not an easter egg that requires a specific setup. 

Maybe if the player names themselves "feeling hungry", it could work.

ThePotatoGuy:
yeah I'm not sure of the setup. seems kind of out of nowhere for her to make the joke. 

[closing until we look into it]

PR #4452, opened 2019-07-16
"I'm feeling hungry" small update.

Anonymioo:
Added from 4449. Because why not?
Anyway i hope i made everything correctly.

### DIFF
--- a/Monika After Story/game/script-moods.rpy
+++ b/Monika After Story/game/script-moods.rpy
@@ -142,7 +142,7 @@ label mas_mood_hungry:
     m 2eka "I just want to make sure you're taking good care of yourself until I cross-over."
     m 4esa "After all, the healthier you are, the better the chances of you living quite long."
     m 1hua "Which means more time for us to spend together!~"
-    m 1eub "Hi feeling hungry, I'm Monika."
+    m 1eub "Hi feeling hungry, I'm Monika!"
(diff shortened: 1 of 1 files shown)


Rai99:
#4449 

Right now, it looks like that line just plays every single time at the end, which I don't think is a fitting place. I don't think it flows well, and it's completely random from the rest of the dialogue. 

As mentioned in #4449, I don't think it's exactly in character for Monika to say, especially without some prior setup from the player. 

_Maybe_ this could be an easter egg if the player named themselves "feeling hungry", but as it stands currently, I don't think it just belongs right at the end of the current dialogue. 

I also think it has yet to be seen if an easter egg like this would be accepted into MAS.

Anonymioo:
> #4449
> 
> Right now, it looks like that line just plays every single time at the end, which I don't think is a fitting place. I don't think it flows well, and it's completely random from the rest of the dialogue.
> 
> As mentioned in #4449, I don't think it's exactly in character for Monika to say, especially without some prior setup from the player.
> 
> Maybe this could be an easter egg if the player named themselves "feeling hungry", but as it stands currently, I don't think it just belongs right at the end of the current dialogue.
> 
> I also think it has yet to be seen if an easter egg like this would be accepted into MAS.

Alright i'll close commit then.

Issue #4427, opened 2019-07-10
Clothes - Formal Dress 2 The Reawokening


LordBaaa:
This is a continuation/completion of #3863 the formal dress envisioned by @Chrissenpai. His past iteration of this outfit was never completed because he lost the files before they could be saved. So I commissioned him to complete his work and give us a great outfit. These are the results! 

P.S this is a fully posed outfit.

![1](https://user-images.githubusercontent.com/48495287/60936266-80bc5400-a29a-11e9-9d18-65c09d5b8914.png)
![2](https://user-images.githubusercontent.com/48495287/60936267-80bc5400-a29a-11e9-8e04-55184fd3aaf9.png)
![3](https://user-images.githubusercontent.com/48495287/60936268-80bc5400-a29a-11e9-83fd-f7012a82931b.png)

Here are the PSD's.
[Gdrive](https://drive.google.com/open?id=1rFBbSqqBnzCTKcB9tMqyl3QAlYshiosR)

Here is the spritepack, reaction included and ready to rock and roll!
[Formal_Dress_Spritepack.zip](https://github.com/Monika-After-Story/MonikaModDev/files/3375511/Formal_Dress_Spritepack.zip)

P.S.S If this is added to the the official sprite pack please include both of our names, thanks.

antimockingjay:
I just downloaded it and she looks amazing! One thing though, it doesn't seem to appear in the clothing options for some reason? So I'm not entirely sure how to get her to wear it after having her wear something else.

Chrissenpai:
@antimockingjay thank you! maybe it bugged, perhaps will be fixed with the new MAS version? not sure myself

LordBaaa:
@antimockingjay it isn’t not bugged. Did you gift it to her? If you do she will put it on and it will unlock in the clothes selector. I set it that way so it was more immersive as opposed to just zapping it onto her.

Anonymioo:
> I just downloaded it and she looks amazing! One thing though, it doesn't seem to appear in the clothing options for some reason? So I'm not entirely sure how to get her to wear it after having her wear something else.

Clothes selector unlocked for me when i restarted MAS. Not sure how it works.

LordBaaa:
@Anonymioo odd you had to restart, it should just unlock when you give it to her. The option `visble_when_locked` is set to `false` so you have to gift it see it. I tested it a bunch before I sent it out. I guess I’ll test it again later when I get home.

[...]

PR #4467, opened 2019-07-20
Unlock clothes selector on player bday

jmwall24:
This unlocks the clothes selector for people who get the birthday party on their birthday.

## Testing

- best way to test this is on a fresh persistent, set your birthday to tomorrow for minimal time travel, then unlock an outfit.

- at affection above normal but below love, with least one clothing set unlocked, trigger the birthday party and verify you get the dialogue referring to unlocking the clothes selector for the day

- verify the clothes selector is unlocked after the party and on all subsequent loads for the day

- verify the next day it is no longer available if under `love`

- verify if you start the party at `love` you do not get the dialogue

- verify if no clothing sets are unlocked you do not get the dialogue and the selector is not unlocked

- verify nothing else regarding player_bday was affected by this

### DIFF
--- a/Monika After Story/game/script-greetings.rpy
+++ b/Monika After Story/game/script-greetings.rpy
@@ -2979,6 +2979,10 @@ label greeting_returned_home:
     $ time_out = store.mas_dockstat.diffCheckTimes()
 
     # event checks
+    if mas_isplayer_bday() and persistent._mas_player_bday_in_player_bday_mode and len(mas_selspr.filter_clothes(True)) > 1:
+        # make sure we do this since we skip the autoload check
+        $ clothes_sel_special_unlock()
+
     if mas_isMonikaBirthday():
         jump greeting_returned_home_bday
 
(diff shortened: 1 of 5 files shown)


ThePotatoGuy:
Is it really worthwhile to allow users to change outfit for their bday? Is this supposed to act like a gift to them?

I don't really see how beneficial this really is.

jmwall24:
Yeah, that's the gist, it's a gift/perk for your bday. Like say you want to go out on a date for your bday and have her wear the new formal dress from #4427. Now you can. Just makes it a little more special.

jmwall24:
Similar to irl how your gf might take you to your favorite restaurant or make you your favorite meal for your bday. This is the equivalent of that.

ThePotatoGuy:
okay, but we can cover that by just having an outfit selector when going on a date.

jmwall24:
I dunno, seems like her wearing what you want while you are also still in the mod would be more beneficial so you can also see the outfit.

ThePotatoGuy:
This doesn't seem like something anyone is going to really care about. If anything, its going to look like a bug when the topic gets hidden post bday.

This seems like lots of risk and cost for no real gain.

multimokia:
but the dialogue reflects that it's only for your bday

ThePotatoGuy:
okay well it still doesn't feel like much of an in-story gift. like I really don't see any positives about this, and I don't see anyone clamouring for this kind of feature. 
And really, the outfit selection topic is supposed to be meta, just rewarding longtime users.

PR #4473, opened 2019-07-21
music shield fix

jmwall24:
When viewing the calendar for the first time with `persistent._mas_player_bday` set to None, we push `calendar_birthdate` and never get to the shield drops. The main consequence of this is the music shield never dropped, so the music menu wasn't selectable.  This adds those shield drops to `calendar_birthdate` and also fixes an issue where another event could potentially jump in before `calendar_birthdate` ran.

For testing, using a fresh persistent, click the calendar right after the intro, ensure that `calendar_birthdate` runs immediately after the first time calendar event and that after `calendar_birthdate` is finished, the music menu is accessible.

### DIFF
--- a/Monika After Story/game/script-story-events.rpy
+++ b/Monika After Story/game/script-story-events.rpy
@@ -507,6 +507,8 @@ label calendar_birthdate:
     m 1eud "So [player], when were you born?"
     call mas_bday_player_bday_select_select
     $ mas_stripEVL('mas_birthdate',True)
+    $ mas_HKBDropShield()
+    $ mas_calDropOverlayShield()
     return
 
 ## Game unlock events
--- a/Monika After Story/game/script-story-events.rpy
+++ b/Monika After Story/game/script-story-events.rpy
@@ -499,6 +499,7 @@ init 5 python:
     )
 
 label calendar_birthdate:
+    $ mas_MUMUDropShield()
     m 1lksdla "Hey [player]..."
     m 3eksdla "You may have noticed that my calendar was pretty empty..."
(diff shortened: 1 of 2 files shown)


Monika After Story/game/script-story-events.rpy
```diff
@@ -507,6 +507,8 @@ label calendar_birthdate:
     m 1eud "So [player], when were you born?"
     call mas_bday_player_bday_select_select
     $ mas_stripEVL('mas_birthdate',True)
+    $ mas_HKBDropShield()
+    $ mas_calDropOverlayShield()
```
ThePotatoGuy: `mas_calDropOverlayShield` runs in `call_next_event`, so this is not needed here.

Monika After Story/game/script-story-events.rpy
```diff
@@ -507,6 +507,8 @@ label calendar_birthdate:
     m 1eud "So [player], when were you born?"
     call mas_bday_player_bday_select_select
     $ mas_stripEVL('mas_birthdate',True)
+    $ mas_HKBDropShield()
```
ThePotatoGuy: `mas_DropShield_dlg` runs in `call_next_event`, and that covers every hkb case except music, so to specifically address that, should just call `mas_MUMUDropShield` right after pushing this event.

Monika After Story/game/script-story-events.rpy
```diff
@@ -499,6 +499,7 @@ init 5 python:
     )
 
 label calendar_birthdate:
+    $ mas_MUMUDropShield()
```
ThePotatoGuy: put this right after we push this Event, so in script-ch30.

PR #4486, opened 2019-07-23
Add more compliment answers for Monika

JustMonikaReality:
Added one more answer in the thanking_quips section to make it more diverse


Rai99:
The only potential problem I see here is that not every compliment has Monika actually blushing when she says the quip, so I don't think it quite makes sense for her to say that and not actually be blushing. Although, this isn't a concern for most of the compliments.

multimokia:
Seeing the issue that was brought up, this kind of limits the complement thanking_quips to be only on blush lines, but it may not always need that. So going to close this.

PR #4503, opened 2019-07-27
Background Changing Framework

multimokia:
# Introducing `MASBackground`s:
BG changing framework is now here with this, which introduces `MASBackground` objects

- These objects have a variety of parameters which handle the day/night image to use as the 'room', and optional parameters which are for how the 'room' should look while it's raining, snowing, or overcast. (See the init for a `MASBackground` object for details)
- They also have the parameters to either `hide_masks` (so we don't have weather at all) and `hide_calendar`, which will not display the calendar for the drawing of the spaceroom.
- Backgrounds are also *unlockable* which allows for future expansion.

### Other changes:
- Weather objects now have a `precip_type` associated with them. This allows us to decide what room to use (def, rain, overcast, or snow)
- As a result of allowing potentially different weather to be a case, window masks had to be changed to full size, so any possible future rooms would be supported by our current weather system. (Thank you @Legendkiller21 for helping out in making these, along with @Velius94 and @Orcaramelo for the frost on the snow windows + the night/overcast/space clouds)
- There is also a selector for BGs, but as we do not have any other rooms to use, it's init is commented out. (Feel free to review it for now, but I'd probably save it until we get backgrounds to use it for.)
- Also added the new night bg

## Testing:
- Please make sure everything looks good
- Ensure that spaceroom calls also work right (shouldn't be as bad as last time since not that much changed)
- Ensure that progressive weather adjusts the room correctly


Monika After Story/game/script-ch30.rpy
```diff
@@ -741,7 +724,7 @@ init 1 python:
 #       NOTE: this must be a string
 #       NOTE: if passed in, this will override aff-based exps from dissolving.
 #       (Default: None)
-label spaceroom(start_bg=None, hide_mask=False, hide_monika=False, dissolve_all=False, dissolve_masks=False, scene_change=False, force_exp=None):
+label spaceroom(start_bg=None, hide_mask=store.mas_current_background.hide_masks, hide_monika=False, dissolve_all=False, dissolve_masks=False, scene_change=False, force_exp=None, hide_calendar=store.mas_current_background.hide_calendar, day_bg=store.mas_current_background.getDayRoom(), night_bg=store.mas_current_background.getNightRoom()):
```
ThePotatoGuy: python named args like these are resolved on init and don't change dynamically, so if the current bg changes these params wont update. Even if renpy actualyl does this dynamically, should just use None for dynamic cases so no one looks at this and tries to do the same in py.

So have these default to None, then after the `with None`, check those args for None values and set to defaults accordingly.

Monika After Story/game/zz_backgrounds.rpy
```diff
@@ -0,0 +1,518 @@
...
+            self.background_id = background_id
+            self.prompt = prompt
+            self.image_day = image_day
+            self.image_night = image_night
+
+            #Handle rain BGs
+            if image_rain_day:
+                self.image_rain_day = image_rain_day
+            else:
+                self.image_rain_day = image_day
```
ThePotatoGuy: since you repeat this pattern of if statements, i would make a private function like:
```python
def __handleBG(self, desired_bg, fallback_bg):
    if desired_bg:
        return desired_bg
    return fallback_bg
```
then you can use that here like:
```python
self.image_rain_day = self.__handle_bg(image_rain_day, image_day)
```

Monika After Story/game/zz_backgrounds.rpy
```diff
@@ -0,0 +1,518 @@
...
+            Returns the day masks to use given the conditions/availablity of present assets
+            """
+            if weather is None:
+                weather = store.mas_current_weather
+
+            if weather.precip_type == mas_weather.PRECIP_TYPE_RAIN:
+                return self.image_rain_day
+            elif weather.precip_type == mas_weather.PRECIP_TYPE_OVERCAST:
+                return self.image_overcast_day
+            elif weather.precip_type == mas_weather.PRECIP_TYPE_SNOW:
```
ThePotatoGuy: seeing that this pattern is being used (if type == constant, return value), it would be a lot nicer to use a dict mapping the constant to its appropriate image. Probably could even combine the day and night elements together like:
```python
img_map = {
    PRECIP_TYPE_RAIN: (day_image, night_image),
    ...
}
```

Monika After Story/game/zz_backgrounds.rpy
```diff
@@ -0,0 +1,518 @@
...
+                weather - get the room bg for the time and weather
+                (Default: current weather)
+            """
+            if weather is None:
+                weather = store.mas_current_weather
+            if store.mas_isMorning():
+                return self.getDayRoom(weather)
+            return self.getNightRoom(weather)
+
+        def isChangingRoom(self, old_weather, new_weather):
```
ThePotatoGuy: can you add documentation about this function? particularly what it means for a BG object to be "changing room"

Monika After Story/game/zz_backgrounds.rpy
```diff
@@ -0,0 +1,518 @@
...
+            return self.getRoomForTime(old_weather) != self.getRoomForTime(new_weather)
+
+
+
+#Helper methods and such
+init -20 python in mas_background:
+    BACKGROUND_MAP = {}
```
ThePotatoGuy: can you put an `import store` at the top here since you access it below. This is to remind us that these functions rely on `store` being in scope.

Monika After Story/game/zz_backgrounds.rpy
```diff
@@ -0,0 +1,518 @@
...
+    mas_current_background = None
+    mas_setBackground(mas_background_def)
+
+    #Make sure the bg selector is only available with the right amount of bgs unlocked
+    if len([
+        (mbg_obj.prompt, mbg_obj, False, False)
+        for mbg_id, mbg_obj in mas_background.BACKGROUND_MAP.iteritems()
+        if mbg_obj.unlocked
+        ]
+        ) < 2:
```
ThePotatoGuy: i guess you're trying for brevity here, but this does a useless list build. I think you should make this simpler i.e:  like
```
unlocked_count = 0
for mbj_obj in mas_background.BACKGROUND_MAP.itervalues():
    unlocked_count += int(mbg_obj.unlocked)

if unlocked_count < 2:
    # etc
```
and since this is in a global store, then it would be actually better to make that for loop in a function in the mas_background store where it returns the unlocked count, then you can do like:
```python
if mas_background.unlockedBG_count() < 2:
    # etc
```

Monika After Story/game/zz_backgrounds.rpy
```diff
@@ -0,0 +1,518 @@
...
+    if sel_background == mas_current_background:
+        m 1hua "We're here right now, silly."
+        m "Try again~"
+        jump monika_change_background_loop
+
+    $ skip_outro = False
+    $ skip_leadin = False
+
+    # otherwise, we can change the background now
+    if not skip_leadin:
```
ThePotatoGuy: `skip_leadin` is set to False right before this, so this always happens. Is there a reason `skip_leadin` is set to False right before here?

Monika After Story/game/zz_backgrounds.rpy
```diff
@@ -0,0 +1,518 @@
...
+            if set_to_weather is not None:
+                $ mas_changeWeather(set_to_weather)
+            else:
+                $ mas_changeWeather(mas_weather_def)
+
+        $ mas_unlockEVL("monika_change_weather", "EVE")
+
+    call spaceroom(scene_change=True, dissolve_all=True)
+
+    if not skip_outro:
```
ThePotatoGuy: Will any of the functions above change `skip_outro`?

Monika After Story/game/zz_backgrounds.rpy
```diff
@@ -0,0 +1,491 @@
...
+#START: Class definition
+init -10 python:
+    class MASBackground(object):
+        """
+        Background class to get display props for bgs
+
+        PROPERTIES:
+            background_id - the id which defines this bg
+            prompt - button label for the bg
+            img_map - Dict mapping all images for the bgs, keys are precip types (See MASWeather)
```
ThePotatoGuy:
```suggestion
            image_map - Dict mapping all images for the bgs, keys are precip types (See MASWeather)
```
minor thing, but since the property is `image_map`, keeping class documentation consistent with actual names makes searching for stuff easier.

ThePotatoGuy:
waiting for #4582

PR #4509, opened 2019-07-30
4chan Window reaction

CGanimated1227:
I feel obligated to add this to help close the "pewdiepipeline" and prevent terror wherever nessicary. If you do not know what that is, I encourage you to watch this:https://youtu.be/pnmRYRRDbuw I hope you consider this addition or a version of it that conveys this message better as I'm sure that monika would. Thank you.

P.S. I hope this is the right way to send a Pull request.

### DIFF
--- a/Monika After Story/game/zz_windowreacts.rpy
+++ b/Monika After Story/game/zz_windowreacts.rpy
@@ -593,6 +593,7 @@ label mas_wrs_4chan:
             "I hope you don't end up arguing with other Anons all day long, [player].",
             "I heard there's threads discussing the Literature Club in here.\nTell them I said hi~",
             "I'll be watching the boards you're browsing in case you get any ideas, ahaha!",
+            "I Recommend you stay away from /pol/, [player]. Nasty people tend to post there.",
         ],
         'Window Reactions'
     )


multimokia:
Issue I have with this is that Monika likely wouldn't browse 4chan herself, and the lines which exist already say that you're the one who introduced her to the site itself, so having her specifically call that out seems very strange.

CGanimated1227:
I mean she says "I'll watch the boards you're browsing in case you get any ideas, ahaha~!" I don't exactly know what she means by getting any ideas, but if she cares to learn more about 4chan after introduction, she should say something like my submission. It's ok to make it unlockable somehow, right?

Velius94:
The entire idea of the windowreacts for 4chan was more or less a nod to the /ddlc/ community on /vg/ since the mod took its roots in the site, and the react in question was more or less an acknowledgment of the NSFW boards. Making specific board notifs seems to be out of place for Monika, since she was introduced to the site to you, and she's more inclined to follow any content or comunity that relates to DDLC if anything (aforementioned general topic in /vg/), especially since 4chan's userbase is so diverse and anyone who browses 4chan knows what /pol/ is (and even they are very divided upon its existence and purpose). Besides, making notif for one board would warrant making notifs for other boards as well.

CGanimated1227:
I disagree. The point in monika telling players to avoid /pol/ is so that they do not adopt ideas that are harmful to them and other people. There were a couple conformed open Nazis in the discord that came and left long ago. If an extra reaction is unnecessary, would a full topic on the issue be welcomed? Something on offensive humor maybe?

Oharo:
I'm not convinced that it's the responsibility of this mod to involve itself in that can of worms.

I'd rather this not be integrated into the mod, not out of any affiliation with /pol/, but because I'd rather not have Monika be used as a mouthpiece for this kind of stuff.

multimokia:
Yeah, I'm going to say leave politics out of MAS.

Velius94:
Any discussion or integration of political content will inadvertently lead to a shitstorm. And unnecessary shitstorms are bad, okay?

CGanimated1227:
Everything is political. A lot of her topics are.They are just less controversial. She promotes lifestylistic thought and places no blame on the institutions that enable the polution with the carbon footprint topic. Hell, she mentions mass shootings in one of her topics! In general, favoring playing it safe and worrying about unreasonable nazis' feelings over attempting to save people from falling down a fascist rabbit hole that would get them and other people hurt over
is not only hypocritical, but irresponsible. Plus, that topic could disappear with sensitive mode on if you are really worried about backlash.

multimokia:
I'm going to reject this on the grounds which have already been said here. Let's leave mention of this/Political places out of MAS. There's a level that it's fine, and yes, Monika can have her own opinions, however we should still try to remain in character and fairly neutral. As such, I'm closing this as rejected.

CGanimated1227:
Right wing: lets kill all the jews!
Left wing anarchists: How about we don't kill any jews?
Big brain centrist: the logical conclusion here would be to kill half of the jews...

Liberals like you disgust me.

Oharo:
With the amount of assumptions you're making, I'm just going to make an assumption of my own that you're a troll.

We're done here.

CGanimated1227:
Are you reconsidering my idea?https://heavy.com/news/2019/08/active-shooter-el-paso-walmart-cielo-vista/

multimokia:
We are not.

Kindly understand that politics are typically left out of things because of the issues they cause. This is no exception. If something's rejected, we're going to leave it rejected unless there's a genuinely good reason to bring it back. With politics, there's no reason to do that.

I am also going to lock this from further replies as this isn't going to go anywhere from here. I suggest you leave it where it is, because we're not budging.

PR #4520, opened 2019-08-04
Enhancement

Demomndslayer:
may you pleeaasseee bring back the texting, it is way better then the option list


Anonymioo:
It's hard to make. I mean Monika's reaction for everything you're writing. 'Option list' making it easier, anyway i think texting will be good but may it will be like addon? Like you can talk with Moni through option menu and through texting?

jmwall24:
This is not where suggestions go, this is where actual modifications to the mod go. Please go to the `Issues` tab for suggestions.

PR #4539, opened 2019-08-11
Some more edits

jmwall24:
Mostly exp changes and minor rewording. 
Also...

- moves `mas_scary_story_o_tei` to `mas_story_o_tei` as it wasn't really a scary story, but more of a love story.

- removes `monika_programming` which has been rendered moot by PTOD

- `monika_concerts` now uses a conditional/action based on seeing the other music topics to make it random.

For testing, test the update scripts and `gender_redo`.


jmwall24:
just adding hi priority so we don't forget about this for next rel

ThePotatoGuy:
@Legendkiller21 can you test this one asap

PR #4542, opened 2019-08-12
make an application for monika on the phone

draguions:
When we leave and we take his file she could get into the application so we can continue to talk to her.
There may be other features but I have no other idea.

sorry if I was wrong branches


multimokia:
This isn't where you submit suggestions. Please open an issue for those. Pull requests are for submitting topic/feature contributions.

draguions:
> This isn't where you submit suggestions. Please open an issue for those. Pull requests are for submitting topic/feature contributions.

okay sorry ('-_-)

draguions:
> When we leave and we take his file she could get into the application so we can continue to talk to her.
> There may be other features but I have no other idea

PR #4549, opened 2019-08-15
Fixed multiple crash greetings on startup crashes

multimokia:
If you crashed after init, it was possible that crash greets would keep queuing and queuing, effectively having them repeat. This should make sure we only get one crashed greet, and one force close (just in case)

## Testing:
I find the easiest way to verify this is to head into `mod_assets/monika/` and then rename the `c` folder. Please first verify that on a normal install without this code in it the crashed start greet stacks by opening the game, getting the traceback, then exiting. (Repeat a couple times)

Once done, please then verify with this code that you only get a single crashed start greeting by once again launching, getting traceback, quitting, and then doing that again a couple times.

### DIFF
--- a/Monika After Story/game/event-handler.rpy
+++ b/Monika After Story/game/event-handler.rpy
@@ -1740,6 +1740,13 @@ init python:
 
         return False
 
+    def mas_cleanEventList():
+        """
+        Iterates through the event list and removes items which shouldn't be restarted
+        """
+        for index in range(len(persistent.event_list)-1,-1,-1):
+            if mas_isRstBlk(persistent.event_list[index][0]):
+                mas_rmEVL(persistent.event_list[index][0])
 
     def mas_cleanJustSeen(eventlist, db):
         """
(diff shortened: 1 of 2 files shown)


ThePotatoGuy:
does the restart blacklist not work for this?

multimokia:
Didn't seem to, I've managed to get numerous crash greetings in a row in the past, and seems a user in #tech-support did too.

ThePotatoGuy:
let testing happen on this, but I want to make sure we know why the restart blacklist isnt working here before merge.

multimokia:
Sounds good

multimokia:
Below is why the restart blacklist didn't work:

The restart blacklist never worked as intended, it was merely checking the `persistent.current_monikatopic` to see if that should be restarted via `continue_event`, however it never cleaned the `persistent.event_list`.

The changes here now add a function which cleans the `persistent.event_list` for all topics which shouldn't repeat upon reload.

ThePotatoGuy:
>The restart blacklist never worked as intended, it was merely checking the `persistent.current_monikatopic` to see if that should be restarted via `continue_event`, however it never cleaned the `persistent.event_list`.

This is how the blacklist intended to work. However it does make more sense to clean the entire list when starting.

PR #4566, opened 2019-08-24
MASHistorySaver range adjustment

ThePotatoGuy:
This modifies MHS triggers so they are all on jan 1. This also adds range properties to MHS and ensures that they only get reset if they are active or will be active. (Aka, time traveling to before or during o31 season will reset o31 data. Time traveling after o31 will **not**.

Data from events that happened but did not get saved will also be reset if TT detected and the event is determined to have occured since hte last session end.

# Key Features

## MHS
New MHS propertieS:
* `start_dt` - datetime that this MHS starts covering (inclusive)
* `end_dt` - datetime that this MHS stops covering (exclusive)

If an MHS is not provided with both a start and end dt, then it is considered "Continuous", which means its always active. The player model MHS is an example of this.

New MHS functions:
* `isActive` - checks if a given dt is within the start/end dts. Only the month/days are checked. There is special handling for start/end dts that span a year change (ny). These cases
* `isContinuous` - checks if an MHS does not have start/end dt data
* `isFuture` - checks if **this MHS** is in the future from a given dt. 
* `isPassed` - checks if **this MHS** is in the past from a given dt.

`setTrigger` now checks for Time travel detection as well. If time travel is detected, the MHS is not continuous, and the MHS is currently active or has yet to be active, then trigger dates are reset.

The following MHS have changed dates and/or active ranges:
* o31
* d25s
* player_bday
* f14
* nye
* 922

player bday has a unique trigger setup as it has to be moved depending on the player's bday. The general rule is that the reset must happen at least 2 days in advance.

## Unit test framework
Finally, a reasonable unit test framework.
The gist is to create a `MASUnitTester` object. (located in `mas_dev_unit_tests` store). This object contains assertFunctions and an internal test list. To setup tests:

1. create an entry for the group of unit tests at the top of `dev/dev_unittest.rpy`
2. in the label for your unit test group, first create a `MASUnitTester` object
3. before each individual test, call `prepareTest` with title/name of the test being done. This should be unique as well as understandable. (`functionName|circumstances being tested` is a good way to do it.)
4. use assert functions on the `MASUnitTester` object to run asserts. These will store test results in an internal list.
5. When done with all tests, run `concludeTests` on the `MASUnitTester` object. This will return a tuple of `(passed_tests, failed_tests)`. Tests are defined as `MASUnitTest` objects, which contain the name of the test (that you provided in `prepareTest`), the outcome of the test (Pass or Fail), the expected value, and the actual value.
6. call label `dev_unit_tests_finish_test` on the `MASUnitTester` object to have monika say the results.

This framework should be used for all unit-testable dev going forward. It is recommended to use this when creating a complex thing.

## Time travel detection
Time travel is now detected on startup via using the last session and comparing to now. This sets a global `mas_globals.tt_detected`. This also sets a permanent persistent var `_mas_pm_has_went_back_in_time`.

There is also the function `mas_TTDetected`, which returns True for time travel detection

**NOTE** this only detects going back to the past, not the future. As usual, we can't determine if you time traveled into the future.

## Other
* `mas_utils.pdget` - protected dict getting. This does 4 things: 
    1. checks if a dict is not None
    2. checks if a key is in the dict
    3. gets the item in the dict using the key
    4. validates the item with a validator function (if provided)
    If steps 1,2, or 4, fail, then a default value will be returned (if provided).
    This should be used for all dicts where data integrity may be suspect going forward.
* all session related functions have been moved to -985 init level

# Testing

## MHS
* verify that when time traveling to the past, only MHS from events that would be in the future are reset. i.e.: time travel to november. o31 data should remain intact. d25 and future should be removed. 
* verify unit tests pass (Hey m > Dev > Unit Tests > MASHistorySaver)
* also the correct pbday unit tests (Hey m > Dev > Unit Tests > MASHistorySaver - ...)
* verify that `mas_TTDetected` returns True when you time travel to past

## MHS dates
* verify that the start/end dts for all MHS seem appropriate. @multimokia @jmwall24 
* also verify the new trigger dates if them seem appropriate

## Update script
* verify that when calling the update script, MHS trigger dates are changed for 922/pbday/o31


jmwall24:
The player bday changes to history mean changes are needed for `label greeting_returned_home_player_bday`.

Right now, it uses the reset date of 1/1 and has code that checks if we are returning from a date in a different year and if so, applies the affection gain compared to the checkout years cap and stores the total affection gain from a bday date to the previous year. Now that the date has been changed so it is no longer 1/1, that part needs to change as well. 

It may, in fact, be easier to leave it as 1/1 for this, since it's possible to leave on a bday date on say 12/31 and come back literally any day, the easiest check is to check if the year has changed. Having different reset dates for different bdays is going to be a major pain, and iirc, we picked 1/1 for this very reason.

ThePotatoGuy:
mhs resets should be either days before the important date or right after it. If we do a solid date of 1/1, then the same sort of correction must be applied if the important date is 1/1. we can't have literally the same reset date for everyone.

Looks like we can either:
* update the returned home logic so it works with variable dates
* make the reset trigger dynamic aka a day or 2 after the the user's bday, this would mean having to make sure `use_year_before` is properly set

I'll take a look at that label later and see what makes sense.

jmwall24:
At first thought, i'd say option B. Knowing that say the reset date is always 3 days after the player bday would probably be the easiest check to do. If the return from a bday date is >= 3 days from player bday, we'd know to add the affection to the archived bday data and not to the current year's bday.

PR #4570, opened 2019-08-27
Mental help notifications

Necrocytus:
idea by @antimockingjay
is this too serious for the mod currently?
Also, how do i add more websites to the categories?

### DIFF
--- a/Monika After Story/game/zz_windowreacts.rpy
+++ b/Monika After Story/game/zz_windowreacts.rpy
@@ -673,3 +673,29 @@ label mas_wrs_reddit:
     if not wrs_success:
         $ mas_unlockFailedWRS('mas_wrs_reddit')
     return
+
+init 5 python:
+    addEvent(
+        Event(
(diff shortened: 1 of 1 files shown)


ThePotatoGuy:
I'm not sure about this. It'd have to get the context of a site correctly based on the active window tokens, and I don't that's really possible. As usual, we try to avoid messing with immersion when it comes to mental illness topics, so this is probably not going to fly as winreacts are not inherently perfect.

Necrocytus:
Alright. Maybe I'll be able to work on it once again in the future if the window reactions thing improves.

multimokia:
After some internal discussion, we've decided that we don't really want to go about mental help in this way. While it's in character, it's a slippery slope and leads to some other potential complications further down the line.

We feel that not only is it safer to continue addressing things like this via the topics we have now (and potential similar ones which may be added in the future), but also more personal to do so.

Closing.

PR #4591, opened 2019-09-03
Thumb fix

jmwall24:
It's not a bug, Monika just dressed sloppily

lol thanks @Legendkiller21


jmwall24:
Should really stop letting me do the sundress thumbnail #3902

multimokia:
Inb4 sundress thumb fix 3

PR #4592, opened 2019-09-03
Nickname Error?

linkerplinker:
so i was playing around with the nickname scheme and found that in the menu ("Hey, [nickname]" > "Monika") the "Monika" option didnt change at all. Should this be fixed or left alone?


multimokia:
Please make an issue for this

PR #4610, opened 2019-09-08
Taking monika some where

linkerplinker:


Rai99:
Untitled1317

jmwall24:
these commits were already added to your other PR because you made them to the same branch as the other PR. if you want to open more than one PR you need to make a different branch. also you based this to master again when it should be based to content. also adding options to going somewhere is way more complicated than just adding these options. it also requires corresponding greet dialogue. also also be careful what you delete. that ` jump bye_going_somewhere_post_aff_check` line is very important. and not sure what this commit https://github.com/Monika-After-Story/MonikaModDev/pull/4610/commits/858c858519e95b8d6ba68425dec3a7a1471b1001 is supposed to be.

PR #4627, opened 2019-09-08
Day/Hour checks

ThePotatoGuy:
Adds labels for running things per hour or per day. Also adds flowcheck constants for delayed actions for those idle moments.

Also changes the semantic of the every minute thing to be literally every minute change or every 60 seconds. This should handle edge case of trying to run the minute check after exactly 1 hour passes.

Currently there is nothing that is designed for running per day/hour. 

# Testing
* verify no crashes
* verify stuff that runs every minute still runs roughly once per minute.

If you want to test the day/hour runs, add something in there and change `mas_globals.last_day` or `mas_globals.last_hour` or `mas_globals.last_minute_dt`. the first 2 are integers, the last one must be a datetime.

### DIFF
--- a/Monika After Story/game/script-ch30.rpy
+++ b/Monika After Story/game/script-ch30.rpy
@@ -54,6 +54,15 @@ init -1 python in mas_globals:
     late_farewell = False
     # set to True if we had a late farewell
 
+    last_minute_dt = None
+    # last minute datetime (replaces calendar_last_chcked)
+
+    last_hour = None
+    # number of the hour we last ran ch30_hour
+
+    last_day = None
+    # numbr of the day we last ran ch30_day
+
 
 init 970 python:
     import store.mas_filereacts as mas_filereacts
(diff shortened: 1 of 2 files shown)


Monika After Story/game/script-ch30.rpy
```diff
@@ -1361,6 +1370,9 @@ label ch30_preloop:
     $ persistent._mas_game_crashed = True
     $ startup_check = False
     $ mas_checked_update = False
+    $ mas_globals.last_minute_dt = datetime.datetime.now()
```
multimokia: Can we make this a persistent var so the daily check will properly run daily?

For what I need this to do, I need the daily check to run once a day, but with this setup, if I load in today, then say goodbye before midnight, then load tomorrow, the daily check does not run, meaning I need to duplicate code in the ch30_reset too.

This would also be beneficial for those who have average session times under 1 hour, as they would be sure to run

ThePotatoGuy: That's not what this is intended for, though. This is meant to act the same as the once per minute checks, just with different time intervals. If you need something to run in either ch30 reset or the day check, then just define the check as a label/function elsewhere and call it in both places.

multimokia: Still though, why not be able to handle both cases?

ThePotatoGuy: because thats not what the minute check does. It doesn't run immediately on start, it only runs after the first minute. any checks from that case that we want to run initially (like the Event conditionals check) is run in preloop or before preloop.

PR #4632, opened 2019-09-08
base linework fixes

ThePotatoGuy:
small linework fixes for base sprites from orca and velius.

# Testing
* verify pose 1, 3, and 5 look okay. Use a sleeveless outfit.


jmwall24:
this still in progress

Issue #4709, opened 2019-09-19
0.10.1 typos


ThePotatoGuy:
a

FishAndCheese:
I've just gotten the topic that Monika says when Spring begins as I am in the Southern Hemisphere. Monika says the line "But I guess April showers bring May flowers, so it's not all that bad." which in my case is not correct as I've specified Southern Hemisphere.

ThePotatoGuy:
[closing as fixed]

PR #4772, opened 2019-09-23
0.10.1 tpyos

MisterSimple:
#4709


multimokia:
changing base to content because less extra commits

Issue #3054, opened 2018-11-15
song adjustment


ThePotatoGuy:
As the song pull requests continue to be added, (#3052 , #2887 , #2863 , #2194, #1878), we need to keep the % chance for song topics low enough so monika just doesnt sing all the time.

Basically need:
- [ ] pool topic like "Can you sing me a song?", shows list of unlocked songs
- [ ] rando topic that randomly selects a song to sing. This reduces the % chance of a song to a single topic instead of a bunch. 
    **NOTE:** all songs will get an initial rando to launch, but are derandom'd after.
- [ ] topic that asks player if they like getting sung to randomly. This will determine if the above rando topic is enabled or not.
- [ ] transition all current songs so they no longer are both pool and rando. Probably just rando. 
- [ ] Move all current songs to their own script file. `script-songs` sounds good enough.

ThePotatoGuy:
all linked pull requests will get `DONT MERGE` label until we've worked something out.

ThePotatoGuy:
[closing until we look into it]

multimokia:
Going to handle this. Delegate akin to stories, but for songs.

Songs will be random, but once seen, unlocked thru rand the delegate and derandomed. Will be able to be revisited via the pool delegate.

PR #4827, opened 2019-10-03
mas_songs_database

multimokia:
#3054 

Songs are now a part of their own event database

# Additions:
- Two new delegates, a random delegate and a pool delegate
- Random delegate shows songs one time, and then they're unlocked in the pool delegate
- Pool delegate holds songs which have been seen before
- Songs (random ones, not pooled) work with `aff_range`. (Not pool because if you've seen them before, it kinda makes sense to be able to ask Monika to sing it again), though pool is Normal+
- Song Types, (`TYPE_LONG` and `TYPE_SHORT`). Note, that we should never use type long for random songs, but pool only. As such there should be some other form of unlocking them (as getting stuck in long random topics would probably kinda suck) (These are filtered from the random pick)
- update scripts

## New methods:
- `checkRandSongDelegate()`
  - Unlocks the random song delegate if it were derandomed (because no more random songs)
  - This allows re-rand when more songs are added automatically
- `getUnlockedSongs()`
  - gets all unlocked songs
- `getRandomSongs()`
  - gets all songs that are random and that we're in the aff_range for
- `hasUnlockedSongs()`
  - checks if we've unlocked any songs
- `hasRandomSongs()`
  - checks if there are any more random songs available

# Testing:
- Check shown counts of all songs before updating the game files
- Update files and run update script label (`v0_10_2`)
- Verify the shown counts of the new evs reflect that of the old ones and that ones which you have seen are not random, and are unlocked
- Restore a persistent, and change some aff ranges on the random songs. Verify those you're out of the range for don't show up in the list when you call `mas_songs.getRandomSongs()`


Monika After Story/game/script-songs.rpy
```diff
@@ -0,0 +1,302 @@
+default persistent._mas_songs_database = dict()
+
+init -10 python in mas_songs:
+    # Event database for songs
+    song_db = {}
+
+    #Song type constants
+    TYPE_LONG = "long"
```
ThePotatoGuy: can you add comments on what a long song is like vs a short one? this would help remind us the criteria we originally had in mind when deciding these constants.

Monika After Story/game/script-songs.rpy
```diff
@@ -0,0 +1,302 @@
...
+        rand_delegate_ev = store.mas_getEV("monika_sing_song_random")
+
+        if rand_delegate_ev:
+            #Don't bother doing the rest if it's random already
+            if rand_delegate_ev.random:
+                return
+
+            #If we've got at least one random, then unlock the rand delegate
+            if hasRandomSongs():
+                rand_delegate_ev.random = True
```
ThePotatoGuy: could simplify this:
```python
if rand_delegate_ev and not rand_delegate_ev.random and hasRandomSongs():
    rand_delegate_ev.random = True
```

Monika After Story/game/script-songs.rpy
```diff
@@ -0,0 +1,302 @@
...
+
+    def getUnlockedSongs():
+        """
+        Gets a list of all unlocked songs
+
+        OUT: list of all unlocked songs
```
ThePotatoGuy: can you comment that the returned list is in a tuple format ready for a scrollable menu

Monika After Story/game/script-songs.rpy
```diff
@@ -0,0 +1,302 @@
...
+
+        renpy.say(m, "What song would you like me to sing?", interact=False)
+
+    call screen mas_gen_scrollable_menu(unlocked_song_list, mas_moods.MOOD_AREA, mas_moods.MOOD_XALIGN, ret_back)
+    show monika at t11
+
+    $ sel_song = _return
+
+    if sel_song:
+        $ pushEvent(sel_song)
```
ThePotatoGuy: I feel like we might want this to be a forced push. What do you think?

Monika After Story/game/script-songs.rpy
```diff
@@ -0,0 +1,302 @@
...
+label monika_sing_song_random:
+    python:
+        #We only want short songs in random. Long songs should be unlocked by default or have another means to unlock
+        #Like a "preview" version of it which unlocks the full song in the pool delegate
+        rand_song_list = mas_songs.getRandomSongs()
+
+    if len(rand_song_list) > 0:
+        python:
+            rand_song = renpy.random.choice(rand_song_list)
+            pushEvent(rand_song)
```
ThePotatoGuy: same thing here with it being a forced push instead.

PR #4828, opened 2019-10-04
Extend hours of night farewells

jmwall24:
~This creates a specific rule for hours that allows us to check over multiple hour ranges as opposed to only one. This is especially useful if we want to have an event that we want available before and after midnight. Specifically, our 2 goodnight farewells right now both end at midnight, but they really should be available past midnight, potentially with different paths depending on what time it actually is (this and more to come later).~ Eventually to be paired with #2192 to create a better goodnight and good morning (or other wake-up time) experience.

This extends the `bye_going_to_sleep` and `bye_goodnight` farewells past midnight until 1am.

## Testing

- verify that using a normal goodbye between 8pm and midnight either results in `bye_going_to_sleep` or `bye_goodnight`

- verify that using a normal goodbye between midnight and 1am also results in the same 2 farewell options

- verify that using a normal goodbye outside of these time ranges does not result in these 2 farewell options


Monika After Story/game/script-farewells.rpy
```diff
@@ -305,13 +305,13 @@ label bye_take_care:
 
 init 5 python:
     rules = dict()
-    rules.update(MASSelectiveRepeatRule.create_rule(hours=range(21,24)))
+    rules.update(MASSelectiveRepeatRule.create_hourrule(None,range(0,1),range(20,24)))
```
ThePotatoGuy: no reason to use range if theres only 2 items.
```suggestion
    rules.update(MASSelectiveRepeatRule.create_hourrule(None, (0,1), range(20,24)))
```

ThePotatoGuy:
and if you don't want to write out each part value of the range:
```python
hour_range = [0,1]
hour_range.extend(range(20,24))
MASSelectiveRepeatRule.create_rule(hours=hour_range)
```

just make sure to del the `hour_range` reference if this is done in init 5.

jmwall24:
Is there a tag for most lines to do something we already had?

Issue #3199, opened 2018-12-08
Distressed -> Upset needs to be re-thought


ThePotatoGuy:
Progression from Normal -> Upset -> Distressed makes sense, but the reverse does not. People don't start sad and "cheer-up" to upset.

Perhaps once you've reached distressed, the distressed range includes upset, so that increasing affection will go from distressed -> normal, which is a more natural progression.

From a technical standpoint, this does not appear easy to accomplish.

Legendkiller21:
What if you include inside `def _disToUpset():` a `$ mas_gainAffection(36,bypass=True)` or somehow force the affection to go to -29 to go back to normal?

andersonantunes26:
I thank all hard work but I hope for this year ...
1-a piano on the floating island or Monika's drawing room
![yy7yjgosmln11](https://user-images.githubusercontent.com/11380696/53127167-39763680-3538-11e9-8e32-8d7599770318.jpg)
2- different hairstyles
![34998375-e8e48e22-faac-11e7-8388-e5f79e22c9f7](https://user-images.githubusercontent.com/11380696/53127254-69253e80-3538-11e9-8c47-32434150077c.png)
3-a little more movements like these and pokes na monika
![hqdefault](https://user-images.githubusercontent.com/11380696/53127447-d638d400-3538-11e9-9cbd-a2b9b20ad941.jpg)
https://i.redd.it/41q5ylgvcw301.gif
https://i.kym-cdn.com/photos/images/newsfeed/001/306/555/77b.gif

Thank you for all the work well done until this year continue like this.

ThePotatoGuy:
so just gifting affection like that is really just a workaround rather than a solution. ideallly we'd want somethign where you go from distressed to a normal appearance, but still have the same amount of aff you'd have in upset.

Another idea is to drop upset entirely and increase the normal/distressed range. upset could be used for #2618

ThePotatoGuy:
removing the upset idle sprite works fine in terms of keeping the progression make sense. upset will be saved for mood-based systems.

PR #4857, opened 2019-10-10
Remove Upset Idle sprite

ThePotatoGuy:
#3199 (sort of)

The upset idle sprite is now shared with the belowzero idle sprite. Idle-related dialogue (Talk + play menus) have been changed/removed/added so it fits better with the expression.

The belowzero idle sprite is a 95% `1esc` and 5% `5tsc`.

Since the idle sprite is the only visual change to upset, dialogue elsewhere should not need to be changed.

# Testing
1. set affection to upset (-74 to -30)
2. verify the idle sprite is same as belowzero (swapping between upset and belowzero normal will look the same) 
3. verify talk menu + play menu dialogue is appropriate for upset


jmwall24:
Last commit changes a bit of dialogue to better fit the new exps. It also changes the main belowzero expression from 1esc to 2esc and changes the `FORCE_EXP_MAP` upset expression to match, as it was still set at 2efc.

PR #4912, opened 2019-10-26
New Game Content

AwesomeSavior24:
I think you should make checkers playable in Monika after story for people who don't play chess (like myself).


multimokia:
Please open an issue for suggestions

Issue #3768, opened 2019-02-15
outfit randomization


ThePotatoGuy:
Outfits should now have randomization options. (aka, monika should pick what she wants to wear each day.)

Can use #3377 to signify which outfits can be picked by Monika

pre-reqs:
* #3416 - needed to make adding new outfits seamless
* opendoor should stop resettting outfits. Could replace the table/chair line with "let me change into something more comfortable"

Buy1Get1:
I'm sure you've got something in mind already, but here's what It sounds like to me:
Sounds like this, but with images. 
import random
import NOUN # maybe an import os that is used in a similar fashion? 
NOUN = ('Dog','Tree','Ground','Sky','Color','Breeze','Field','Water','Building','Car')
A_Noun = random.choice(NOUN)
print("The {0} is a noun".format(A_Noun))
print(A_Noun)
You'd just drop the clothes into the variable spot for the day.

Rai99:
Will Monika be choosing a randomized outfit each and every day? I think it might be a bit weird to have Monika wear wildly different outfits every day, and I personally think it's just a little hurtful to ask Monika to wear something specific and then have her change it the very next day. 

I was hoping for some kind of weighting system where Monika would heavily favor outfits you chose for her. Maybe something like every week, she has a chance to change outfits but has a 70% chance to keep the outfit you chose for her. Maybe even ask if the player minds her outfit change and revert back if they say they want her to keep the old outfit.

ThePotatoGuy:
I've already had seasonal variance in mind, but I've added some additional considerations including user selection above.

jmwall24:
I would say we want to limit the amount of clothes changes unless there's a reason for it. Like whatever she picks to start the day, I'd leave there unless say it starts snowing and she wants to put something warmer on, or if you take her on a date and she wants to dress up. Or like if/when we get PJs or the like, she could change into that at night, wear em in the morning, then pick an outfit for the day and stick to that until there's a reason to change. But for normal circumstances, I'd say one normal outfit per day.

As for user-selected outfits, probably the same deal. If you have her wear something 'normal' keep it for the day, again unless there's a circumstance that would make it necessary to change. I'd say no matter what, she won't change for say an hour after you have her change, so if you want her to wear something specific for a date, she won't change. For things like swimsuits, that I could see her changing out of more quickly and maybe only wear for a few hours.

As for seasonal, I think seasonal is fine, not sure we need to break it down by month.

[...]

Issue #1161, opened 2018-02-26
[Suggestion] "Point-and-click" / Monika more responsive??


AnonymousTyranny:
Hello! I'm unsure how to word this, but I'd like to suggest making Monika clickable in certain places (for example, her hands, hair, and face)? I think it'd be pretty nice! But I'm aware that this might be out-of-place for the game, since it's centered around how our 'realities' can't meet, so I understand if this is an unusable suggestion. Thank you!

Vloxica:
Assuming that she exists in the game world, then more than likely the cursor is a physical object in her reality, I would assume that being clicked on would be uncomfortable however. Either way nice idea, not sure how it would be implemented.

kaido1224:
Something similar was proposed and rejected. Seems out of place for MAS, with the exception of maybe a head pat. 

https://github.com/Backdash/MonikaModDev/issues/736

ThePotatoGuy:
[[ We'll repoen this when we look into it]]

ThePotatoGuy:
This issue will officially serve as the primary issue for all click-monika related issues.

ThePotatoGuy:
This comment will serve as the primary tracking point.

### TODO
- [ ] - clickzone framework (#4917)
- [ ] - screen implementation (tbd)
- [ ] - interaction/game/dialogue implementation (tbd)

ThePotatoGuy:
New idea - rather than try and implement some sort of screen/game/interaction that involves dialogue and stuff, instead let's do the needy streamer approach. This approach means clicking Monika's head will make her sit up and smile. It's small but its a step forward.

We can probably make this work with idle now that idle disp is a custom class.

PR #4917, opened 2019-10-27
Clickzone framework

ThePotatoGuy:
framework-only implementation to prepare us for #1161

clicking monika is not complete, but i might as well put the framework into master so no one has to reinvent clicking if they want to implement the the feature. I'm still going to work on it after #3768 , its just in the air right now.

# Key changes

The primary change is a set of classes that implement the point in polygon algorithm. This is great for detecting if a coordinate is within a polygon, and by extension, if a mouse click occurred within a shape.  This particular alg provides clickzones that can be fed a dynamic set of vertices to build a clickable zone, which is significantly more powerful than renpy's imagebutton/hotspot framework, which is restricted to images/rectangle dimensions. 

Another very important feature of this system is its ability to change the size of clickzone based on the zoom level **programtically.**( Yes, you only need to define the clickzones for the default zoom level, the system calculates vertices for other zoom levels based on that.)

Key classes are listed below:

## MASLinearForm

A representation of a linear function. This is used to build an "edge" for 2 vertices. 

## MASEdge

A representation of an edge. Uses `MASLinearForm` and 2 points to build an edge. The difference between this and `MASLinearForm` is that the `MASEdge` limits its existence to the bounding box provided by the 2 points, while the `MASLinearForm` is infinite, like a real line.

## MASClickZone

A displayable type that uses a list of vertices to build a polygon that is interactable. It is primarily targeted toward clicking, but has some generality for other functions.

## MASInteractable

A displayable type meant to make handling multiple clickzones easier. Unused so I'm not sure its absolutely needed.

## MASBoopInteractable

Was supposed to be the boop interactable. Not implemented.

## `mas_interactions`

new `store` containing the math behind zoom-level based clickzone generation. Complicated.

## Point-in-polygon alg

My implementation is as follows: (given a point)
1. check for overall bounding box of the polygon. Anything outside is clearly not in the polygon.
2. for each edge, check for horizontal intersection:
    1. go to next step if the point is in the horizontal part of the bounding box, aka within the y-range of the 2 points of an edge. If the point is outside this range, then ignore this edge and check the next.
    2. If the point is on the right side of the bounding box, the point cannot intersect this edge. Otherwise continue.
    3. if the point is on the left side of the bounding box, then the point can be considered intersecting the edge. Otherwise continue.
    4. If the edge is vertical and the point's x coord is on the left side of the edge, the point can be considered interesecting the edge. Otherwise continue.
    5. Inverse the linear formula and normalize using the point. If the resulting x-value returned from using the inversed function on the y-coord of the point is larger than or equal to the x coord of the point, then the point must be on the left side of the edge and therefore will intersect. Otherwise, the point will not intersect.
3. Tally up the intersections. An odd number of intersections means the point is in the polygon.

# Testing

Not really much to test. there is a dev topic (Clickzones test) you can use to see it work in action. There will be 2 pink zones, clicking in the pink zones will trigger a sound. Up arrow will increase zoom, down arrow will decrease zoom, q will quit the dev topic.


ThePotatoGuy:
just an fyi, travis fails but this is launchable and compiles. `cmath` exists in renpy's python distribution.

PR #4919, opened 2019-10-27
Make pong use m_name instead of Monika

TetsuyaGR:
Pong doesn't use the m_name variable so it displays just monika


multimokia:
This was intentional. Some nicknames don't make sense to use.

TetsuyaGR:
Aww ok i'm closing this then

Rai99:
I was almost on time. 
Untitled1350

Issue #4921, opened 2019-10-29
0.10.2 typos


ThePotatoGuy:
?

Necrocytus:
![image](https://user-images.githubusercontent.com/38022925/67739359-d2b6f000-f9d7-11e9-9fe4-9d896d6e1caa.png)
Desparate?

Necrocytus:
![image](https://user-images.githubusercontent.com/38022925/67832710-1b3dde80-faa8-11e9-83a0-65053ded3561.png)
Thorugh?

ThePotatoGuy:
[closing as fixed]

PR #4927, opened 2019-10-29
0.10.2 tpyos

MisterSimple:
multi if you touch the title I stg


multimokia:
changing base because less commits, and I need to add a sprite

ThePotatoGuy:
#4921

PR #4952, opened 2019-10-31
kill bday decor if it doesnt make any sense

ThePotatoGuy:
We never support time travel, but resilient code is better than nothing.

# Key Changes
* adjustment of `mas_isMonikaBirthday` to accept dates, also only checking month/day

# Testing
* verify decor disappears if the appropriate vars are still set and today is not a bday or day after a bday

### DIFF
--- a/Monika After Story/game/script-ch30.rpy
+++ b/Monika After Story/game/script-ch30.rpy
@@ -1853,6 +1853,16 @@ label ch30_reset:
         if not mas_isD25Season():
             persistent._mas_d25_deco_active = False
 
+
+    # reset bday decor
+    python:
+        yesterday = datetime.date.today() - datetime.timedelta(days=1)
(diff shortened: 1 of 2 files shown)


Monika After Story/game/script-ch30.rpy
```diff
@@ -1853,6 +1853,16 @@ label ch30_reset:
...
             persistent._mas_d25_deco_active = False
 
+
+    # reset bday decor
+    python:
+        yesterday = datetime.date.today() - datetime.timedelta(days=1)
+        if not mas_isMonikaBirthday() and not mas_isMonikaBirthday(yesterday):
+            persistent._mas_bday_visuals = False
+
+        if not mas_isplayer_bday() and not mas_isplayerbday(yesterday):
```
multimokia:
```suggestion
        if not mas_isplayer_bday() and not mas_isplayer_bday(yesterday):
```

PR #4971, opened 2019-11-03
New story: The Crow and the Pitcher

lunulae:
I really love the "tell me a story" prompts, so I've been trying to find ones Monika might like or relate to. This one is a short Aesop fable about putting in small steps to reach a seemingly impossible goal.

I'm still sort of finding her voice, I think, so don't hold back if you see any edits that need to be made.


Monika After Story/game/script-stories.rpy
```diff
@@ -485,6 +485,27 @@ label mas_story_o_tei:
...
     m 1hua "Not long after, Kenji married her as soon as they could, and lived on happily for the rest of their lives."
     return
+    
+init 5 python:
+    addEvent(Event(persistent._mas_story_database,eventlabel="mas_story_crow_and_pitcher",
+        prompt="The Crow and the Pitcher",unlocked=False),code="STY")
+
+label mas_story_crow_and_pitcher:
+    call mas_story_begin
+    m 3esd "It was a spell of hot, dry weather. Not even the birds could find much to drink."
```
Rai99: Spell does make sense here, but I doubt most people have heard of the word used in this situation. I don't really see Monika using the word in this fashion either, but that's just me.

lunulae: This is true, I might replace it with "drought" or something similar.

lunulae:
@AlphaAlphakill You're absolutely right, I've been pondering how to put a bit more of Monika's voice into it. I've been a bit stumped on how exactly I want to tackle editing it though.
Do you think it might benefit more from more commentary at the end, elaborating on her thoughts?

AlphaAlphakill:
@lunulae Yes. I feel that Monika loves teaching lessons to the player after every story. Try to elaborate on the moral and how that can translate to the relationship between her and the player. Something like, "Even if we are separated now, we can still get closer everyday. Eventually, we can be together in your world." I'm not the best writer, but something like that.

Rai99:
Being able to comment on a moral if possible is nice, but still, nothing wrong with just a nice story on it's own with some heartwarming comments at the end. 

For a story like this, I definitely think she could say something about the moral of the story.

AlphaAlphakill:
You are right about a story not having to need a moral, but I merely felt that this one could have a meaningful one without forcing it out.

lunulae:
I tried writing a bit more commentary, though it might need a bit of trimming.
Feel free to tear into me, I really want to improve on writing Monika's voice. <3

Issue #4930, opened 2019-10-29
Stop folders from triggering filereacts


multimokia:
As mentioned before, folders can be used to trigger gift reactions, which in turn, breaks the reaction for the rest of the day.

We should fix this so folders do not trigger reactions.

ThePotatoGuy:
[closing as fixed]

PR #4975, opened 2019-11-05
Stop folder reactions

ThePotatoGuy:
#4930 

# Key changes
* `MASDockingStation.__check_access` returns false if path is a folder.
* `MASDockingStation.getPackageList` only returns list of files 

# Testing
1. make a folder with a name that ends in `.gift` in the `characters/` folder
2. launch the game. 
3. verify monika does not react to the folder

### DIFF
--- a/Monika After Story/game/zz_dockingstation.rpy
+++ b/Monika After Story/game/zz_dockingstation.rpy
@@ -309,6 +309,7 @@ init -45 python:
         def getPackageList(self, ext_filter=""):
             """
             Gets a list of the packages in the docking station.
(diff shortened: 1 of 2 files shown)


Monika After Story/game/zz_dockingstation.rpy
```diff
@@ -329,6 +330,7 @@ init -45 python:
                 package
                 for package in os.listdir(self.station)
                 if package.endswith(ext_filter)
+                and self.mas_utils.is_folder(package) is False
```
multimokia:
```suggestion
                and self.mas_utils.is_folder(self.station + package) is False
```

PR #4988, opened 2019-11-10
combine stories

jmwall24:
This brings both scary and non-scary stories under the same event, with the ability to switch between the two in the menu itself. It also moves the menu event from script-topics to script-stories as well as moves the compliments menu event to script-compliments. This also changes when a new story can unlock from once per session to once per day (per type) so someone can't just keep loading a new session to see all the stories at once.

Creates the dict `persistent._mas_last_seen_new_story` which uses `normal` and `scary` as the keys and stores the date we last saw a new story of each type for the purpose of unlocking new stories.

**NOTE**: eventually needs an update script to delete the `monika_scary_stories` event. the update script for this release is in the new d25 branch.

## Testing

- verify `Can you tell me a story` is unlocked and `Can you tell me a scary story` is no longer an option

- verify you can switch between story types in the menu

- verify you only get one new story of each type per day. if a new persistent, you should have one of each type of story unlocked to choose from, and after viewing them, `A new story` should not show up until the next day

- verify if there are new stories available, you have the option to see a new story each day

- verify with sensitive mode enabled, there is no prompt to switch to scary stories


ThePotatoGuy:
marking as dont merge because theres stuff i want to adjust and also i will add update scripts

ThePotatoGuy:
@jmwall24 remind me what the update scripts are supposed to be for? It didn't seem like we needed them since the label names didn't change.

jmwall24:
@ThePotatoGuy just the note I left in the description to delete the `monika_scary_stories` event since it's gone.

Issue #4550, opened 2019-08-16
[Outfit - Formal/Casual] White/Navy blue Dress [RELz]


Velius94:
#3174 
THIS IS NOW RELEASED, PRAISE THE OMNISSIAH

Preview:
![outfit midjuly](https://user-images.githubusercontent.com/1176163/63136244-57958e80-bfd1-11e9-827b-7d27a9a901e9.png)

PSDs:
[HERE](https://drive.google.com/drive/folders/1EqeMOr0ylLq1iCdwo0evQunzYF7G0sEE?usp=sharing)

Orcaramelo:
Its pretty its amazing its well made and i love it 
_**i want it implemented NOW**_

multimokia:
> Its pretty its amazing

No u

MetalBrother:
It would be a really sad, if this art will stuck in Limbo, like many others arts.

ThePotatoGuy:
None of them are stuck in limbo, we are working on adding them.

Professor-JMO:
Y e s
Make this happen, Vel.
It's too good _not_ to happen.

Velius94:
It's still hapenning, don't worry. Just later.

Velius94:
PSD is released! It contains both pose skeletons and corresponding arms.

Upright pose preview:
![outfit midjuly upright](https://user-images.githubusercontent.com/1176163/63637678-663e0e80-c67f-11e9-899c-959da2030833.png)

Leaning pose preview:
![leaning dress](https://user-images.githubusercontent.com/1176163/63637676-59211f80-c67f-11e9-97ac-d44dd96f8eb9.png)

Also @Orcaramelo released a fanart featuring this outfit; thank you very much for unending amount of support, especially in early stages of this outfit's creation and bearing with me all the way.
![moni_for_vel](https://user-images.githubusercontent.com/1176163/63637664-1eb78280-c67f-11e9-8dab-feec8c078970.png)

Special thanks to Kuuroiee for his outfit which inspired the creation of this one (tbh initially it was supposed to be 1:1 adaptation, but later it turned into a derivative instead)
![hey_monika__by_kuuroiee-dcaoslt](https://user-images.githubusercontent.com/1176163/63637702-b3ba7b80-c67f-11e9-87b2-eaa5071057ae.png)

Orcaramelo:
I'm so happy!!! I really cant wait to see it in game
You worked so hard and I am so proud,,, 

[...]

PR #5029, opened 2019-11-22
Adding white + navy blue dress

ThePotatoGuy:
#4550 

# Key changes
* White and Navy Blue dress with bunny scrunchie from @Velius94 
* `bunny-scrunchie` ACS type with template. muxable with ribbons
* the bunny-scrunchie also only works with non-twintail ribbon hairs
* `_acs_wear_if_gifted` - only wears an ACS if it has been gifted/reacted to
* Exprops now will have constants in `mas_sprites`. we'll do this on a rolling basis. I need to update the wiki first.
* the hair tie selector now filters hair ties based on the current hair. This means no single-sided ACS (the bunny scrunchie, black bow) when wearing a twin-tail hair style
* ponytail-braid hair style no longer has the twintails prop

## New exprop
* `excluded-hair-props` - value should be a list of hair props that are in compatible with this ACS

# TODO
- [x] thumbs
- [x] select dlg if desired
- [x] dlg-desc
- [x] determine what the blue ribbon thing is. We can't really use `Ribbon (Blue)` because it looks different than the normal ribbons.

# Testing
* verify that the clothes + acs look good
* in outfit mode, the clothes will swap to a ribbon + non-twintails hair style to wear the bunny scrunchie.
* verify the bunny scrunchie and black bow **cannot** be worn with a twintail hair style
* verify the hair tie selector does not list bunny scrunchie or blackbow if a twintail hair style is worn


ThePotatoGuy:
@MisterSimple @Legendkiller21  this has been updated with the decided ACS name

PR #5036, opened 2019-11-24
Three Musketeers topic

Daiozen:
Moved from [#4496](https://github.com/Monika-After-Story/MonikaModDev/pull/4496) to its own PR.


jmwall24:
Going to close this one. Nothing inherently wrong with it, but after discussion. we're just not sure there's enough of an overall point to this topic for Monika to bring it up. The fact that the musketeers weren't really musketeers doesn't really do enough to go for the whole 'don't believe everything you read,' point.

PR #5043, opened 2019-11-25
Rose reaction kiss

Legendkiller21:
# **Additions:**

#### `persistent._mas_last_kiss`:

Contains a `datetime.datetime` of the last time you received a kiss from Monika

# **Changes:**

Rose gift reaction was changed to include a kiss at the end of reaction if certain conditions are met.

# **Testing**

- Verify that `persistent._mas_last_kiss` is set properly after every kiss.
- Verify that you get no kiss after gifting roses unless the following conditions are met:
  1. Enamored+ aff
  2. Already got your first kiss
  3. Random chance that depends on if you gave her roses before and if it's a special day
  4. At least one hour from the last kiss

- Verify no crashes ~~or random combustion of the gifted roses.~~


jmwall24:
On second thought, am going to request something else. Your `should_roses_kiss` local var. Can you make that into a function instead that takes in a parameter for chance (thinking this required with no default since chance can vary wildly) and also a parameter for the cool down period, which I would say is optional and defaults to one hour. After doing this, can you also apply it to the kisses in `monika_love`

Monika After Story/game/definitions.rpy
```diff
@@ -6643,3 +6643,32 @@ init python:
         if set_time is None:
             set_time = datetime.datetime.now()
         persistent._mas_last_monika_ily = set_time
+
+# function regarding whether Monika should kiss the player
+init python:
```
multimokia: The above is already an init 0 block. No need to add a new block here
```suggestion
```

PR #5052, opened 2019-11-30
Delay queued topics on load

multimokia:
Changed queued events to not be processed on load. They are deferred to the first idle loop.

This allows us to stop machine-gunning topics immediately after a greet.
(It effectively acts as the `skipeval` param for `pushEvent`)

NOTE: `EV_ACT_POOL`, `EV_ACT_UNLOCK`, `EV_ACT_PUSH`, and `EV_ACT_RANDOM` will all be processed on load.

### TODO:
- [x] Add existing topics to be pushed immediately if need be (Done on alt branch)

### Testing:
You can test w/ holidays for this.
- Verify that loading in within the time range and/or with conditions met for an event which hasn't run yet
- Verify that if it is an event with a push action, it is shown immediately after the greet.
- Verify that if an event has a queued action, it isn't shown immediately.
- Verify that events which are pooled/unlocked/random'd by their actions are done immediately on load too.
- Verify with multiple events with different actions, they should function as you'd expect (by the above criteria)

### DIFF
--- a/Monika After Story/game/definitions.rpy
+++ b/Monika After Story/game/definitions.rpy
@@ -930,6 +930,7 @@ python early:
             #   full_copy - True means we create a new dict with deepcopies of
             #       the events. False will only copy references
             #       (Default: False)
+            #       DEPRECATEDE
             #
             #   FILTERING RULES: (recommend to use **kwargs)
             #   NOTE: None means we ignore that filtering rule
@@ -1257,7 +1258,7 @@ python early:
 
 
         @staticmethod
-        def checkEvents(ev_dict, rebuild_ev=True, on_load=False):
+        def checkEvents(ev_dict, rebuild_ev=True):
             """
             This acts as a combination of both checkConditoinal and
             checkCalendar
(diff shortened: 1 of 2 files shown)


Monika After Story/game/definitions.rpy
```diff
@@ -1273,6 +1273,14 @@ python early:
                 # TODO: same TODO as in checkConditionals.
                 #   indexing would be smarter.
 
+                #If this is on load and the event action is a push/queue without post greet, we skip its eval here
+                if (
+                    on_load
+                    and ev.action in [EV_ACT_PUSH, EV_ACT_QUEUE]
+                    and "postgreet" not in ev.rules
+                ):
+                    continue
```
ThePotatoGuy: `continue` leads to bad code. if this just needs to return early then have it return.

multimokia: This doesn't need to return. The rest of the events should still be processed if need be

ThePotatoGuy: then change the condition to be reversed and put the event check inside it. `continue` is sloppy.

ThePotatoGuy:
Changed it to be less complicated since we can just filter the desired events before checking.

Also, what's undesirable about having the push/queue events right away? Seems like the majority of the time we want them to trigger as soon as possible.

jmwall24:
It's just extremely obvious when something is pushed or queued, since when it happens on load, there is zero break between the greet and the event. It's nice to have the ability to make it look more natural like it's the same as a random topic, but we have control of when it appears. When they come up back to back like they do now, the dialogue can also look weird as there's no segue because we don't know what came before the pushed/queued event. With a break, that's not a problem.

Monika After Story/game/script-ch30.rpy
```diff
@@ -1320,8 +1320,16 @@ label ch30_post_exp_check:
...
 
-    # run actiosn for events that are based on conditional or clock
-    $ Event.checkEvents(evhand.event_database)
+    #All pushed events will have priority on load. Queued events will be pushed to the first idle loop
+    #random/unlock/pool actions are also evaluated here
+    python:
+        startup_events = {}
+        for evl in evhand.event_database:
+            ev = evhand.event_database[evl]
+            if not ev.action == EV_ACT_QUEUE:
```
ThePotatoGuy:
```suggestion
            if ev.action != EV_ACT_QUEUE:
```

PR #5085, opened 2019-12-09
getVolume fix

### LINKED ISSUE #5073: Music not returning after using Shift+M
After muting the music using Shift+M, the music won't come back on, even after using Shift+M and/or selecting different songs. Other sounds work fine, such as the box highlighting sound, weather, and games. Closing and opening MAS resumes the music, but using Shift+M kills the music again and it doesn't return after using Shift+M again. Issue began after installing 10.4.

ThePotatoGuy:
#5073 

### Cause
**There are three types of volumes:**
* preference volume - mixer slider in settings
* `chan_volume` - channel-specific volume with no known access/modification functions
* `context.secondary_volume` - volume that gets changed when using `set_volume`. for clarity, this will be referred to as channel volume

The `set_volume` function always [set secondary volume](https://github.com/renpy/renpy/blob/6.99.12.4.2187/renpy/audio/music.py#L411).
`getVolume` was changed to return `actual_volume`, which is a combination of `chan_volume` and the slider setting. Thus, music-selector based functions relying on the actual channel volume stopped working.

# Key Changes
* revert `songs.getVolume` back to returning `context.secondary_volume`
* new function `songs.getUserVolume` which returns the preference-based volume.
* new function `songs.setUserVolume` which allows for setting the preference-based volume. the existing `set_volume` functions should be used for normal channel-based volume setting.
* muting music and volume adjustment hotkeys have been changed to affect the volume-slider (preference-based volume) instead of channel volume.

**NOTE:** since we don't do anything with `chan_volume`, the secondary volume and preference-based volume should be enough to manipulate volume levels. 

**NOTE:** as a reminder:
* use `songs.getVolume` and `renpy.<audio/sound/music>.set_volume` to change volumes that the user cannot change
* use `songs.getUserVolume` and `songs.setUserVolume` to change volumes that the user **can** change.

# Testing
* verify muting (Shift+M) mutes and unmutes as it used too. this should affect the slider, but it won't be live (aka muting while the game  menu is open will not update it right away)
* verify volume adjustment hotkeys (+/-) change volume. this should affect the slider, but it wont be live (aka adjusting while the game menu is open will not update it right away)
* verify the aiwfc song (and its d25 lead-up) still works.

### DIFF
--- a/Monika After Story/game/script-holidays.rpy
+++ b/Monika After Story/game/script-holidays.rpy
@@ -2606,7 +2606,7 @@ label monika_aiwfc:
         m 1eksdla "I hope you don't mind, but I prepared a song for you."
         m 3hksdlb "I know it's a little cheesy, but I think you might like it."
         m 3eksdla "If your volume is off, would you mind turning it on for me?"
-        if songs.getVolume("music") == 0.0:
+        if songs.getUserVolume("music") == 0.0:
             m 3hksdlb "Oh, don't forget about your in game volume too!"
             m 3eka "I really want you to hear this."
         m 1huu "Anyway.{w=0.5}.{w=0.5}.{nw}"
@@ -2651,8 +2651,18 @@ label monika_aiwfc_song:
     $ disable_esc()
     $ mas_MUMURaiseShield()
 
-    $ play_song(None, 1.0)
+    # always unmute the music channel (or at least attempt to)
+    # TODO: there should probably be handling for sayori name case.
+    if songs.getVolume("music") == 0.0:
+        if songs.music_volume > 0.0:
+            $ mute_music()
+        else:
+            $ renpy.music.set_volume(1.0, channel="music")
+
+    # save background sound for later
     $ amb_vol = songs.getVolume("backsound")
+
+    $ play_song(None, 1.0)
     $ renpy.music.set_volume(0.0, 1.0, "background")
     $ renpy.music.set_volume(0.0, 1.0, "backsound")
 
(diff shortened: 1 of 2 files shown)


Monika After Story/game/script-holidays.rpy
```diff
@@ -2651,8 +2651,15 @@ label monika_aiwfc_song:
     $ disable_esc()
     $ mas_MUMURaiseShield()
 
-    $ play_song(None, 1.0)
+    # always unmute the music channel (or at least attempt to)
+    # TODO: there should probably be handling for sayori name case.
+    if songs.getVolume("music") == 0.0:
```
multimokia: I'm not sure that setting the user's volume is necessarily good practice as some people do keep it off for a reason (to avoid attention etc)

As well, the dialogue flow from the `monika_aiwfc` event somewhat leaves it in the player's court to do this themselves.

Monika After Story/game/script-holidays.rpy
```diff
@@ -2651,8 +2651,15 @@ label monika_aiwfc_song:
...
+        $ renpy.music.set_volume(1.0, channel="music")
+
+    # save background sound for later
     $ amb_vol = songs.getVolume("backsound")
```
multimokia: We need to get the exact volume of the amb mixer so we can revert back to it later.
```suggestion
    $ amb_vol = songs.getUserVolume("backsound")
```

Monika After Story/game/script-holidays.rpy
ThePotatoGuy: this does not set user volume, it sets channel volume which is now completely not controlled by the user

Monika After Story/game/script-holidays.rpy
ThePotatoGuy: as mentioned in the pr, there is a difference between channel volume and user volume. using set_volume sets the channel volume, which is not controlled by the user. 

the code here sets channel volume so getting the user volume and setting the channel volume to it could change the volume in a way the user cannot revert back.

PR #5091, opened 2019-12-10
Properly remove undo action rules

multimokia:
Undo action rule removal as of 0.10.4 caused a dictionary changed size during iteration traceback due to poor placement of the remove_rule function call.

This fixes that and also runs a check to make sure we don't test on bad data if we have it in there for any reason.

### DIFF
--- a/Monika After Story/game/event-rules.rpy
+++ b/Monika After Story/game/event-rules.rpy
@@ -858,11 +858,11 @@ init python:
                 _start_date = ev.start_date
                 _end_date = ev.end_date
 
-                #We prevent a NoneType addition by removing the rule if any fields here are None.
+                #We return none here
                 if not _start_date or not _end_date:
-                    MASUndoActionRule.remove_rule(ev)
-                else:
-                    MASUndoActionRule.adjust_rule(ev, _start_date, _end_date)
+                    return None
+
+                MASUndoActionRule.adjust_rule(ev, _start_date, _end_date)
 
                 #We're now past the dates and need to undo the action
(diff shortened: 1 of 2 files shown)


Monika After Story/game/event-rules.rpy
```diff
@@ -879,11 +877,24 @@ init python:
             IN:
                 per_rules - persistent dict/list to get rules from
             """
+            remove_list = list()
```
ThePotatoGuy: why are we using a remove list and not just iterating over `per_rules.keys()`?

`.keys` returns a copy of the keys so its fine to use removal with them.

Monika After Story/game/event-rules.rpy
```diff
@@ -879,11 +877,24 @@ init python:
...
+                if ev is not None and should_undo:
                     Event._undoEVAction(ev)
 
+                #We ended up getting a none, so we need to add the ev to a list of ons we need to remove
+                elif should_undo is None:
+                    remove_list.append(ev)
+
+            #Now we go thru the items in remove list and remove their rules
+            for ev in remove_list:
+                MASUndoActionRule.remove_rule(ev)
```
ThePotatoGuy: `check_persistent_rules` takes a dict of rules to evaluate. However, `MASUndoActionRule.remove_rule` removes rules from the persistent dict. This is bad if we wanted to evaluate and clean an arbitrary collection of rules not from the persistent dict.

`remove_rule` shouldn't be used here. This should just remove directly from the given dict of rules.

PR #5095, opened 2019-12-11
Various Fixes

### LINKED ISSUE #5096: [bug] floating islands
seeing the new dialogues of the islands during the rainy and cloudy weather, there is the dialogue that monika says it would be nice to go there with an umbrella etc, but after that it ends the dialogue of when the weather is sunny, with she saying it's a great time for a picnic

### LINKED ISSUE #5097: Monika doesn't count the number of holidays season spent together accurately
Screenshot_1

The line above suggests that this is our second Christmas with her, but I got this dialogue even though in my case, this is actually the third time.  
If counting the number of Holidays spent together is possible, it might be interesting to use a counter to alter this line slightly as to be more accurate ("two whole years have gone by", "three whole years", and so on).

### LINKED ISSUE #5160: Very small bug- Christmas Cookies
Hello! During the Christmas event I ran into a tiny continuity issue.
Monika is reacting to the Christmas Cookies gift as though the ones I gave from last year were recent ("I still haven't finished the last batch!")
It wasn't much of an issue, but I figured it wouldn't hurt to drop here in case others had the same occur.

Hoping all the devs had a lovely holiday!

multimokia:
`monika_mountain` re-randomed if it was already answered, despite it being a player model topic which shouldn't rerandom. I have conditionaled the re-random in the 0_10_4 update script and as well the seasonal prog points.

`monika_aiwfc` had one of its checks changed to a seen_label fro the song (singing bit, not song framework) label, which means the first time dialogue would never be seen, adjusted to a shown_count check of `monika_aiwfc`.

Fixed `if`/`elif` chains for islands dialogue, wrong flows should no longer be appended (fixes #5096, commit is tagged wrong)

Added hist check to `mas_d25_monika_holiday_intro` so she gets the year right (fixes #5097)

`monika_event_clothes_select` now unlocks in `mas_addClothesToHolidayMap()` (which has also been moved to `init 10` as a result). Locks for the selector have been placed in holiday autoloads.

Added update script to fix the 922 events. Combined with the amended undo action rules in #5091, this should forever fix that issue.

Eventlabels adjusted:
```
"mas_bday_surprise_party_hint",
"mas_bday_pool_happy_bday",
"mas_bday_spent_time_with",
"mas_bday_postbday_notimespent"
```

Added actual select dialogue for dress

Fixed a gifting issue where the wrong end date was used (Gifting on D25 meant something under the tree, only would be cleared on nyd through silent react).
Also changed (in retrospect) so that the gift range is from the start of the d25 season itself.
(NOTE: Also removed `mas_isD25Gift()`, replaced usages with `mas_isD25Pre()`)

christmascookies have had their react fixed. The persistent variable now maps to history and will be reset accordingly (so the correct dialogue will show) (Fixes #5160)

Added extra mux type for the hair flower so it removes the flower crown when worn.

Added NoneType checking for gifts whose ev's can be `None`

Added proper handling for jsons which are removed from the install via D25 Gifts

Fixed d25 gift reaction skips

#### For `monika_mountain`:
- Derandom the event via console
- Verify that if `persistent._mas_pm_would_like_mt_peak` is either `True` or `False`, the topic does not re-random via `mas_seasons._pp_spring()` or calling `v0_10_4`

#### For `monika_aiwfc`:
- Verify that with a shown count of 0 for the label, you get the first time dialogue after Monika sings the song and subsequent times Monika sings this, you'll get the second only. (use `pushEvent('monika_aiwfc')` to do this easily)

#### For `monika_event_clothes_select`:
- Verify that the event clothes selector is unlocked when it needs to be (after clothes are added to the map) and locked when needed (no longer having event clothes/via flows etc)

#### Update script:
- Verify 922 events have correct data (Set their conditionals/actions/start_dates/end_dates to None, make sure they now have the correct data)

#### New Gift Range:
- Verify gifting works properly as it *should have* before.
- Verify that gifting works from the start of the season as well, but only if the decorations are out (so it makes sense)

#### Christmascookies react
- Verify that going to next year, `persistent._mas_d25_gifted_cookies` is `False`, and that giving the cookies for the first time in the season results in the correct dialogue

#### Extra mux:
- Verify that wearing the pink flower with the flower crown on removes the flower crown

#### Removed jsons christmas gifting
- Gift a json sprite before the 25th of December
- Exit MAS and make the sprite invalid (remove an asset image that is NOT a thumb or delete/rename the json)
- Load in on the 25th, Monika shouldn't react to it, and once the rest of the reactions are complete, Monika shouldn't have a gift under the tree.

#### Fixed D25 gift reactions skips:
- Verify that not saying `Merry Christmas, Monika` in time during `mas_d25_monika_christmas` still has Monika react to the gifts under the tree


AsuraShun:
You OK there?

Issue #5099, opened 2019-12-11
[Suggestion] Nevermind when in bookmarks


PollenTheCat:
When in the Bookmarks section of Talk, when you press Nevermind, instead of going out of the whole Talk menu, it could send you back to the first section of the Talk menu.

jmwall24:
We can do that. If we do, we'd add likely add it to all the `Talk` menus, since right now, if you click `That's enough for now` in the repeat convo or hey. monika menus, it takes you out of `Talk` completely. So going to open this to discussion, do people want these 'back out' options to take you to the main `Talk` menu, and then you'd have to hit `Nevermind` on the main menu to get out of `Talk` or do people like how it currently is, with them taking you out of `Talk` completely and having to hit `Talk` again to go to a different option.

ThePotatoGuy:
[closing as fixed]

PR #5105, opened 2019-12-13
Return to Talk menu

jmwall24:
#5099

Currently, categories in the `Talk` menu and topics that solely use the general scrollable menu will return you to the idle loop when clicking the back-out (typically "Nevermind") option. This has them return to the main `Talk` menu instead, so the player can find what they were looking for without having to click `Talk` again.

It also removes the lead-in dlg (usually "Sure") as well as the lead-out dlg when clicking the back-out option to aid transitions. In addition, `monika_sing_song_pool` was changed to match `monika_short_stories_menu` via losing it's initial menu that asked what type of song you want to hear, since the song menu itself allows the player to switch types.

## Topics/menus affected:

- `Unseen`

- `Bookmarks`

- `Hey, [m_name]...` (and all categories therein)

- `Repeat Conversation` (and all categories therein)

- `Goodbye`

- `monika_sing_song_pool`

- `monika_short_stories_menu`

- `mas_topic_rerandom`

- `monika_showpoem`

- ` monika_change_weather`

- selector topics also lost their lead-out dlg due to being unnecessary.

## Testing

- for every affected menu/topic above, verify the back-out options take you back to the main `Talk` menu and also verify the other options in those menus work as intended.

This is a relatively simple PR to test, but am requesting extra testing just because this affects a fair amount of things, so want to make sure this has extra eyes.


Monika After Story/game/event-handler.rpy
```diff
@@ -2104,6 +2104,9 @@ label call_next_event:
             if "love" in ret_items:
                 $ mas_ILY()
 
+            if "prompt" in ret_items:
```
ThePotatoGuy: can you put this after the quit branch? I think quit should have a higher priority.

Monika After Story/game/event-handler.rpy
```diff
@@ -2467,8 +2469,8 @@ label mas_bookmarks:
         bookmarkedlist = mas_get_player_bookmarks()
 
         bookmarkedlist.sort()
-        remove_bookmark = (mas_getEV('mas_topic_unbookmark').prompt, mas_getEV('mas_topic_unbookmark').eventlabel, False, False, 20)
-        return_prompt_back = ("Nevermind.", False, False, False, 0)
+        remove_bookmark = (mas_getEV('mas_topic_unbookmark').prompt, 'mas_topic_unbookmark', False, False, 20)
+        return_prompt_back = ("Nevermind.", "prompt", False, False, 0)
```
ThePotatoGuy: hold up, line 2480 would eval to true now since the return value contains something. False works fine if you change that if statement a little like:
```python
if topic_choice:
    # same stuff
    return True

return False
```

PR #4572, opened 2019-08-28
A new dialogue (line 12129 to 12178)

Marth83:
Wew, hope it works now

PR #5146, opened 2019-12-24
Add ""_mas_pm_read_poetry"

Marth83:
cf #4572 ("Poetry" topic)


multimokia:
We can just add this to your existing PR. There's no need to open a separate PR for this.

Marth83:
I opened the wrong branch, I'll try again

PR #5169, opened 2019-12-30
Restart Blacklist Prop in AddEvent

Legendkiller21:
Adds a `restartBlacklist` prop in `addEvent` allowing us to add an event in the restart blacklist without having to edit the list itself.


ThePotatoGuy:
wait i still want to test this later

multimokia:
I have tested this myself

PR #5197, opened 2020-01-03
S-type ribbons

Velius94:
This add's velius's ribbons as a spritepack.

These ribbons will be referred to as `s-type` and use the `s-type-ribbon` acs type.

==== og msg below ====

Changelog:
- Redone every variant of ribbon from scratch + retuned color schemes and positioning.
- Updated thumbnails.


ThePotatoGuy:
I think the appearance of these are too different from the stock Act 3 ribbons. While we have taken creative liberties with hair, facial expressions, and blazer color, this is too far in terms of changes to stock items.

At best, we can accommodate this as a submod, or spritepack under a different type of ribbon (maybe shorter-ribbon? not sure).

ThePotatoGuy:
i will transition these into spritepacks.

multimokia:
Transition has been made, this is now awaiting testing.

PR #5201, opened 2020-01-05
Player bday update

jmwall24:
This helps the player bday event be more sustainable, offering a little variety and the ability to easily select a new unseen poem every year. The changes to the autoload only affect the normal path where we would have gotten to the open-door surprise greet before.

## Adds

- `persistent._mas_player_bday_saw_surprise` a new var that stores if we got the surprise party variant of the birthday event in a given year.

- `mas_player_bday_greet` a new greet that, depending on circumstances, is a fake open-door greet where Monika already has everything set up, or a normal greet that starts the bday party.

- `getRandomPoem` a `mas_poems` func that gets a random poem of a specified category. can be used to get an unseen poem to facilitate the player getting a new poem they have yet to receive.

## Testing
- verify on your first birthday party with Monika (or if you've only gotten the party prior via `mas_player_bday_ret_on_bday` or `mas_player_bday_no_restart`) you get the special open door greet this year if you load in normally on your bday.

- verify if you ruin the surprise (open the door before knocking), next year, if you load in normally, you do not get the open door greet, but instead get the `else` path of `mas_player_bday_greet`.

- verify the bday poems don't repeat until you've seen them all. may want to add a few dummy poems and just `call mas_showpoem(mas_poems.getRandomPoem("pbday"))` via console to confirm this without time travelling multiple times.

- verify the few dialogue alternatives added make sense based on if it's your first or subsequent bday party spent with Monika.


Monika After Story/game/script-holidays.rpy
```diff
@@ -4402,11 +4405,44 @@ label mas_player_bday_autoload_check:
...
+
+        python:
+            # first we determine if we want to run a surprise greeting this year
+            this_year = datetime.date.today().year
+            years_checked = range(this_year-10,this_year)
+            times_ruined = 0
+            surp_int = 3
+
+            for y in years_checked:
+                if y in mas_HistVerify("player_bday.opened_door",True)[1]:
```
ThePotatoGuy: so `mas_HistVerify` can be run with years. either:
* `mas_HistVerify("player_bday.opened_door", True, *years_checked)`

OR
* `mas_HistVerify_k(years_checked, True, "player_bday", "opened_door")`

the 2nd element of the returned tuple is list of years that matched verification so you can just do a `len` on the returned list as times_ruined instead of doing an extra loop here.

Monika After Story/game/zz_poems.rpy
```diff
@@ -103,6 +103,30 @@ init 11 python in mas_poems:
...
+            unseen:
+                whether or not we only want unseen poems
+                defaults to True 
+
+        OUT:
+            A random poem
+        """
+        unseen_poem_amt = len(getPoemsByCategory(category, unseen=True))
+        total_poem_amt = len(getPoemsByCategory(category, unseen=False))
+        poem_num = renpy.random.randint(0,total_poem_amt-1)
```
ThePotatoGuy: this effectively rolls twice when getting an unseen poem. better to not waste a roll by changing this a little:
```python
sel_poem_len = total_poem_amt-1
if unseen:
    if unseen_poem_amt > 0:
        sel_poem_len = unseen_poem_amt-1
    else:
        unseen = False
poem_num = renpy.random.randint(0, sel_poem_len)
```

PR #5220, opened 2020-01-13
Update Fix

multimokia:
*oooooooooooooooooooooooooooooooooooooooooof*

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -6756,8 +6756,11 @@ label monika_meditation:
 
     return "derandom|love"
 
-# do you like orchestral music
+#Do you like orchestral music?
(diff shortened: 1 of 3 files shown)


ThePotatoGuy:
the unbookmarking topic erase was after the del lines and should be added to the update script. erasetopic should be safe to run twice

ThePotatoGuy:
as I mentioned to @jmwall24 , this needs to verify if the update code that may not have run is safe to run twice, or if not, is stopped from doing so.

multimokia:
The code should be safe to run twice, in the end it'll wind up setting the same vars to the same values. No issues in this running twice.

(Setting up the conditions to avoid the check winds up being more hassle, and since just letting it happen anyway doesn't mess it up, we're fine)

PR #5238, opened 2020-01-17
Better Fun Facts

multimokia:
Fun facts now prioritize unseen fun facts first. Fact types (good/bad) no longer exist as bad facts just didn't really make sense (and aren't really in character for Monika since ideally she'd make sure her facts are straight)

Also added a new fact stolen from the yearolder mood.

## Testing:
- If you have any fun facts seen on your persistent, verify that when calling `v0_10_6`, and then running `mas_fun_facts.getUnseenFactsEVL()` returns a list without any of the seen fun facts in it.
- Verify that you must have seen all fun facts once before you can see them again via the random selection.


Monika After Story/game/script-fun-facts.rpy
```diff
@@ -39,28 +72,21 @@ label monika_fun_facts_open:
...
 
     python:
-        #Determines if it is a bad fact, 10% chance.
-        if renpy.random.randint(1,100) <= 10:
-            list_facts = [
-                eventlabel
-                for eventlabel, event in store.mas_fun_facts.fun_fact_db.iteritems()
-                if store.mas_fun_facts.TYPE_BAD in event.category
-            ]
+        if mas_fun_facts.hasUnseenFacts():
```
jmwall24:
```suggestion
        if len(mas_fun_facts.getUnseenFactsEVL())>0:
```
Don't really need a separate function for this.

PR #5256, opened 2020-01-21
Brbs no longer in unseen

multimokia:
BRBs should no longer appear in unseen nor have the italicized prompt if they're unused

Done because as we add more BRBs, they may not apply to everyone, so keeping them clear of unseen would be good.

Also adjusted import_ddlc to be a lot more efficient and only do merges

## Testing:
- Verify that on a persistent with unseen brbs, they no longer show up in unseen and no longer have an italicized prompt.
- Verify that importing ddlc still works properly (Correct data is transferred)


multimokia:
Adding don't merge as requires some changes to `import_ddlc.rpy`

PR #5258, opened 2020-01-21
Default ex_props to None

Legendkiller21:
Currently, empty dicts in the constructors of objects will keep the same instance over all objects created from the constructor. Meaning if any of them are changed, all of them change.

This defaults the `ex_props` of the constructors to `None` and sets it to an empty dict if it's `None` in the MASSpriteBase constructor.

## Testing
- Verify that only the intended `ex_props` are applied to the objects.
- Verify that everything still works.


ThePotatoGuy:
interesting, but `ex_props` should never change. what is the desired use case?

PR #5263, opened 2020-01-23
Added note file reading feature

scpketer:
When I've seen this 'Surprises' topic for the first time and opened her note (I was on `love` affection level by that moment) I'd thought it would be nice if Monika could read the file back and see if player responds on her question 'I want to be yours forever, so would you be mine?'

The example is shown below:
image

Comment if I need to add some other 'responses'.
Or drop and reject this PR if you want.

Detection is regexp-powered, so it's really flexible. I've added just 'I would' with optional dash preceding, but I think something else like 'I would not' or related could suit, too (just w/ some other responses)

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -5472,6 +5472,20 @@ label monika_surprise:
         m 1hua "Alright!"
         m 1eua "What are you waiting for? Go take a look!"
         m "I wrote it just for you~"
+
+        # Check if user has appended 'I would' to 'would you be mine?' in note file
+        # NOTE: dependent on note content in mas_suprise()
+        if mas_isMoniLove():
+            python:
+                from re import search
+                with open(renpy.config.basedir + "/My one and only love.txt", "r") as note:
+                    text = note.read()
+                    if search("would you be mine\\?(?:(?:\\s*-\\s*)|(?:\\s*))[iI] would.?", text):
+                        renpy.show("monika 1suo")
+                        renpy.say(m, "Ah{cps=*0.25}...{/cps} {w=1}[player]{cps=*0.25}...{/cps}")
+                        renpy.show("monika 1dkbsa")
+                        renpy.say(m, "[player]{cps=*0.7}...{/cps} {w=1}I'm so glad that you mean it...")
+                        
         m 1ekbsa "I really do truly love you, [player]~"
 
     # Normal and Happy


multimokia:
So the issue with this is that no matter what you do, you *cannot* reasonably expect to catch every possible response.

While it's a good thought and something which would be nice, many people simply wouldn't just respond in the two words you gave. Some might do it in one, some might write an entire letter back.

scpketer:
> So the issue with this is that no matter what you do, you _cannot_ reasonably expect to catch every possible response.
> 
> While it's a good thought and something which would be nice, many people simply wouldn't just respond in the two words you gave. Some might do it in one, some might write an entire letter back.

Yeah, but isn't that just something like menus gives in some topics? Player can't write their own sentence back as an answer.

multimokia:
I'm not sure what you mean. What I mentioned before is the same very reason we don't let the player type. It's not possible to account for every response.

scpketer:
I know, I know there's no way we can catch every response. But some phrases? Player gives predefined gifts, too.

multimokia:
> I know, I know there's no way we can catch every response. But some phrases? Player gives predefined gifts, too.

Okay, but think about this. There's likely going to be people who write an *entire* letter back. Putting their heart and soul into it. They would get *nothing* for it, whereas people who simply put two words down get that.

That just doesn't seem right.

AsuraShun:
I don't know if it would help but there's an app called ai dungeon, it's a text based game with seemingly endless possibilities in what you can type and it responds accordingly, maybe it could somehow be used as a reference

scpketer:
> > I know, I know there's no way we can catch every response. But some phrases? Player gives predefined gifts, too.
> 
> Okay, but think about this. There's likely going to be people who write an _entire_ letter back. Putting their heart and soul into it. They would get _nothing_ for it, whereas people who simply put two words down get that.
> 
> That just doesn't seem right.

Alright, maybe I had to point that out from the very beginning, but that's not something I'd like to expose to player; I'd rather leave it as an 'undocumented feature' or an 'easter egg'.

Maybe you're right, and it's actually not a good idea.

multimokia:
In my opinion, you either do the full thing, or you don't at all for this scenario.

These are words that don't just get taken at face value, they're words that carry raw feeling for people who do choose to respond. As such, we cannot just ignore the people who write otherwise, as it's almost like a rejection in a sense, and I just don't feel like that's the right thing to do.

We also must consider that there may be people who would abuse just assuming anything written was accepted, so it needs a way to filter the text and figure out whether or not it's good or bad and comment accordingly. It's a much bigger task than the scope of this, though.

KarmaCreations:
One way this could maybe be done is, rather than having the player respond in the actual file, Monika could verify that the player read it and then ask something along the lines of "Well, would you?" and the player has the option of saying yes or no.

scpketer:
> One way this could maybe be done is, rather than having the player respond in the actual file, Monika could verify that the player read it and then ask something along the lines of "Well, would you?" and the player has the option of saying yes or no.

Brilliant, thanks! I'll work on it.

Booplicate:
~~I thought it was one of questions that don't need to be answered.~~

scpketer:
Well, uh... I'm even more confused now.

PR #5265, opened 2020-01-24
Changed after-kiss expression to soft eyes

fromulus:
Very small and simple change, this changes the expression when she zooms out right after a kiss to a new one using the soft eyes.

That way those eyes are actually used, and I personally feel that this new expression fits better than the smug face she currently does.


multimokia:
Tested this myself (after adding the sprite locally), and personally I'm not too fond of how this looks versus the current expression used in the `mid_exp`

As well, with some internal discussion we also don't really feel that it needs to be changed at all.
Closing as a result.

PR #5291, opened 2020-02-03
TimePastSince in the event class

multimokia:
If we were to pass in a `event.last_seen` into `mas_timePastSince`, we couldn't reliably use `.date()` without some extra work (and can't do that directly in an if statement). 

This pull request adds two new functions into the `Event` class for checking if it has been x amount of time since we've last seen the event. First function handles `date`s, the second in precise `datetime`s.

## Testing:
- Verify that you get proper results from using the two new functions with a both, a `datetime.date` and `datetime.datetime` passed into each
  - Verify that `timePastSinceLastSeen_d` is precise to new day only, rather than specific hours/minutes/seconds/microseconds
  - Verify that `timePastSinceLastSeen_dt` is precise to hours/minutes/seconds/microseconds
- Verify no crashes

### DIFF
--- a/Monika After Story/game/definitions.rpy
+++ b/Monika After Story/game/definitions.rpy
@@ -5198,7 +5198,7 @@ init 2 python:
             timekeeper - variable holding the time since last event
             _now - time to check against (Default: None)
         """
-        return mas_timePastSince(timekeeper, datetime.timedelta(days=1), inexact, _now)
+        return mas_timePastSince(timekeeper, datetime.timedelta(days=1), _now)
 
 # Music
 define audio.t1 = "<loop 22.073>bgm/1.ogg"  #Main theme (title)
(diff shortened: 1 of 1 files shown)


Monika After Story/game/definitions.rpy
```diff
@@ -5153,17 +5158,23 @@ init 2 python:
         if not isinstance(timekeeper, datetime.datetime):
             timekeeper = datetime.datetime.combine(timekeeper, datetime.time())
 
+        #If we converted from datetime.date to datetime.datetime, this is already inexact
+        #so we only check this if we'd originally passed in a timekeeper with a datetime.datetime
+        elif inexact:
+            timekeeper = timekeeper.replace(hour=0, minute=0, second=0, microsecond=0)
```
ThePotatoGuy: if a `date` is passed in instead of a `datetime`, we already set the time portion to 0, so we could just pass in the date portion of a datetime rather than add an extra parameter. 

I don't see the usefullness of the inexact parameter. If we want inexact, pass in dates. If we want precise, pass in datetimes.

ThePotatoGuy:
Also update ur pull request description

PR #5301, opened 2020-02-07
Submod Framework

multimokia:
Introducing the submod framework
~~Finally making use of this project~~
(Big thanks to @Booplicate for helping me get this working)

# Adds:
- Submod class
- Submod settings screen

Finally, a system where submods can check for dependencies, have a general settings menu, and have their own update scripts.

## Testing:
first, make a new rpy, just put this in for the sake of example:
```renpy
screen submod_screen():
    vbox:
        box_wrap False
        xfill True
        xmaximum 1000


        hbox:
            style_prefix mas_ui.cbx_style_prefix
            box_wrap False

            textbutton _("Switch setting #1") action NullAction()
            textbutton _("Switch setting #2") action NullAction()
            textbutton _("Switch setting #3") action NullAction()
```
This allows us to generically add the same screen to the new submods settings screen.

- Make an rpy file and initialize a `Submod` object in it. Storing it doesn't matter as you shouldn't need to interact with it after.
```python
init -990 python in mas_submod_utils:
    Submod(
        "author",
        "generic name",
        "a description",
        "0.0.0"
    )
```
(this will do for that instruct)
- Verify that update scripts work for submods individually (pass in a dict similar to our update structs (follow the notes in `__init__` if you're confused. It's basically a {"from version": "to version"} system and it follows a chain up. Start with the version you initialized the submod object with initially)
```python
{
    "<author>_<name>_v<version>": "<author>_<name>_v<version>"
}
```
- Say goodbye, increment the version number, and load in again. The label stored in the value should now be seen.
- Assuming you haven't passed in `"submod_screen"` into `settings_pane`, verify that the `Submods` option does not show up in the settings screen (on the side with all the tabs)
- Say goodbye, pass `"submod_screen"` into the `settings_pane` param of the `Submod` and load in again. The `Submods` tab should be visible, selecting it should show that screen.
- Add a dependency to the submod (you can make up the name) pass in a value for the dependency in the following form: ("minimum_version", "maximum_version") (Maximum can be none)
- If you do not initialize another submod object with the required fields of the dependency (not installed/not recent enough version/too new version) verify that you get the correct tracebacks for each.

NOTE: This is likely confusing as hell, so ask me if you need help with this


Monika After Story/game/zz_submods.rpy
```diff
@@ -0,0 +1,307 @@
...
+                        author name: MonikaAfterStory,
+                        submod name: Example Submod
+                        submod vers: 1.2.3
+
+                    becomes:
+                        label monikaafterstory_example_submod_v1_2_3(version="v1_2_3")
+            """
+            #First make sure this name us unique
+            if name in submod_map:
+                raise SubmodError("A submod with name '{0}' already exists.\n".format(name))
```
ThePotatoGuy: I think if this type of error is reached, we should just log the error and quietly ignore the submod. Especially since the likely user to encounter this would be a player installing a submod rather than the submod creator.

Monika After Story/game/zz_submods.rpy
```diff
@@ -0,0 +1,307 @@
...
+                    - True if the version number has incremented from the persistent one
+                    - False otherwise
+            """
+            old_vers = persistent._mas_submod_version_data.get(self.name)
+
+            #If we don't have an old vers, we're installing for the first time and aren't updating at all
+            if not old_vers:
+                return False
+
+            old_vers = map(int, old_vers.split('.'))
```
ThePotatoGuy: this should have exception handling and clear bad data from the persistent just in case.

Monika After Story/game/zz_submods.rpy
```diff
@@ -0,0 +1,307 @@
...
+            """
+            #Iter thru all submods we've got stored
+            for submod in submod_map.itervalues():
+                #If it has updated, we need to call their update scripts and adjust the version data value
+                if submod.hasUpdated():
+                    submod.updateFrom(
+                        "{0}_{1}_v{2}".format(
+                            submod.author,
+                            submod.name.replace(' ', '_'),
+                            persistent._mas_submod_version_data[submod.name].replace('.', '_')
```
ThePotatoGuy: should use `.get` here in case of missing data

Monika After Story/game/zz_submods.rpy
```diff
@@ -0,0 +1,318 @@
...
+        def _checkUpdates():
+            """
+            Checks if submods have updated and sets the appropriate update scripts for them to run
+            """
+            #Iter thru all submods we've got stored
+            for submod in submod_map.itervalues():
+                #If it has updated, we need to call their update scripts and adjust the version data value
+                if submod.hasUpdated():
+                    submod.updateFrom(
+                        "{0}_{1}_v{2}".format(
```
ThePotatoGuy: is that `v` supposed to be in front of the `{2}`?

Monika After Story/game/zz_submods.rpy
```diff
@@ -0,0 +1,318 @@
...
+                name - name of the submod to get
+
+            OUT:
+                Submod object representing the submod by name if installed and registered
+                None if not found
+            """
+            return submod_map.get(name)
+
+        @staticmethod
+        def _checkVersions(curr_vers, comparitive_vers, is_current=True):
```
ThePotatoGuy: so a comparison function like this where we want to know if something is greater/less/equal, is normally created as a `compareTo` method that returns 1,0, or -1 if the left arg is greater than, equal, or less than the right arg. I would restructure this function to be a `compareTo` style function so `is_current` is not needed.

Also I would make this **not** a static method (and have the "left arg" be considered itself) since all the places that this function is used refers to a submod object anyway. If you forsee the using the version checker for non submod-related version checking, then make the meat of the version check in `mas_utils` and have a class function here that calls the meat.

Monika After Story/game/zz_submods.rpy
multimokia: Yeah, that form is mainly to enforce a similar formatting to our update labels (`v0_10_x` for example) but also adding the author + name to it for differentiation and avoidance of overlap

Monika After Story/game/zz_submods.rpy
multimokia: Will do.

Monika After Story/game/zz_submods.rpy
```diff
@@ -0,0 +1,299 @@
...
+
+            try:
+                old_vers = map(int, old_vers.split('.'))
+
+            #Persist data was bad, we'll replace it with something safe and return False as we need not check more
+            except:
+                persistent._mas_submod_version_data[self.name] = Submod.FB_VERS_STR
+                return False
+
+            return self.checkVersions(old_vers) == 1
```
ThePotatoGuy:
```suggestion
            return self.checkVersions(old_vers) > 0
```
-1, 0, 1 might be what it returns, but the norm is that compareTo-style functions should not be assumed to return -1 or 1. It's always less than or greater than 0.

ThePotatoGuy:
@multimokia add a wiki page for this (does not have to be done before release)

multimokia:
Wiki updates are complete.

authorofbelethrun:
so in a nutshell all this does is just makes it easier for people who want to mod the mod am i correct...im not fluent in this language (hell i am more fluent in c# then this XD ) so i just need an understanding

Issue #4456, opened 2019-07-17
option to stay with monika when sick


ThatMonkeyBiz:
I'm currently down with a case of Strep Throat. I decided to go to Monika, maybe for a little comfort, I'm not sure, but it brought up the idea, what if when you insisted to stay with Monika when you were sick, the dialogue would change a bit? I don't know if this is something that has been suggested already, but it would be very nice if Monika might be caring towards the sick.

lunulae:
I could see a few things that could be done with this. Maybe if you've told her you're feeling sick in a particular session, she'd have a different response for things like "I'm going to go to sleep." I'd maybe add a little teasing dialogue if you ask for a kiss while sick too. 

I'm not really sure how to do conditional dialogue but I am considering trying to give this a go :D

ThePotatoGuy:
[closing until we look into it]

PR #5333, opened 2020-02-13
Stay with Monika when sick and expanded options

Mugnika:
based on issue #4456


jmwall24:
(ignore me for now, i totally wasn't on the wrong PR)

PR #5336, opened 2020-02-14
Telling Monika you are going to draw

masterliamy:
I think it'll be a good idea to add to the "Be right back" category in the game an option to tell Monika that you are going to draw for a while


multimokia:
This is not the place to make suggestions. Please make them on the issues page.

PR #5341, opened 2020-02-15
Added Indonesian Language Greeting

yaziddio:
I wonder since there's a Latin and Italian greetings, so I made the Indonesian one, feel free to change it if there's something wrong with it


jmwall24:
@yaziddio you didn't need to close this, i removed this part from the other PR since it should only be in this PR

Issue #5304, opened 2020-02-08
0.10.7 typos


ThePotatoGuy:
24

Necrocytus:
![image](https://user-images.githubusercontent.com/38022925/74079291-d3114f00-49fa-11ea-9a91-cba26f0ee9b1.png)

fromulus:
![Screenshot_1](https://user-images.githubusercontent.com/45933038/74596278-5101d580-504d-11ea-8523-6b5197d30c82.png)

Finally **being** able to

proudust:
https://github.com/Monika-After-Story/MonikaModDev/blob/da2102baeb135b3caea8194a33da3a9be437af60/Monika%20After%20Story/game/script-farewells.rpy#L540

`m 6ckc "..."`

AsuraShun:
when you're so ancient the word evolves into anicent

Necrocytus:
![image](https://user-images.githubusercontent.com/38022925/78805436-0767a400-7987-11ea-89b2-aff5d99a16ea.png)
Missing apostrophe in Im

jmwall24:
It's actually there, it's just touching the I due to that font being a little wonky

jmwall24:
Figured out how to make it look better though, so it's been changed.

PR #5342, opened 2020-02-15
0.10.7 typos

MisterSimple:
#5304


jmwall24:
Make sure when we resolve conflicts that for the one in compliments we use the new var for the thanks_quip and not the old so we don't crash.

PR #5374, opened 2020-02-24
Hero song

jmwall24:
This adds the song 'Hero' by Mariah Carey as a song Monika can sing to the player. It includes a short version with inspirational dialogue after that is dependant on the `persistent._mas_pm_love_yourself ` variable. It also includes the full song as a long version. It should be a nice motivational boost for the player.

For testing just ensure the song plays randomly and then derandoms when done, and then unlocks itself and its long version in `Hey, [m_name]...>Music>Can you sing me a song`


AsuraShun:
For a moment I thought this was about opm

PR #5389, opened 2020-02-29
Scrolling doesn't progress PauseDisplayable

multimokia:
Currently, during holdme, the pausedisplayable will accept scrolling up/down as an interaction to progress from holding to the reactions.

This PR fixes that behaviour to make it consistent with all other dialogue, making it click-to-progress only.

## Testing
Verify that during holdme, scrolling up or down doesn't progress it to the reactions.

### DIFF
--- a/Monika After Story/game/screens.rpy
+++ b/Monika After Story/game/screens.rpy
@@ -2429,7 +2429,7 @@ init python:
             return renpy.Render(width, height)
 
         def event(self, ev, x, y, st):
-            if ev.type == pygame.MOUSEBUTTONDOWN:
+            if ev.type == pygame.MOUSEBUTTONDOWN and ev.button not in [4, 5]:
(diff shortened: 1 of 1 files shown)


Monika After Story/game/screens.rpy
```diff
@@ -2429,7 +2429,7 @@ init python:
             return renpy.Render(width, height)
 
         def event(self, ev, x, y, st):
-            if ev.type == pygame.MOUSEBUTTONDOWN:
+            if ev.type == pygame.MOUSEBUTTONDOWN and ev.button not in [4, 5]:
```
ThePotatoGuy:
```suggestion
            if ev.type == pygame.MOUSEBUTTONDOWN and ev.button not in (4, 5):
```
use tuple if it doesnt need to change

Issue #5380, opened 2020-02-26
[suggestion] Monika react to Livestreams (with detail)


WingdingsMaster:
<With the windows react feature, you could watch a video or live stream, Perhaps if MAS is open, could you have Monika ask if she could watch it with you?>

lunulae:
Currently she reacts to watching things on Youtube, a similar reaction could definitely be made for streaming sites.
I know of Twitch, but are there other sites you know of that are commonly used for this? That would be useful information for writing up the window reacts.

Naturally, she wouldn't yet be able to have unique lines for what kind of stream it was or what was happening (our technology isn't quite there yet :P)

multimokia:
I guess some would be easier to catch, like those 24/7 playlists which all have a constant title. Or the Lazy Sunday Playlist, etc.

ThePotatoGuy:
[closing as implemented]

PR #5398, opened 2020-03-06
Twitch.tv Window React

lunulae:
Requested in #5380 
I wasn't sure what other livestream services people use, so I only wrote up one for Twitch. I also wasn't sure if I used the right tag for the window detection (just "twitch" rather than "twitch.tv") so please feel free to change it if the full url would work better.

(I haven't made a window reaction before, and I mostly tried to use the Youtube one as a template here.)

### DIFF
--- a/Monika After Story/game/zz_windowreacts.rpy
+++ b/Monika After Story/game/zz_windowreacts.rpy
@@ -706,7 +706,7 @@ label mas_wrs_twitch:
         m_name,
         [
             "Watching a stream, [player]?",
-            "Do you mind if I watch with you?"
+            "Do you mind if I watch with you?",
             "What are we watching today, [player]?"
         ],
         'Window Reactions'
--- a/Monika After Story/game/zz_windowreacts.rpy
+++ b/Monika After Story/game/zz_windowreacts.rpy
@@ -688,3 +688,31 @@ label mas_wrs_reddit:
     if not wrs_success:
         $ mas_unlockFailedWRS('mas_wrs_reddit')
     return
+
+init 5 python:
+    addEvent(
+        Event(
+            persistent._mas_windowreacts_database,
+            eventlabel="mas_wrs_twitch",
+            category=['twitch'],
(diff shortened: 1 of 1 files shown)


multimokia:
Potential issue I see with this is that there's a potential ambiguity here. (We can't use `twitch.tv` since that's not going to be in the window handle)

But it means it's also going to trigger if I searched up something like `"Why do birds twitch their wings?"`

lunulae:
That's a really good point, and I'm unsure if there's a good solution for it. The only thing I can see that distinguishes between streams in the Twitch window handle is the individual stream titles themselves.

No worries if this can't be used for that reason, I'd rather not make any awkward false positives.

multimokia:
Yeah, sadly with the way windowreacts work, it's based *purely* off the name of the window handle. So for Twitch in google chrome, if I was watching a stream by myself for example, it'd show up as `Multimokia - Twitch - Google Chrome`. Which just doesn't give us enough information, since they're supposed to be generic in order to actually work with programs other than just your browser.

Not really your fault at all, but merely a limitation based on the information we get from the site. If Twitch actually used the `Twitch.tv` bit in the url it'd be fine (even though there'd be some false positives, at least it'd be related to the actual service and can be accounted for in dialogue)

Monika After Story/game/zz_windowreacts.rpy
```diff
@@ -688,3 +688,31 @@ label mas_wrs_reddit:
     if not wrs_success:
         $ mas_unlockFailedWRS('mas_wrs_reddit')
     return
+
+init 5 python:
+    addEvent(
+        Event(
+            persistent._mas_windowreacts_database,
+            eventlabel="mas_wrs_twitch",
+            category=['twitch'],
```
jmwall24:
```suggestion
            category=['-twitch'],
```
I did some testing on this, and using `-twitch` allows it to catch it when you are on any individual streams ('think user - Twitch' for the window name) without it also catching any window that just had `twitch` on it.

PR #5408, opened 2020-03-13
hotkey interruption fix

### LINKED ISSUE #5406: [Bug] Exception on hotkey press during first few seconds of game startup
~~sure, go ahead, tell me this one is intentional~~

**Steps to reproduce**

1. Start game
2. Once its window has appeared, hit any hotkey
3. Observe the exception (screenshot follows)

image

EDIT: Experimentally I found out that sometimes this exception screen does not appear immediately upon hotkey press, but rather on 'Talk > Nevermind' button press; exception seems to be the exact same one.

ThePotatoGuy:
fixes #5406 

Cause of above issue was hotkey keymaps being set and enabled before spaceroom was loaded and greeting shown. Hotkeys can cause jumps, which breaks startup flow and allows vars set in `ch30_preloop` to remain undefined.

# Key Changes
* disabled hotkeys from start until the all topics/greeting in event list have ended **except** for music menu in cases where greetings skip visuals (monikaroom/hair down/etc..)
* those hotkeys are renabled after returning to `ch30_loop` from topics (`call_next_event`) if not in idle mode
* Rearranges some of visual setup startup to be less 🍝 
* Sets certain global vars on init to avoid dt-None operations

# Testing

## issue
1. Launch game
2. As soon as the game window appears (even with nothing in the window), spam hotkeys. 
3. Verify that no hotkey is responsive until **after** the greeting + any initial topics 
4. Repeat steps 1-3 but try clicking on the hotkey buttons (clicking in the area they would load). They should also not be responsive until **after** the greeting + any initial topics.
5. Additionally, verify no datetime + None operation crash after using a hotkey.

The above steps should also be tried with:
* returned home greeting
* open door greeting
* hairdown greeting

## regression
* verify that when in idle mode, hotkeys should be disabled and the "Play" button should be disabled.
* verify intro flow enables hotkeys/buttons as usual (after greeting + initial topics)

### DIFF
--- a/Monika After Story/game/script-affection.rpy
+++ b/Monika After Story/game/script-affection.rpy
@@ -2375,7 +2375,6 @@ label mas_affection_yesapology:
     m 1duu "Thank you for putting my heart at ease~"
     show monika 1esa
     $ mas_DropShield_core()
-    $ set_keymaps()
     jump ch30_preloop
 
 label mas_affection_apologydeleted:
--- a/Monika After Story/game/script-ch30.rpy
+++ b/Monika After Story/game/script-ch30.rpy
@@ -926,6 +926,9 @@ label ch30_main:
     # set session data to startup values
     $ store._mas_root.initialSessionData()
 
+    # skip weather
+    $ skip_setting_weather = True
+
     # lastly, rebuild Event lists for new people if not built yet
     if not mas_events_built:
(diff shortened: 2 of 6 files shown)


ThePotatoGuy:
@multimokia @Legendkiller21 ready for review/testing again

PR #5416, opened 2020-03-16
Slow down the intro phase

multimokia:
Intro phase had a habit of machine-gunning new unlocks for games and such, which made the mod feel like it ran out of content significantly faster.

This drastically slows down the game unlocks. Additionally also cleans up the logic for the gender set topic.

Values are open to adjustment.

-- EDIT:
**With the new changes to exp, unlocks should be as follows:**
- Chess should unlock after roughly a week with Monika
- Hangman roughly after about 12 days
- Piano, roughly a month (must also be Aff+ with Monika)

# EXP changes
* XP is now based on daily playtime.
* games and topics do not give xp anymore
* the XP to lvl conversion has changed to be a static 420xp/lvl
* the storage mechanism for XP has changed to only current level and XP to next level.
     * `persistent._mas_xp_lvl` - number of levels
     * `persistent._mas_xp_tnl` - amount of xp To Next Level
     * `persistent._mas_xp_hrx` - number of hours the user has spent with Monika today. 

## XP daily rate
1. the first 2 hours are applied at a rate of 60xp/hr
2. each subsequent 2 hours are applied at a rate divided by a factor of 2. aka:
    * 2-4 hours - 30xp/hr
    * 4-6 hours - 15xp/hr
    * 6-8 hours - 7.5xp/hr
    * 8-10 hours - 3.75xp/hr
    * 10-12 hours - 1.875xp/hr
    * 12+ hours - 1xp/hr
3. the rate is only reset on a new date with a new session.

Multiple sessions in the same day are tallied together such that the overall rate is applied is if they were one single session. However, the calculations are done every hour in `ch30_hour` and incrementally add xp. (aka we dont just tally and then add a ton of xp)

## new functions
* `mas_utils.eqfloat` - variable accuracy float equivalence check
* `mas_utils.floatsplit` - splits a float into int + float parts
* `mas_utils.td2hr` - converts a timedelta to a fractional hours amt
* `mas_getTotalPlaytime` - gets total playtime from sessions data
* `mas_getTotalSessions` - gets total sessions from sessions data
* `mas_xp.calc` - calculates xp gains since the last call to `calc`
* `mas_xp.grant` - grants xp based on the current state. this is used in `ch30_hour`
* `mas_xp.level` - returns the user's current level

## Testing:
- Verify games unlock at a *decent* rate (note this will take at least a day ideally to begin unlocking things)
- Verify the gender topics still work fine (both first time and redo)
- Verify conditionals aren't set on games which have already been unlocked when calling the update script
- Verify when running the mod, XP is gained in the rate described above.
- Verify that running the update script gives u a correct amount of levels. To calculate:
    1. Find number of days since your first session
    2. Take total playtime and convert to hours
    3. divide total playtime (hours) by number of days since first session
    4. Run that amount through `mas_xp.calc_by_hours` with a starting rate of `60.0`.
          * alternatively, you can calculate this by yourself by taking every 2 hours and applying the rate as described above
    5. divide the amount of xp by 420. The integer portion of this result should be the number of levels you gain. `420 - (float part * 420)` then should be the amount of xp to next level.
    * you can also run `mas_xp._grant_on_pt` to get a tuple with the levels + xp to next level based on your current playtime/first session data. Just set the appropriate session data to run the alg.
    * if running the update script, verify `persistent._mas_pool_unlocks` is set to the same amount of levels gained


ThePotatoGuy:
what is the amount of time before unlock of each thing that was changed? think in terms of users who maybe spend 1/4/8/12/24 hours a day with Monika.

multimokia:
From before, at least from just doing testing alone, you could unlock pong/chess within a few hours (w/o TT)

(Combined with TT is worse since you gain a substantial amount of xp for being away, even just one/two days away can get you up to chess+)

ThePotatoGuy:
I mean what are the timelines now? I want calculated approximations like how 1k affection was set to be about 5 months worth of time with the 7-aff cap.

ThePotatoGuy:
Here's the exact scenarios I want calculated:
* user runs idle 1 hour a day, same time each day (60 idle + 230 rest xp (10 * 23 hours))
* user runs optimal xp gain strat, which is 120 idle (equiv to 2 hour session), then 60 rest xp (6 hours) , then 2 hour session, and repeat

And additional numbers:
* maximum xp gain from all current topics triggered by `call_next_event` - probably a blanket multiply with all topics from all dbs will suffice
* maximum xp gain from playing games and winning them - should be constant value unless one of the games has repeatable `grant_xp`

PR #5421, opened 2020-03-17
Updater Links

### LINKED ISSUE #5417: Links on final updater screen
Players that use the in-game updater never really have a reason to visit GH or the MAS website and may not think to read release notes, know that sprite packs exist or know we have a Wiki that can provide help. So it seems like it could be useful on the final screen of the updater (the one that says Please restart Monika After Story) to include links to these 3 pages just to remind people that these things exist.

multimokia:
closes #5417 

Adds links to the latest releases page and the wiki page.

Adding a direct link to the latest spritepack zip isn't quite as possible, but ideally linking to the latest release w/ patch notes should suffice as the zip is a scroll down

## Testing:
- Verify that links work correctly
- Verify links show
- Verify that updating with rpy files doesn't keep the window showing up during the updater screen (oops, that happened for a while)

### DIFF
--- a/Monika After Story/game/screens.rpy
+++ b/Monika After Story/game/screens.rpy
@@ -35,7 +35,8 @@ init -1 python:
 
     _TXT_FINISHED_UPDATING = (
         "The updates have been installed. Please reopen Monika After Story.\n\n"
-        "See the patch notes and get spritepacks {a=https://github.com/Monika-After-Story/MonikaModDev/releases/latest}{i}{u}here{/u}{/i}{/a}.\n"
+        "Get spritepacks {a=http://monikaafterstory.com/releases.html}{i}{u}from our website{/u}{/i}{/a}.\n"
+        "See the patch notes {a=https://github.com/Monika-After-Story/MonikaModDev/releases/latest}{i}{u}here{/u}{/i}{/a}.\n"
         "Confused about some features? Take a look at our {a=https://github.com/Monika-After-Story/MonikaModDev/wiki}{i}{u}wiki page{/u}{/i}{/a}."
     )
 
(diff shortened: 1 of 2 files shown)


Monika After Story/game/screens.rpy
```diff
@@ -33,7 +33,11 @@ init -1 python:
         "and offer some comments based on what you're doing."
     )
 
-
+    _TXT_FINISHED_UPDATING = (
+        "The updates have been installed. Please reopen Monika After Story.\n\n"
+        "See the patch notes and get spritepacks {a=https://github.com/Monika-After-Story/MonikaModDev/releases/latest}{i}{u}here{/u}{/i}{/a}.\n"
```
ThePotatoGuy: this should be linked to the website (http://monikaafterstory.com/releases.html)

jmwall24: thoughts of instead changing the link each release to the actual patch notes for that release instead?

ThePotatoGuy: using direct download from the website has reduced the amount of users installing source versions.

jmwall24: we aren't asking them to download the release, the only people that will ever see this are people that have already updated automatically.

ThePotatoGuy: spritepacks should be downloaded from the website as well. Anything to limit the chance for users to dl source versions. I'm fine with two links, one for patch notes and one for spritepacks.

jmwall24: so a link to the actual zip file then for the sprite pack? that's the only link for the sprite pack directly

ThePotatoGuy: no, link to the releases page for spritepacks. "Latest Spritepacks" is listed there front and center.

PR #5455, opened 2020-03-30
Fixed Pigtails (But made the pr correctly this time)

kkrosie123:
thanks for the help, multi :"D one day I'll be able to do this myself.


multimokia:
NOTE: Since #5447 removes night sprites, not adding night versions of these adjusted images here

PR #5457, opened 2020-03-30
Taking Monika somewhere

Polaris9255:
Looking for more topic options when returning Monika from taking her out somewhere, she could ask where you went. Responses could be (but not limited to) "Just for a drive", "I brought you to work", "For a walk on the beach", "To a mall", "To the movies", etc. Just more interaction with that would be nice. Thanks for reading and doing all you do. <3


multimokia:
The pull requests tab isn't for suggestions, please open a new issue.

PR #5490, opened 2020-04-09
Small fix for MASFilterableSprite

Booplicate:
Right now `MASFilterableSprite` renders its image on the entire screen. This means that it makes a huge displayable. That doesn't matter unless you want to make events with that displayable. Then the user is able to interact with that displayable from any point of the screen (because it's bigger than the actual image is). This will set the render's size to its image's size so the displayable will behave as you expect it to.

### DIFF
--- a/Monika After Story/game/sprite-chart-matrix.rpy
+++ b/Monika After Story/game/sprite-chart-matrix.rpy
@@ -88,7 +88,8 @@ python early:
 
             # render and blit
             render = renpy.render(new_img, width, height, st, at)
-            rv = renpy.Render(width, height)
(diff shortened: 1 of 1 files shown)


Booplicate:
> FYI, `MASFilterableSprite` should **not** be used by anyone right now as it was untested and likely incomplete. Additionally this may not be the actual way we want arbitrary sprites to be filtered with.

I tested it for a bit, and it works just fine for my purposes (no lags, leaks, artifacts). I'm using it as a temporal solution. As soon as we get another tool to work with, I'll move to it.

PR #5521, opened 2020-04-12
Better Game Handling

multimokia:
Completely changed how game locking/unlocking works, as well as games themselves.

# Changes
- Games are now `Event` objects. As such, they may have `aff_range` and `conditional` properties (`action`, `start_date`, and `end_date` do not do anything here)
- `mas_isGameUnlocked()` now reflects the state of the game conditionals and aff range as well. While the `Event` may be unlocked, if it fails the `aff_range` check or `conditional`, it is considered locked.
- `persistent.game_unlocks` is now irrelevant, and has been removed
- `pick_a_game` has been changed to `mas_pick_a_game`. Additionally it has been changed to use the `mas_gen_scrollable_menu`

All in all these changes allow less manual expansion of the game system.

# Testing
- Verify all game unlocks still work
- Verify that having chess locked (both timed and permanent) works properly, and the game is considered locked
- Verify that chess still unlocks properly with only the exp conditional
- Verify that all looks fine with the new play menu and works fine as well (going to and from games is clean, additionally so with `Nevermind.`)
- Verify XP gain works when session times are less than an hour.

## Update scripts
- Verify that a user who started after 2020-04-04 with less than 2 hour avg session time gets appropriate amount of xp/levels


ThePotatoGuy:
@jmwall24 @Legendkiller21 double check that update script for xp fixing works for persistents that start on 2020-04-04 and have less than 2 hours avg session time

PR #5541, opened 2020-04-14
Various UI code fixes

Iniquitatis:
This PR is actually pretty basic, nothing groundbreaking from a user perspective. It is mostly just some sort of a massive code cleanup.
That said, it introduces a couple of important internal changes:
* Almost every _styled_ button has been derived from the single generic style (`MASButtonDisplayable`s are still unaffected);
* The dark mode is now automagically supported on each style which has the `_dark` suffix at the very end of its name;
* There's no more need for reassigning the style prefixes to support the dark mode.

I'm almost done with main changes, so it can be tested and merged at any point (the earlier, the better). Everything else could be done in other PRs.


Iniquitatis:
Oops, I'm so sorry. I should've been testing it with an unmodified game.
Will fix.

Iniquitatis:
Actually, I think we can get rid of every "hkbd_" style in favor of proper assignment of "insensitive_" properties to the "hkb_" buttons.
It does work and simplifies codebase a bit.

ThePotatoGuy:
Yeah, using `Frame`. All buttons created after the square frame was added should have used Frame, but the remaining old ones were left unchanged as they were working fine and it wasn't worth adding that scope to the PRs that added frames.

If you've managed to transition everything correctly then go ahead and put them in this pr.

ThePotatoGuy:
@Legendkiller21 for this one, testing should be:
* verify that every UI element (**except custom displayables like chess + piano**) looks correct in both light mode and dark mode
* verify toggling between light/dark mode affects the UI appropriately

Also this changes the disabled font color to grey, which I think looks better than previous so that will look different when compared to custom disp buttons.

Legendkiller21:
There's a small issue with the scroll bars. Up until now, while we had dark bar styles, like `mas_adjust_vbar_dark` and grey bars that are used by these styles, we never used them because we used to use bars only in menus like the extra menu or the scrollable menus. In these menus it's harder to see the grey bar because of the classroom bg so we always used the stock black ones.

The new functions automatically find and use all dark styles but because of that, these menus now use the grey bars. The problem is that now that we have a submod menu we need them as it's hard to see a black bar on a dark bg.

The only solutions I have found is:
- Delete the dark bar styles so that every menu will use the stock ones. This way though, we will have to manually adjust all the screens, like the submods screen, to use a grey bar depending on  `mas_globals.dark_mode`'s value
- Or keep the styles as they are and manually adjust the menus like the extra menu and the scrollable menus, so that they always use a black bar.

Iniquitatis:
@Legendkiller21 all done.

Monika After Story/game/zz_extrasmenu.rpy
```diff
@@ -311,28 +301,43 @@ style mas_mbs_button_text is default:
...
     bar_vertical True
 
-style mas_adjustable_button is default:
-    idle_background Frame("mod_assets/buttons/squares/square_idle.png", left=3, top=3)
-    hover_background Frame("mod_assets/buttons/squares/square_hover.png", left=3, top=3)
-    hover_sound gui.hover_sound
-    activate_sound gui.activate_sound
+style mas_adjust_vbar_dark:
+    xsize 18
+    base_bar Frame("gui/scrollbar/vertical_poem_bar_d.png", tile=False)
```
Legendkiller21, line 324:
```suggestion
    base_bar Frame("gui/scrollbar/vertical_poem_bar.png", tile=False)
```

Extras menu also needs to always use the stock bar, just like the scrollable menus. We can either change `mas_adjust_vbar_dark` or create a new style similar to `classroom_vscrollbar`

Iniquitatis: Done.

PR #5554, opened 2020-04-16
Filtering Framework Expansion

ThePotatoGuy:
# Key Changes
* `MASFilterSwitch` - filter-based version of ConditionSwitch. Basically builds a ConditionSwitch that uses filters appropriately. Should be used for most cases, although this may change as we use more filters.
* `MASFilteredSprite` - generates a filtered version of an image as a displayable
* `MASFilterableSprite` - updated to include highlight via `MASFilterMap`. In general, this should only be used if you need to combine a sprite and its highlight for whatever reason. 
* filter adding and adjustment framework:
    * `mas_sprites.add_filter` - adds a new filter. Must be called before init level -1.
    * `mas_sprites.get_filter` - gets the current filter
    * `mas_sprites.set_filter` - sets the current filter. ignores invalid filter values
    * `mas_isCurrentFlt` - checks if the current filter is the same as the passed in one
* **the `spaceroom` label is now in charge of progressing the filter.** It will handle scene change and dissolve all when filter change is appropriate.
* added parameter `progress_filter` to the `spaceroom` label. Passing in True will prevent spaceroom from progressing the filter.
* added scene change request option to idle mailbox. Also updated weather to use this.
* replaced `morning_flag` in the calendar displayable to `self.day_mode` which is set by the `mas_current_background.isFltDay` function. 
* Marked some TODOs for `MASBackground` integration later
* removed `morning_flag`
* removed `dev_sprite_timer.rpy`
* removed `mas_weather.should_scene_change` - if you want the scene to change, call `mas_idle_mailbox.send_scene_change`

## new API functions
* `mas_progressFilter` - progresses the filter depending on rules. This should only be called by spaceroom label. **DO NOT USE ANYWHERE ELSE.**
* `mas_isDay` - replaces `mas_isSunny`
* `mas_isDayNow` - variant of `mas_isDayNow` except uses current time. This should replace previous uses of `mas_isMorning`
* `mas_isNightNow` - variant of `mas_isNight` except uses current time. This should replace previous uses of `not mas_isMorning`

## New `MASBackground` functions
* `isFltDay` - checks if the given/current filter is considered a "day" filter. This will be fully implemented with `filtering-backgrounds` development in the future. For now it is specific to day/night filtering. This should replace all uses of the global `morning_flag`
* `isFltNight` - checks if the given/current filter is considered a "night" filter. See `isFltDay`. This should replace all uses of `not morning_flag`

# Testing
* verify spaceroom + monika looks okay
* verify spaceroom + monika transitions between day and night appropriately. When changing suntimes via settings, the spaceroom may not update right away.
* verify exp previewer looks ok and the Filter option works ok
* verify calendar still looks okay in day and night UI
* verify opendoor/crashed/islands greetings still look okay
* verify islands still looks okay
* verify weather progression/changing still works okay
* verify the following images look fine day and night:
    * `mas_o31_deco`
    * `mas_d25_banners`
    * `mas_mistletoe`
    * `mas_d25_lights`
    * `mas_d25_garlands`
    * `mas_d25_tree`
    * `mas_d25_gifts_1`
    * `mas_d25_gifts_2`
    * `mas_d25_gifts_3`
    * `mas_bday_cake_monika` - lit and unlit
    * `mas_bday_cake_player` - lit and unlit
    * `mas_bday_banners`
    * `mas_bday_balloons`
    * `chibi_peek`


ThePotatoGuy:
oops there are more `morning_flag` cases that need to be transitioned. please hold

Monika After Story/game/definitions.rpy
```diff
@@ -4490,6 +4490,15 @@ init -1 python:
         return persistent._mas_sunrise <= _curr_minutes < persistent._mas_sunset
 
 
+    def mas_isSunnyNow():
```
multimokia, line 4493: I think we may want to change the names of these functions, as at a glance they come across to account for weather when in reality they don't.

Ideally this should be `mas_isSunUpNow()` and `mas_isSunny()` changes to `mas_isSunUp()`

This way there won't be any potential misunderstandings/misuses of the function

ThePotatoGuy:
@multimokia changes addressed

ThePotatoGuy:
marking as high priority because of future dev that will the additional filtering options

Monika After Story/game/zz_calendar.rpy
```diff
@@ -222,8 +223,12 @@ M̼̤̱͇̤ ͈̰̬͈̭ͅw̩̜͇͈ͅa̲̩̭̩ͅs̙ ̣͔͓͚̰h̠̯̫̼͉e̗̗̮r
             """
             super(renpy.Displayable, self).__init__()
 
+            # dont really feel like changing every image line to not do inline
+            # if statements so this will work for now.
+            self.day_mode = not mas_globals.dark_mode
```
Legendkiller21, line 228: Calendar was not affected by `mas_globals.dark_mode` as it is not considered as part of the ui but part of the background, so it should change depending on the day or the filter cycle.

ThePotatoGuy: but why? calendar is all UI elements so it should follow the UI rules.

multimokia: Calendar is only somewhat a button. Since it hangs on the spaceroom walls and also functions as a part of the room, we treat it as such. Additionally, it makes no sense for the calendar to change with ui modes because it's a part of the room.

Issue #5064, opened 2019-12-06
[Enhancement] - What day is it?


Thimeo:
I was thinking that whenever I look at the calendar, I don't get to know what day it is and I have to look at window's clock at the bottom right of my screen.

Thought that it could be an easy and useful addition to the calendar.

ThePotatoGuy:
[closing until we look into it]

Issue #1801, opened 2018-06-23
[Suggestion] Monika's handwriting font on calendar


jmwall24:
As the title suggests, it seems to me that since it's Monika's calendar she would write the notes on there herself so it'd be in her handwriting, which I just love to look at.  Maybe the actual holidays would be in a generic font because they come that way on calendars anyway, but the personal stuff Monika would have to write on there.

ThePotatoGuy:
[Closing until we look into this]]

ThePotatoGuy:
reopening as in progress

multimokia:
closing as completed

PR #5564, opened 2020-04-17
Handwritten calendar notes

Iniquitatis:
(See: #1801, #5064, #5605)

Handwritten
One picture is worth a thousand words.

---
This PR is got a bit out of its scope (once again >_>), so here's the list of key changes:
* Obviously, changed the font for the notes/events in the calendar cells
* Any cell can now contain three events, and an ellipsis is displayed in the bottom-right corner of the cell if there are >3 events
* Current day is now highlighted
* A bunch of alignment fixes here and there, so the calendar is now a little bit more polished
* Fixed multiple layout issues in the calendar event list
* Fixed the day/night color issues in the calendar event list
* Fixed the calendar button "bleeping" while the calendar is opened (by @Booplicate)
* Added a soft shadow to the calendar button (um... it was just too flat)


Monika After Story/game/zz_calendar.rpy
```diff
@@ -617,13 +618,28 @@ M̼̤̱͇̤ ͈̰̬͈̭ͅw̩̜͇͈ͅa̲̩̭̩ͅs̙ ̣͔͓͚̰h̠̯̫̼͉e̗̗̮r
...
 
+                    day_number_text = Text(
+                        self.DAY_NUMBER_DISPLAY_FORMAT.format(str(current_date.day)),
+                        font=gui.default_font,
+                        size=self.DAY_NUMBER_TEXT_SIZE,
+                        color=self.DAY_NUMBER_COLOR,
+                        outlines=[]
+                    )
+
+                    note_text = Text(
```
ThePotatoGuy, line 629: Can you mark a TODO here for adding a way to decide which font to use on the calendar? 
We're thinking that standard holidays (christmas/halloween/etc...) should use the usual UI font, and only custom ones should appear handwritten. This kind of change requires rework of the calendar DB and is out of scope for this.

Monika After Story/game/zz_calendar.rpy
```diff
@@ -879,13 +895,19 @@ M̼̤̱͇̤ ͈̰̬͈̭ͅw̩̜͇͈ͅa̲̩̭̩ͅs̙ ̣͔͓͚̰h̠̯̫̼͉e̗̗̮r
             cal_r_buttons = [
                 (
                     x.render(width, height, st, at),
-                    (x.xpos, x.ypos)
+                    (x.xpos, x.ypos),
+                    y.render(width, height, st, at),
+                    (x.xpos + self.DAY_BUTTON_WIDTH - y.render(width, height, st, at).get_size()[0] - 7, x.ypos + 5),
+                    z.render(width, height, st, at),
```
ThePotatoGuy, line 901: since note text doesn't exist for every calendar date, this is a wasteful render (and arguably a wasteful generation of a Text object) when building these buttons. Rather than always building a Text object for note text, can you only create and render Text objects if we have actual text to show. 

Additionally, instead of creating a larger tuple for each day cell, I would restructure this loop to not use list comprehension and instead add the new renders to the list in the same format `(render, xy_tuple)`. That way the blit loop doesn't have to unpack so many items and doesn't need a branch for cases without note text.

Iniquitatis:
@ThePotatoGuy can I also commit a couple of unrelated positioning fixes to this branch?
(I've managed to fix all those broken offsets in the calendar, but amount of changes is still too small to deserve a separate PR.)

Monika After Story/game/zz_calendar.rpy
```diff
@@ -626,7 +646,7 @@ M̼̤̱͇̤ ͈̰̬͈̭ͅw̩̜͇͈ͅa̲̩̭̩ͅs̙ ̣͔͓͚̰h̠̯̫̼͉e̗̗̮r
                         self.DAY_NUMBER_DISPLAY_FORMAT.format(str(current_date.day)),
                         font=gui.default_font,
                         size=self.DAY_NUMBER_TEXT_SIZE,
-                        color=self.DAY_NUMBER_COLOR,
+                        color=self.DAY_NUMBER_COLOR if not is_today else self.TODAY_NUMBER_COLOR,
```
ThePotatoGuy, line 649: this is now 4 additional branches per button creation. Can you move these so you just check `is_today` once per button, set temp vars as appropriate, and pass them into the `MASButtonDisplayable` accordingly?

Iniquitatis: Done.

(The amount of `_bg`s is becoming a little bit messy, though.)

Booplicate:
Will you change the font for this menu as well?
Unknown

Iniquitatis:
> Will you change the font for this menu as well?

TBH, I'm not really sure if this would be conceptually applicable. Might need to ask someone from the core team.
... Not to mention that I still haven't seen this menu.

Booplicate:
Oh, okay. That menu appears when you have 4+ events for a day.

ThePotatoGuy:
@booplicate do u have a submod that is using the 4+ event popup?

ThePotatoGuy:
alright so some currently in development is using the calendar event list popup, which means that we need to pause this for a second.

you need to check how many events can be listed with Monika font.
additionally there is a "click here to see more line" that will appear for the 4+ case, which doesnt seem right to be in Monika font.

I was not planning to bring these up here since we dont officially have 4+ events in a cell yet, but that is going to change real soon, so this dev needs to account for that now.

Booplicate:
No, not really.
I found that menu when was playing a bit with multi's custom calendar events system. Thought I should mention it since I doubt many people know that the menu exists lol.
But looks like you already figured that out.

Iniquitatis:
@ThePotatoGuy ah, I see. The line height is too big.

(IMHO, if we're not planning to change the calendar layout then four lines in a single cell may look a bit too cluttered even with default font. I feel like this "event menu" should be openable by default.)

---
Three Events
Maybe something like this would be suitable?

ThePotatoGuy:
Actually, we already do 3 lines per cell, just need to make sure it looks correct with the monika font.

That looks okay so far, but what does it look like if an ellipsis is used instead of the "Click to see more"? If that looks fine then we could keep it in monika font.

Iniquitatis:
Oh, I have an another idea: any cell can contain three events and, if there are more than three, show ellipsis in the bottom-right corner. That might be more intuitive and should not clash stylistically with Monika's handwriting.

ThePotatoGuy:
show a screen when you have tried that

Iniquitatis:
But it is already a part of a screenshot.

ThePotatoGuy:
@Legendkiller21 adding you to this as this has grown in scope 
@Booplicate you should also give this a run through

Iniquitatis:
@Legendkiller21 could you please test the event menu without this PR? I don't think I've touched any of its code. It might've been broken after the "dark style PR".

@Booplicate noticed that issue too, but decided to defer its fixing, because it is not that important part of the calendar.
Also, see the first code comment (the resolved one) from ThePotatoGuy about the calendar DB rework.

Iniquitatis:
@ThePotatoGuy yes, I've already went ahead and fixed it locally.
Fixed Glitch

Iniquitatis:
@Legendkiller21 I don't think I've touched any of its code.
This menu appears to be broken in the 0.11.0 release as well (aside from obvious color regressions from the "dark mode PR").
Broken Event Menu

Should I push the fixes to this branch?

ThePotatoGuy:
Sure

Iniquitatis:
@ThePotatoGuy how about something like this?
Event Menu
(The back button is always at the bottom of the list, regardless of the number of items.)

Iniquitatis:
@Legendkiller21 any suggestions?
1. Day/light theme
2. Day/dark theme
3. Night/light theme
4. Night/dark theme
Event Menu Fixes

Legendkiller21:
I think it looks great now

Iniquitatis:
Pushed the latest changes.

Monika After Story/game/zz_calendar.rpy
```diff
@@ -2055,6 +2180,8 @@ screen calendar_overlay():
     #     xalign 0.305
     #     yalign 0.4
     #
+    image "mod_assets/calendar/calendar_button_shadow.png" xpos 351 ypos 251
+
     if store.mas_calendar.enabled:
```
Booplicate, line 2185: Looks and works good!
Mind if I ask you to add one bit (yes, it has nothing to do with the fonts, but still)? It was bothering me for a long time. When you open the calendar, the button that you use to open it is still active (you can hear the hovering sound). I think we could add a check if we're showing the calendar or not. And if we are, we make the button inactive.
```suggestion
    if (
        store.mas_calendar.enabled
        and renpy.get_screen("mas_calendar_screen") is None
    ):
```

Iniquitatis: Ah, sure, I've been meaning to fix it anyway, but completely forgot about it. Thanks!

_(This PR is already out of scope, so...)_

PR #5574, opened 2020-04-19
Changes to the emphasized text rendering

Iniquitatis:
_This change may be controversial._

Another simple "fix", which attempts to improve bold/italic text rendering. Yes, it is the same Aller font, but rendered using the actual TTF files instead of Ren'Py text styling capabilities.
Though, it may seem a bit unusual to people who are used to the original style.

---
Comparison screenshots (from left to right, from top to bottom):

Buttons:
Buttons Comparison

Bold text in the dialog window:
Bold Comparison

Italic text in the dialog window:
Italic Comparison


Iniquitatis:
**TL;DR:** Oh, I'm an idiot. I'm so used to the OFL fonts, which are available as a free download, that I didn't even think about reading the license of this typeface. Sorry.

---
I have downloaded the font files from the [Font Squirrel](https://www.fontsquirrel.com/).
Actually, the most questionable moment here is usage of this typeface in derivative non-profit works of the free commercial product, which (purposefully?) didn't provide the bold/italic variants, but only the regular one.
I may need to consult someone who is better at all these licensing details, so I'm closing it until resolved.

_(The files are still in my branch, so I hope it will not give you any troubles. In any case, I'm the potential lawbreaker here, not any of you. :P)_

PR #5579, opened 2020-04-20
Fixed MASButtonDisplayable rendering

Iniquitatis:
Smallest possible change, so any custom button will now support frames.
Any other user of this class is unaffected. If an instance of the `Image` class is passed as a background, then it still being drawn as usual.


ThePotatoGuy:
because of changes in the linked PR, only one change from this PR would actually be needed as the rest would be obsoleted. I have moved the commit over to that PR so credit will still be given in the commit history.

PR #5580, opened 2020-04-20
Reimplemented graphics menu

Iniquitatis:
Because, uh... why overcomplicate?

_Still, I don't know about the reasons why the original code was written this way, so this PR definitely can be rejected without further discussion._

### DIFF
--- a/Monika After Story/game/zz_graphicsmenu.rpy
+++ b/Monika After Story/game/zz_graphicsmenu.rpy
@@ -24,10 +24,7 @@ style graphics_menu_current_renderer_text is graphics_menu_text:
 
 init python in mas_gmenu:
 
-    # this will be true if a renderer has been selected
-#    renderer_selected = False
-
-    # this is the selected renderre
+    # this is the selected renderer
     sel_rend = ""
 
 init -1 python:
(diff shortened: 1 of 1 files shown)


ThePotatoGuy:
The graphics menu wasn't broken and wasn't a mess like our styles were, so this doesn't really do anything besides add to the testing queue. I'm going to reject this on the grounds that it isn't a fix nor a notable enhancement.

Iniquitatis:
Actually, it was a little broken. Just a little (text color on disabled buttons). But adding code to fix it would bloat a codebase even more.

Iniquitatis:
@ThePotatoGuy I've been lurking in the codebase and came to conclusion that such kind of changes may be mandatory to the _(spoiler alert)_ forthcoming color scheme swapping/UI customization/etc.
So I have to ask you to reconsider your decision, despite what was written in the description of the PR. It could be marked as low priority, but eventually these changes have to be made.

ThePotatoGuy:
There is no color scheme swapping/UI customization development planned. 
The text color on the disabled button not matching other places isn't enough to warrant a complete rewrite of the entire menu. The text color can be fixed by building the appropriate text object and passing it in, which I ended up doing in #5587

PR #5587, opened 2020-04-21
MASButtonDisplayable Enhancements and fixes

ThePotatoGuy:
Includes fix from #5579 but obsoletes all other changes from that PR, so only merging that change here. The other PR will be closed.

# Key Changes
* From #5579 - fix button backs rendering to entire canvas in MASButtonDisplayable.render
* adds two new functions for button creation
    * `MASButtonDisplayable.create_st` - creates a button where the idle/hover/disable text is all the same 
    * `MASButtonDisplayable.create_stb` - creates a button where the idle/hover/disable text is all the same and the backgrounds use `choice_button`'s `background` frames if possible 
* all instances of MASButtonDisplayable now use grey text when disabled.

### API additions
* `mas_getPropFromStyle` - retrieves value of a property from a style. Use full names for a style (not style prefix). This obtains the actual value used by the style, so making any changes to this will affect anything using the style. Duplicate the object if you need something specific (duplicating is non-trivial so its up to the caller)
* `mas_prefixFrame` - extracts an image path from a frame and applies prefix to it using substitute, then builds a new frame with that image path. Only copies over image and left/top/right/bottom props (borders).

# Testing
* verify correct button behavior and appearance (grey text when disabled) in the following places:
    * chess
        * save - idle/hover/disabled (becomes not disabled after reaches the user's 5/6th turn)
        * give up - idle/hover/disabled (disabled during Monika's turn)
        * done - idle/hover
    * updater menu
        * update - idle/hover/disabled (disabled while checking for update, enabled when update found. could trick this by changing your version to a previous one)
        * cancel - idle/hover
        * retry - idle/hover/disabled (only appears if the first update check fails because of internet problems. becomes disabled while retrying. might be possible to trigger by killing internet connection while first update check runs)
        * Ok - idle/hover - appears if game is up-to-date
    * graphics menu
        * return - idle/hover
        * all the other buttons - idle/hover/disabled (disabled if they are the current renderer)
    * piano (normal mode)
        * config - idle/hover/disabled (disabled when a song has been completed)
        * quit - idle/hover
    * piano (config mode)
        * done - idle/hover/disabled (disabled while configuring a key)
        * cancel - idle/hover/disabled (disabled while **not** configuring a key)
        * reset all - idle/hover/disabled (appears while **not** configuring a key, disabled if no keyboard configuration exists)
        * reset - idle/hover/disabled (appears while configuring a key, disabled if the key has not been configured)

### DIFF
--- a/Monika After Story/game/definitions.rpy
+++ b/Monika After Story/game/definitions.rpy
@@ -1904,7 +1904,7 @@ python early:
             # pull out the current button back and text and render them
             render_text, render_back = self._button_states[self._state]
             render_text = renpy.render(render_text, width, height, st, at)
-            render_back = renpy.render(render_back, width, height, st, at)
+            render_back = renpy.render(render_back, self.width, self.height, st, at)
 
             # what is the text's with and height
             rt_w, rt_h = render_text.get_size()
--- a/Monika After Story/game/definitions.rpy
+++ b/Monika After Story/game/definitions.rpy
@@ -1794,6 +1794,7 @@ python early:
                         - disable_back
                 **kwargs - keyword args to pass into constructor
             """
+            # determine disabled stuff
             if incl_disb_text:
                 disb_button = Text(
                     text_str,
@@ -1802,12 +1803,7 @@ python early:
                     color=mas_globals.button_text_insensitive_color,
                     outlines=[]
                 )
-                disb_back = Frame(
-                    mas_getTimeFile(
-                        "mod_assets/buttons/generic/insensitive_bg.png"
-                    ),
-                    Borders(5, 5, 5, 5)
-                )
+                disb_back = MASButtonDisplayable._gen_bg("insensitive")
             else:
                 disb_button = Null()
                 disb_back = Null()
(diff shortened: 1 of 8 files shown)


Iniquitatis:
Well, that's a pretty nice compromise.
But still, is there any way we could export the style properties for the piano/calendar/etc.?
Let me explain: I have a prototype for a submod, which allows to read themes from the JSON files, then load the assets and switch them at the runtime. For the most part it works just fine, but I cannot change some styles without replacing the original assets.

Monika After Story/game/definitions.rpy
```diff
@@ -1700,22 +1699,147 @@ python early:
...
+                ),
+                Text(
+                    text_str,
+                    font=gui.default_font,
+                    size=gui.text_size,
+                    color=mas_globals.button_text_hover_color,
+                    outlines=[],
+                ),
+                disb_button,
+                Frame(
```
Iniquitatis, line 1831: Like, for example, here: can we read this `Frame` directly from the `generic_button`'s `background` property?

ThePotatoGuy: they now read from choice button's background.

PR #5615, opened 2020-04-25
New song 'Every rose has its thorn'

Daiozen:
This one was originally part of [this PR](https://github.com/Monika-After-Story/MonikaModDev/pull/4496) but was shelved while the song framework was being worked on.


Daiozen:
Rewrite is done. Hope it matches what you had in mind.

PR #5619, opened 2020-04-25
Generalization of bookmarks/derand

### LINKED ISSUE #5616: [Suggestion] Make Monika sing again.
So, this is just my two cents on the subject, but I miss Monika singing out of the blue; it's something that always helped put a smile on my face. Having to ask her directly to sing does not feel the same...

I know it was derandomized so she wouldn't keep singing all the time as the number of songs kept increasing, but if there was a way to still let her sing once in a while, that's something I'd really enjoy.

multimokia:
closes #5616 

This PR implements a generalized framework for bookmarks/derandoms, which allows more use cases, for example in this PR, allowing songs to be derandomed.

## Changes:
- Massive generalization of framework
- `getBookmarkedTopics`/`getDerandomedTopics` now accept arguments specifying which persistent variables to check to get bookmarked/derandomed eventlabels
- Moved the core workings of `mas_topic_rerand` to its own sublabel, generalized for use elsewhere
- Made songs no longer automatically derandom
- Bookmark blacklisting/whitelisting is now handled by event rules

## Testing:
- Verify bookmarks/derand still work as they did before, not allowing any labels w/o the `monika_` prefix and with the other criteria they had before (no topics w/o prompt, must be normal+ [bookmark only])
- Verify songs can be both bookmarked and derandomed
- Verify derandoming songs works as well as rerandoming them (likewise for normal topics)
- Verify that topics which have the bookmark whitelist rule can be bookmarked
- Verify that topics with a bookmark blacklist rule cannot be bookmarked


Monika After Story/game/event-handler.rpy
```diff
@@ -1499,11 +1499,10 @@ init python:
                 (Default: False)
         """
         if ev:
-
             if unlock:
                 ev.unlocked = True
 
-            if _random:
+            if _random and store.mas_bookmarks_derand.shouldRandom(ev.eventlabel):
```
ThePotatoGuy, line 1505: Why are bookmarks derand rules being enforced here?

The event APIs here are for helping us do common actions regardless of any topic restrictions. 
It's the caller's responsibility to make sure they aren't randoming a topic that shouldn't be randomed. Where are the problem lines that call this without checking? If it's truly a significant amount, we should just have a separate API for randoming a topic that also checks whatever restrictions we may have.

Monika After Story/game/event-handler.rpy
```diff
@@ -2040,9 +2039,10 @@ init 1 python in evhand:
             ev - event to random
             rebuild_ev - True if we wish to notify idle to rebuild events
         """
-        ev.random = True
-        if kwargs.get("rebuild_ev", False):
-            store.mas_idle_mailbox.send_rebuild_msg()
+        if store.mas_bookmarks_derand.shouldRandom(ev.eventlabel):
```
ThePotatoGuy, line 2042: This is a pretty major core change. Is the issue that events are being re-random'd even though they shouldn't be random? What's causing them to have conditionals/actions that make them re-random?

Monika After Story/game/script-topics.rpy
```diff
@@ -535,6 +521,43 @@ label mas_topic_rerandom:
     return
 
 init python in mas_bookmarks_derand:
+    import store
+
+    #Rule constants
+    WHITELIST = "whitelist_bookmark"
+    BLACKLIST = "blacklist_bookmark"
```
ThePotatoGuy, line 528: Since the intent of these is to be used in rules, what happens if a topic has both of these? I think this would be better as a single key, and use the value to determine wl/bl.

Monika After Story/game/event-handler.rpy
```diff
@@ -446,6 +446,33 @@ init 6 python:
...
+            ev_label - label of event to show
+            code - string code of the db this ev_label belongs to
+            unlock - True if we want to unlock this Event
+                (Default: False)
+            _random - True if we want to random this event
+                (Default: False)
+            _pool - True if we want to random thsi event
+                (Default: False)
+        """
+        store.mas_showEvent(
```
ThePotatoGuy, line 470: better to call `mas_showEVL` from here so dont have to copy the `mas_all_ev_db_map` code.

Monika After Story/game/script-topics.rpy
```diff
@@ -365,46 +382,366 @@ init python:
...
+
+        return len(mas_get_player_bookmarks(persist_var)) > 0
 
+    def shouldRerand(eventlabel):
+        """
+        Checks if we should rerandom the given eventlabel
+
+        IN:
+            eventlabel
+        """
```
ThePotatoGuy, line 472: this has no code

PR #5657, opened 2020-04-28
Default selectors to seen

multimokia:
This PR adds a new param to addEvent, `defaultSeenever`. It's the equivalent of adding a `persistent._seen_ever["topic_evl"] = True` after its addEvent.

I have also updated the documentation on the addEvent function.

Additionally, selectors now all default to seen with the exception of the ribbon selector, which is now `aff_range`d (Normal+), to indicate to the user that the Appearance tab is something where they can go to change Moni's appearance.

# Testing:
- Verify BRBs don't show up in unseen still,
- Verify that the happy birthday topics in the compliments menu are not italicized if you haven't used them before
- Verify that all selectors aside from the ribbon selector will default to a True _seen_ever

### DIFF
--- a/Monika After Story/game/event-handler.rpy
+++ b/Monika After Story/game/event-handler.rpy
@@ -1292,7 +1292,7 @@ init python:
         eventdb=None,
         skipCalendar=False,
         restartBlacklist=False,
-        defaultSeenever=False,
+        markSeen=False,
         code="EVE"
     ):
         """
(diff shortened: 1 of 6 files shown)


Monika After Story/game/event-handler.rpy
```diff
@@ -1288,29 +1288,38 @@ init python:
...
-        #       (Default: None)
-        #   skipCalendar - flag that marks wheter or not calendar check should
-        #       be skipped
-        #   code - code of the event database to add to.
-        #       (Default: EVE) - event database
+        event,
+        eventdb=None,
+        skipCalendar=False,
+        restartBlacklist=False,
+        defaultSeenever=False,
```
ThePotatoGuy, line 1295:
```suggestion
        markSeen=False,
```
I think this is a better param name.

PR #5658, opened 2020-04-28
Change init levels for mas_utils

multimokia:
Small submods error. In order to keep initing them at -990, we needed `mas_utils.writelog`, but since we're calling a submod init at -990, the same init level, it's possible that `writelog` may not exist.

This addresses that by putting the block which initializes the function at init -991.

# Testing:
- Verify duplicate installed submods don't crash and do a writelog properly
- Verify nothing else breaks


ThePotatoGuy:
pr test

PR #5663, opened 2020-04-29
[New dialogue] "The Story of an Hour" (Literature)

TheFishDude:
Hi! It is a dialogue about a short story "The Story of an Hour" by Kate Chopin. Would appreciate any suggestions or changes. Thanks!

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -14195,3 +14195,31 @@ label monika_translating_poetry:
         m 5ekbsa "It would be nice if you could read some of them for me sometime..."
         m 5rkbsu "You'd have to translate them for me first, though~"
     return
+
+init 5 python:
+    addEvent(
+        Event(
+            persistent.event_database,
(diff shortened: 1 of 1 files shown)


Mugnika:
I really cannot see Monika telling us this story AT ALL.  As the character in the story slowly realises she is happier without her husband and relishes in her new sense of freedom. It was also used as a tool for a political agenda as the author plotted the idea women were oppressed by unhappy marriages in those times and politics is something I NEVER want to see in MAS.

masterliamy:
Considering Monika's character you'd expect that a story like this will be read by her with disgust in her voice rather than telling us that it's interesting and that we should read it ourselves.
Yea, No, this doesn't belong here.

jmwall24:
My initial comments on this are if you read what is written here, it's not so much about the content of the story as it is about the literature aspect of it. Being able to convey a lot in such a short story. And even if she disagrees with the overall message doesn't mean she'd never read it. It's important to not only read stuff you agree with. This topic probably needs more at the end regarding her opinion on the overall message, but I don't have a problem with it being here.

PR #5686, opened 2020-05-02
Make sure Submod author and name are proper label names

proudust:
You can't use Update scripts function if the author and name of the submod contains characters that can't be used in the label.
To prevent this from happening, we added a process to verify the author and name.

### DIFF
--- a/Monika After Story/game/zz_submods.rpy
+++ b/Monika After Story/game/zz_submods.rpy
@@ -18,6 +18,7 @@ init -990 python in mas_ui:
     ]) > 0
 
 init -991 python in mas_submod_utils:
+    import re
     import store
     import sys
     import traceback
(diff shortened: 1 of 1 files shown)


Monika After Story/game/zz_submods.rpy
```diff
@@ -101,6 +101,14 @@ init -991 python in mas_submod_utils:
             except:
                 raise SubmodError("Version number '{0}' is invalid.".format(version))
 
+            #Make sure author and name are proper label names
+            import re
```
ThePotatoGuy, line 105: put the import at the top of the store module so its not importing everytime a submod object is created

Monika After Story/game/zz_submods.rpy
```diff
@@ -101,6 +101,14 @@ init -991 python in mas_submod_utils:
...
+            prog = re.compile(ur'^[ a-zA-Z_\u00a0-\ufffd][ 0-9a-zA-Z_\u00a0-\ufffd]*$')
```
ThePotatoGuy, line 106: compile is expensive and only needs to be done once. This should be moved to the class's static vars

PR #5703, opened 2020-05-03
Infinite Loop override

ThePotatoGuy:
# Key Changes
* `renpy.execution.check_infinite_loop` replaced with dummy function (that can be used for other overrides as well).
* removed instances of `not_infinite_loop`

# Testing
* verify no crashes/loop-related tracebacks

### DIFF
--- a/Monika After Story/game/definitions.rpy
+++ b/Monika After Story/game/definitions.rpy
@@ -19,7 +19,7 @@ python early:
     import traceback
     _dev_tb_list = []
 
-    def dummy():
+    def dummy(*args, **kwargs):
         """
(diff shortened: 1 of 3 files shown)


Monika After Story/game/definitions.rpy
```diff
@@ -19,6 +19,15 @@ python early:
     import traceback
     _dev_tb_list = []
 
+    def dummy():
```
multimokia, line 22: Thinking we can do this to make it more general use to just about any override func and replace with the dummy func w/o needing to make another one if it takes args

```suggestion
    def dummy(*args, **kwargs):
```

PR #5722, opened 2020-05-05
Added sanity checks for consumable events

### LINKED ISSUE #5721: Traceback Report
Hello! I'm using v0.11.2-unstable.2020.05.02.119, and I got a traceback today. I guess I need to report this here in case this is important.

I'm sorry, but an uncaught exception occurred.

While running game code:
  File "game/script-ch30.rpy", line 1439, in script call
  File "game/event-handler.rpy", line 2086, in script call
  File "game/zz_consumables.rpy", line 1448, in script call
  File "game/zz_consumables.rpy", line 1460, in script
  File "game/zz_consumables.rpy", line 1465, in <module>
IndexError: list index out of range

-- Full Traceback ------------------------------------------------------------

Full traceback:
  File "game/script-ch30.rpy", line 1439, in script call
  File "game/event-handler.rpy", line 2086, in script call
  File "game/zz_consumables.rpy", line 1448, in script call
  File "game/zz_consumables.rpy", line 1460, in script
  File "D:\Games\MAS\renpy\ast.py", line 814, in execute
  File "D:\Games\MAS\renpy\python.py", line 1719, in py_exec_bytecode
  File "game/zz_consumables.rpy", line 1465, in <module>
IndexError: list index out of range

Windows-8-6.2.9200
Ren'Py 6.99.12.4.2187
Monika After Story 0.11.2-unstable.2020.05.02.119

Thank you very much. Have a nice day!

Booplicate:
Potentially closes #5721
It's possible to have 2 queued events. One is about running out of something, another is for gifting a consumable. If we get the gifting event first, we'll refill the consumables. Which will lead to a crash since the former event will get an empty list from `_getLowConsNotWarned`/`_getLowCons` (we just refilled).
I don't want to pop it from the event list, since we still can have another consumable that we want Monika to mention to the player. But we can add a check if we have at least 1 consumable which Monika is low on during the time we see the event.
For testing I set a consumable to low amount and said goodbye. Then I gifted the consumable and opened the game. Monika reacted on the gift and skipped the running out event.

### DIFF
--- a/Monika After Story/game/zz_consumables.rpy
+++ b/Monika After Story/game/zz_consumables.rpy
@@ -1432,8 +1432,7 @@ init 5 python:
 
 label mas_consumables_generic_queued_running_out:
     $ low_cons = MASConsumable._getLowCons()
-    if low_cons:
-        call mas_consumables_generic_queued_running_out_dlg(low_cons)
+    call mas_consumables_generic_queued_running_out_dlg(low_cons)
     return "no_unlock"
 
 init 5 python:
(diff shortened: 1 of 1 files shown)


Monika After Story/game/zz_consumables.rpy
```diff
@@ -1432,7 +1432,8 @@ init 5 python:
 
 label mas_consumables_generic_queued_running_out:
     $ low_cons = MASConsumable._getLowCons()
-    call mas_consumables_generic_queued_running_out_dlg(low_cons)
+    if low_cons:
```
multimokia, line 1435: Ideally, wouldn't this be better to put in the `mas_consumables_generic_queued_running_out_dlg` label?

Since we know that `low_cons` would be an empty list, so checking in that label if the passed in list is empty would suffice and save repeat code

PR #5728, opened 2020-05-06
New song 'Shelter'

Daiozen:
You can find the song/animation [here](https://www.youtube.com/watch?v=fzQ6gRAEoy0).

PR #5734, opened 2020-05-07
Song Analysis Framework

multimokia:
Will be used for #5728 

Adds a new song analysis topic for the songs framework. This allows in-depth analysis for specific songs as a more optional approach.

These analysis topics in this subset will follow aff_range rules, so there have been functions added to all aff prog points which are at/above normal aff to verify if the topic should be unlocked or not.

Additionally, like long songs, these will automatically unlock when seeing their short variant if the analysis topic has the same eventlabel as the short song, but with `_analysis` added onto it.

## Testing:
- Verify that analysis topics unlock automatically if they're correctly labeled
- Verify that the topic to ask Monika to explain a song locks if there's no analyses available, or none available at the aff level you're at, and unlocks when there is something at your aff level.

### DIFF
--- a/Monika After Story/game/script-songs.rpy
+++ b/Monika After Story/game/script-songs.rpy
@@ -198,12 +198,15 @@ label monika_sing_song_pool_menu:
             if song_length == "short":
                 $ song_length = "long"
                 $ switch_str = "short"
+
             else:
                 $ song_length = "short"
                 $ switch_str = "full"
+
             $ end = "{fast}"
             $ _history_list.pop()
             jump monika_sing_song_pool_menu
+
         else:
             $ pushEvent(sel_song, skipeval=True)
             show monika at t11
@@ -249,10 +252,16 @@ label monika_sing_song_analysis:
     call screen mas_gen_scrollable_menu(unlocked_analyses, mas_ui.SCROLLABLE_MENU_TXT_AREA, mas_ui.SCROLLABLE_MENU_XALIGN, ret_back)
 
     $ sel_analysis = _return
-    $ pushEvent(sel_analysis, skipeval=True)
-    show monika at t11
-    m 3hub "Alright!"
+
+    if sel_analysis:
+        $ pushEvent(sel_analysis, skipeval=True)
+        show monika at t11
+        m 3hub "Alright!"
+
+    else:
+        return "prompt"
     return
+
 #START: Random song delegate
 init 5 python:
     addEvent(
--- a/Monika After Story/game/script-songs.rpy
+++ b/Monika After Story/game/script-songs.rpy
@@ -245,7 +245,7 @@ label monika_sing_song_analysis:
             which = "Which"
 
     show monika 1eua at t21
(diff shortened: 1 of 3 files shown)


Monika After Story/game/script-songs.rpy
```diff
@@ -183,7 +285,14 @@ label monika_sing_song_random:
...
 
             #Unlock the long version of the song
-            mas_unlockEVL(rand_song+"_long","SNG")
+            mas_unlockEVL(rand_song + "_long", "SNG")
+
+            #And unlock the analysis of the song
+            mas_unlockEVL(rand_song + "_analysis", "SNG")
+
+            #If we have unlocked analyses for our current aff level, let's unlock the label
+            if hasUnlockedSongAnalyses():
```
Booplicate, line 294:
```suggestion
            if store.mas_songs.hasUnlockedSongAnalyses():
```

Monika After Story/game/script-songs.rpy
```diff
@@ -59,6 +68,53 @@ init python in mas_songs:
...
+                (Default: None)
+
+        OUT:
+            List of unlocked song analysis topics in mas_gen_scrollable_menu format
+        """
+        if curr_aff is None:
+            curr_aff = store.mas_curr_affection
+
+        return [
+            (ev.prompt, ev_label, False, False)
```
Booplicate, line 99: So with this we'll have to have the same prompts for both song and its analysis? I'm not sure if it's ok, so I thought I'd let you know. I think we could:
* use something like `"<song_name> analysis"` (which sounds a bit dumb/formal?)
* or maybe fetch the song's prompt and use it in this menu while not specifying the prompt for the analysis' event at all?
* we also could do nothing if you think it's safe to use the same prompts.

Monika After Story/game/script-songs.rpy
```diff
@@ -158,6 +217,49 @@ label monika_sing_song_pool_menu:
...
+
+        unlocked_analyses = mas_songs.getUnlockedSongAnalyses()
+
+        if mas_isO31():
+            which = "Witch"
+        else:
+            which = "Which"
+
+    show monika 1eua at t21
+    $ renpy.say(m, "[which] song would you like me to sing?", interact=False)
```
Booplicate, line 248: Will she sing songs here too or will she just explain them? Because if it's the latter, then `sing` doesn't work here. Might use this?
```suggestion
    $ renpy.say(m, "[which] song would you like to talk about?", interact=False)
```

Monika After Story/game/script-songs.rpy
multimokia: I think it's fine to use the same prompt. This is in a separate topic where you ask her to explain a song, in said topic, she asks which song you'd like her to explain. So using the actual song name works out

Monika After Story/game/script-songs.rpy
multimokia: Oops, copy/paste lol

PR #5740, opened 2020-05-08
Consumables Dialogue Props

multimokia:
Adds dialogue props dict as a property of `MASConsumable`s

These properties will be used to adjust the dialogue for generic labels to be more flexible for more consumables (prep for when we add more)

Current added props which are used:
- `mas_consumables.PROP_CONTAINER`
  - This is used in place of the old `container` property
- `mas_consumables.PROP_OBJ_REF`
  - This is used for consumables which don't necessarily have a container. For example cake, which would be referenced to as a *slice*
  - At the moment, this is only used if we don't have a container
- `mas_consumables.PROP_PLUR`
  - This is used to mark whether or not this consumable should be referenced to in plural or not

Also added some prefab `dlg_props` dicts to be used for consumables

Additionally, this also implements a maximum stock amount for consumables. This defaults to 150, but can be changed per consumable as it is a property which can be set on init.

- **NOTE:** If none of the props are present, there are fallback bits of dialogue to account for this. They may not always line up, but realistically, there shouldn't be a reason to not include at least one of the above reference properties (container/obj_ref).

# Testing
- Experiment with combinations of the dialogue properties and make sure that dialogue fits and is actually proper English
- Verify that there's no crashes
- Verify that you can only give up to a maximum of 150 servings of coffee/hotchoc (or if you adjust the value, that maximum)
- Verify that if you were over 150 servings of a consumable, the update script will adjust this and set you to the max stock amount

### DIFF
--- a/Monika After Story/game/definitions.rpy
+++ b/Monika After Story/game/definitions.rpy
@@ -5202,7 +5202,7 @@ init python:
             #We want to modify only text
             if _tuple[0] == renpy.TEXT_TEXT:
                 contents[_id] = (_tuple[0], mas_a_an_str(_tuple[1]))
-                break
+                return contents
 
         return contents
 
--- a/Monika After Story/game/definitions.rpy
+++ b/Monika After Story/game/definitions.rpy
@@ -5176,6 +5176,27 @@ init 2 python:
         if ref_str[0] in "aeiou":
             return "an"
         return "a"
+
+#EXTRA TEXT TAGS
+init python:
+    def a_an_tag(tag, argument, contents):
+        """
+        Handles a/an mid-string
+
+        NOTE: This should ONLY surround the exact word needing to be prefixed with a/an
+        All text tags should be kept OUTSIDE of the opening and closing tags for this function
+
+        Usage: I bought [player] {a_an}[tempvar]{/a_an}.
+
+        If tempvar was 'item,' the output is: I bought [player] an item.
+        If tempvar was 'coffee,' the output is: I bought [player] a coffee.
+        """
+        for kind, _text in contents:
+            if kind == renpy.TEXT_TEXT:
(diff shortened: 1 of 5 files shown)


Monika After Story/game/definitions.rpy
```diff
@@ -5161,9 +5161,53 @@ init 2 python:
...
+        """
+        list_to_return = list()
+        for kind, _text in contents:
+            #We want to modify only text
+            if kind == renpy.TEXT_TEXT:
+                list_to_return.append((kind, mas_a_an_str(_text)))
+            else:
+                list_to_return.append((kind, _text))
+
+        return list_to_return
```
Booplicate, line 5208: Looks like I can't break it anymore!
What do you think about using `enumerate` to access and modify the only element of the list we need to? Instead of `append`ing every element to a new list?
```suggestion
        for _id, _tuple in enumerate(contents):
            #We want to modify only text
            if _tuple[0] == renpy.TEXT_TEXT:
                contents[_id] = (_tuple[0], mas_a_an_str(_tuple[1]))
                break

        return contents
```

Monika After Story/game/definitions.rpy
```diff
@@ -5197,15 +5198,13 @@ init python:
...
-        for kind, _text in contents:
+        for _id, _tuple in enumerate(contents):
             #We want to modify only text
-            if kind == renpy.TEXT_TEXT:
-                list_to_return.append((kind, mas_a_an_str(_text)))
-            else:
-                list_to_return.append((kind, _text))
+            if _tuple[0] == renpy.TEXT_TEXT:
+                contents[_id] = (_tuple[0], mas_a_an_str(_tuple[1]))
+                break
```
ThePotatoGuy, line 5205: if we can avoid break, then we should.
```suggestion
                return contents
```

PR #5742, opened 2020-05-09
Persistent MASBackgrounds and MASWeather

multimokia:
*Remember those old creepy test BGs from when we tested the original BG Framework PR? It's time to see them again...*

## Changes:
- `MASWeather`s and `MASBackground`s are now persistent, stored in `persistent._mas_current_weather` and `persistent._mas_current_background` respectively
- BG Change topic is now unlocked and properly locks/unlocks if there's at least two available backgrounds
- Changed the dockstat loop to use the preloop label properly, which also sets weather for when out of the room

## Testing:
- Verify that weather is now persistent
- Using the testcase bg set, verify BGs are now persistent
- Verify that changing bgs still keeps weather consistent
- Verify dockstat loop still works properly (both returning Monika before opening MAS and after)
- Verify that if the current persistent background is locked, if you say goodbye and reload, you're back in the spaceroom
- Verify that if the current persistent weather is not initialized, the weather returns to def when loading in.
- Verify no crashes

### DIFF
--- a/Monika After Story/game/zz_backgrounds.rpy
+++ b/Monika After Story/game/zz_backgrounds.rpy
@@ -1,6 +1,10 @@
 #Here's where we store our background data
 default persistent._mas_background_MBGdata = {}
 
+#Store the persistent background
+#Defaults to def (spaceroom)
+default persistent._mas_current_background = "def"
+
 #START: Class definition
 init -10 python:
 
@@ -312,7 +316,6 @@ init -20 python in mas_background:
 
 #START: BG change functions
 init 800 python:
-
     def mas_setBackground(_background):
         """
         Sets the initial bg
@@ -328,7 +331,7 @@ init 800 python:
         mas_current_background = _background
         mas_current_background.entry(old_background)
 
-    def mas_changeBackground(new_background, by_user=None):
+    def mas_changeBackground(new_background, by_user=None, set_persistent=False):
         """
         changes the background w/o any scene changes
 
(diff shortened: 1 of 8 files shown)


Monika After Story/game/zz_backgrounds.rpy
```diff
@@ -328,7 +331,7 @@ init 800 python:
         mas_current_background = _background
         mas_current_background.entry(old_background)
 
-    def mas_changeBackground(new_background, by_user=None):
+    def mas_changeBackground(new_background, by_user=None, set_persistent=False):
```
ThePotatoGuy, line 334: doesnt look like you use `set_persistent` in here.

Monika After Story/game/zz_backgrounds.rpy
```diff
@@ -337,21 +340,40 @@ init 800 python:
...
 
         mas_current_background.exit(new_background)
         mas_setBackground(new_background)
 
+    def mas_startupBackground():
+        """
+        Sets up the spaceroom to start up in the background you left in if it is unlocked and still exists
+        """
+        if (
+            persistent._mas_current_background in store.mas_background.BACKGROUND_MAP
```
ThePotatoGuy, line 359: what is the plan if `persistent._mas_current_background` is invalid? does it get reset to a valid value somewhere else?

Monika After Story/game/zz_backgrounds.rpy
```diff
@@ -337,21 +340,40 @@ init 800 python:
...
+
     #Just set us to the normal room here
     mas_current_background = None
     mas_setBackground(mas_background_def)
 
     #Make sure the bg selector is only available with at least 2 bgs unlocked
     if mas_background.getUnlockedBGCount() < 2:
         mas_lockEVL("monika_change_background","EVE")
+    else:
+        mas_unlockEVL("monika_change_background", "EVE")
```
ThePotatoGuy, line 376: bg selector topic locking/unlocking should happen in `ch30_reset` instead, in case a late update script/submodl/delayed action/etc makes a change beyond init level 800.

Monika After Story/game/zz_weather.rpy
```diff
@@ -780,14 +779,42 @@ init 800 python:
...
+    def mas_startupWeather():
+        """
+        Runs a weather startup alg, checking whether or not persistent weather should be used
+        Sets weather accordingly
+        """
+        #If the current bg doesn't support weather, we'll just return
+        if store.mas_current_background.disable_progressive:
+            return
+
+        elif not store.mas_weather.force_weather and not store.skip_setting_weather:
```
ThePotatoGuy, line 803: the if above returns so dont need to elif this.

in fact since I don't see any other code in the same line as this elif, the inverse of this condition can just be added to the starting if statement.

PR #5748, opened 2020-05-09
Auto dark mode based on day cycle

Legendkiller21:
~~Currently the auto mode of dark mode checks the current filter to determine which mode it should use. The issue is that, because the filter progression has to happen in the `spaceroom` label, auto mode doesn't work in the main menu and in the open door greeting.~~

~~This changes the auto mode and makes it check the current day cycle instead.~~

This adds a `mas_progressFilter` in the initial label of the open door greeting to update the filter and make the greeting use the correct styles when dark mode's auto mode is enabled.

## Testing:
- Verify that the proper styles are used in open door greeting when auto mode is enabled
- Verify that auto mode still works
- Verify no crashes

### DIFF
--- a/Monika After Story/game/screens.rpy
+++ b/Monika After Story/game/screens.rpy
@@ -1249,7 +1249,7 @@ screen preferences():
                         action [Function(mas_darkMode, persistent._mas_dark_mode_enabled), Function(mas_settings._dark_mode_toggle)]
                         selected persistent._mas_dark_mode_enabled
                     textbutton _("UI: D/N Cycle"):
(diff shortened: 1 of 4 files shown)


ThePotatoGuy:
I'm not too concerned about this, but this would effectively mean UI may not match BG/sprites in some circumstances.

Have you tried calling `mas_progressFilter` in the workflows where UI is shown before spaceroom is called? You would need to pass in True for the `scene_change` param in spaceroom if appropriate.

Also main menu not having dark UI is not really an issue. That menu is not accessible beyond first launch.

PR #5750, opened 2020-05-09
Restrict bad/awk names for the name_input screen

Booplicate:
We don't allow to input those in the change name topic, why don't we ban them in this screen, too?
If the player enters a bad/awk name, Chibika will pop up and ask to use a more appropriate name
### Testing:
* Make sure that when start w/o importing the data from DDLC, you're not allowed to name yourself one of bad/awk names
* The dialogue looks good
* The animation looks good

### DIFF
--- a/Monika After Story/game/screens.rpy
+++ b/Monika After Story/game/screens.rpy
@@ -666,9 +666,22 @@ style quick_button_text_dark:
 ## This screen is included in the main and game menus, and provides navigation
 ## to other menus, and to start the game.
 
-init python:
+init 4 python:
     def FinishEnterName():
-        if not player: return
+        global player
+
+        if not player:
+            return
+
+        if (
+            mas_bad_name_comp.search(player)
+            or mas_awk_name_comp.search(player)
+        ):
+            renpy.play("sfx/glitch3.ogg")
+            player = ""
+            return
+
+        # if the name is correct, set it
         persistent.playername = player
         renpy.hide_screen("name_input")
         renpy.jump_out_of_context("start")
--- a/Monika After Story/game/screens.rpy
+++ b/Monika After Story/game/screens.rpy
@@ -709,25 +709,25 @@ label mas_bad_name_input:
 
             show chibika at mas_chlongjump(x=650, y=405, ymax=375, travel_time=0.8) onlayer screens zorder 10
             "I'm glad you decided to come back!"
-            "I believe that you and Monika will be a great couple together."
+            "I'm sure that you and Monika will be a great couple."
 
             show chibika sad at mas_chflip_s(-1) onlayer screens zorder 10
-            "But if you'll call yourself like that..."
+            "But if you call yourself names like that...{w=0.5}{nw}"
 
             show chibika at sticker_hop onlayer screens zorder 10
-            "You won't win her heart!"
+            extend "You won't win her heart!"
 
             show chibika smile at mas_chmove(x=300, y=405, travel_time=1) onlayer screens zorder 10
-            "Rather just embaress her."
+            "But just embarrass her instead..."
 
             show chibika at mas_chlongjump(x=190, y=552, ymax=375, travel_time=0.8) onlayer screens zorder 10
             window auto
 
     else:
         show chibika smile at mas_chflip(-1), mas_chmove(x=130, y=552, travel_time=0), sticker_hop onlayer screens zorder 10
-        "I don't think she would be comfortable with calling you like that."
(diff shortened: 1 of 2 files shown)


Booplicate:
@ThePotatoGuy What if we show Chibika that will ask to use a more appropriate name? To make a "good impression" or something? Will that remove the confusion?

Using those names kinda ruin the dialogue you get in the intro. Instead of something emotional, you get some kind of a meme.
Hmm, I don't really want to interrupt her speech for a reaction to the player's name. `"What a weird name...so why did you bring me back, daddy?"`. Maybe I can think about something to use there tho.

But if all else fails, can always just say in the end that she's not comfortable with the name, ask to change it, and unlock the name change topic (as you suggested).

ThePotatoGuy:
tbh I think the potential for a memey introduction is hilarious, but a chibiki hardstop at the name entry entry screen sounds fine.

Monika After Story/game/screens.rpy
```diff
@@ -666,13 +666,127 @@ style quick_button_text_dark:
...
+            window show
+
+            show chibika smile at mas_chflip(-1), mas_chriseup(x=700, y=552, travel_time=0.5) onlayer screens zorder 10
+            pause 1
+
+            show chibika at  mas_chflip_s(1) onlayer screens zorder 10
+            "Hey there!"
+
+            show chibika at mas_chlongjump(x=650, y=405, ymax=375, travel_time=0.8) onlayer screens zorder 10
+            "I'm glad you decided to come back!"
```
multimokia, line 711: I think this line only works if you came from act 4. Otherwise if from act 3, this can probably be a "welcome back" instead (also will need some additional pathing for coming from pre-act 3)

Monika After Story/game/screens.rpy
```diff
@@ -666,13 +666,127 @@ style quick_button_text_dark:
...
+            "I believe that you and Monika will be a great couple together."
```
multimokia, line 712:
```suggestion
            "I'm sure that you and Monika will be a great couple."
```

Monika After Story/game/screens.rpy
```diff
@@ -666,13 +666,127 @@ style quick_button_text_dark:
...
+
+            show chibika sad at mas_chflip_s(-1) onlayer screens zorder 10
+            "But if you'll call yourself like that..."
```
multimokia, line 715:
```suggestion
            "But if you call yourself names like that...{w=0.5}{nw}"
```

Monika After Story/game/screens.rpy
```diff
@@ -666,13 +666,127 @@ style quick_button_text_dark:
...
+
+            show chibika at sticker_hop onlayer screens zorder 10
+            "You won't win her heart!"
```
multimokia, line 718:
```suggestion
            extend "You won't win her heart!"
```

Monika After Story/game/screens.rpy
```diff
@@ -666,13 +666,127 @@ style quick_button_text_dark:
...
+
+            show chibika smile at mas_chmove(x=300, y=405, travel_time=1) onlayer screens zorder 10
+            "Rather just embaress her."
```
multimokia, line 721:
```suggestion
            "But just embarrass her instead..."
```

Monika After Story/game/screens.rpy
```diff
@@ -666,13 +666,127 @@ style quick_button_text_dark:
...
+
+            show chibika at mas_chlongjump(x=190, y=552, ymax=375, travel_time=0.8) onlayer screens zorder 10
+            window auto
+
+    else:
+        show chibika smile at mas_chflip(-1), mas_chmove(x=130, y=552, travel_time=0), sticker_hop onlayer screens zorder 10
+        "I don't think she would be comfortable with calling you like that."
```
multimokia, line 728:
```suggestion
        "I don't think she would be comfortable calling you that..."
```

Monika After Story/game/screens.rpy
```diff
@@ -666,13 +666,127 @@ style quick_button_text_dark:
...
+
+    "Let's choose something more appropriate."
```
multimokia, line 730:
```suggestion
    "Why don't choose something more appropriate instead."
```

Monika After Story/game/screens.rpy
Booplicate: It prompts to import a name only if you've not imported your saves. I thought that in this case we assume the player killed Monika.

multimokia: True

Issue #892, opened 2018-01-31
The trolley problem dialog - confusingly written?


MarkusRamikin:
>     m 3h "I wouldn't switch."
>     m 4 "That's the only logical choice, if you think about it."
>     m "It feels nice to say that you'd save the most people, doesn't it? But people only say that because nobody in the example matters to them."
>     m "If they actually cared about the sacrifice, they'd kill five people themselves in a heartbeat."
>     m 2 "Do you know anyone who would want a self-driving car to save other people by killing the driver? Didn't think so, [player]."

What's she on about, what does "If they actually cared about the sacrifice, they'd kill five people themselves in a heartbeat" mean? 

Is she talking about some rare version of the problem where, I dunno, the driver has to either sacrifice 5 people, or crash and sacrifice himself? I'm forced to guess. Because AFAIK the classic trolley problem is 5 other people on a track or 1 other person on a different track. And the nerve that the problem hits is that you have to pull the lever to save the 5 people, taking personal responsibility for killing that one poor schmuck.

therationalpi:
I think it's a version of the trolley problem where the train is currently heading towards 1 person you know, and the question is if you would pull the lever to kill 5 rando's instead. But I agree, it's not clear what she's talking about and that's not the canonical version of the trolley problem.

Needs a fix.

MarkusRamikin:
Well, here's my suggestion (borrowing from the entire current script for this topic):

"Oh, cool. I love thinking about these sorts of thought experiments."
"I guess we're talking about real people, right? I wouldn't have a particular preference if I knew they didn't have free will."
"Hmmm..."
"The classic trolley problem has us choose between letting the trolley run over 5 people, or switching to the track where only 1 person will be killed".
"I think this problem is very easy. Of course I would switch."
"It takes a rare kind of moral cowardice not to, just to avoid personal responsibility for that one person's death."
"A more interesting variant is if the one person is someone you care about. Your loved one."
"Like if it were you, [player]?"
"I wouldn't switch."
"That's the only logical choice, if you think about it."
"It feels nice to say that you'd save the most people, doesn't it? But people only say that because nobody in the example matters to them."
"If they actually cared about that one person, they'd kill the other five people."
"Do you know anyone who would want a self-driving car to save other people by killing the driver? Didn't think so, [player]."
"The truth is, humans are fine with sacrificing other lives to make themselves happy. They do it in little ways, every day."
[...]

PR #5770, opened 2020-05-14
Trolley topic rewrite

LooseEthics:
There have been multiple issues saying the topic is weird or confusing, but no one ever did anything about it. 
So here's a rewrite. Raw text taken from #892.

btw I've been meaning to ask this before but I forgot. Is there some sort of a process to submitting topics and contributing to the code in general? Or does one just make a pull request and wait for reviews?


multimokia:
Just pull requests

Monika After Story/game/script-topics.rpy
```diff
@@ -3641,13 +3641,14 @@ label monika_trolley:
...
     m 1euc "I guess we're talking about real people, right? I wouldn't have a particular preference if I knew they didn't have free will."
     m 1dsc "Hmmm..."
-    m 1eua "If we were talking about the classic problem, of course I would switch."
-    m 1rkc "But if we're talking about real people..."
-    m 3esc "I wouldn't switch."
-    m 4esc "That's the only logical choice, if you think about it."
-    m 1euc "It feels nice to say that you'd save the most people, doesn't it? But people only say that because nobody in the example matters to them."
-    m "If they actually cared about the person, they'd kill the other five in a heartbeat."
-    m 2tsb "Do you know anyone who would want a self-driving car to save other people by killing the driver? Didn't think so, [player]."
+    m 3esa "The classic trolley problem has us choose between letting the trolley run over 5 people, or switching to the track where only 1 person will be killed".
```
multimokia, line 3644:
```suggestion
    m 3esa "The classic trolley problem has us choose between letting the trolley run over 5 people, or switching to the track where only 1 person will be killed."
```

Monika After Story/game/script-topics.rpy
```diff
@@ -3641,13 +3641,14 @@ label monika_trolley:
...
+    m 1esb "I think this problem is too easy. Of course I would switch."
+    m 1esc "It takes a rare kind of moral cowardice not to save 4 lives, just to avoid personal responsibility for that one person's death."
```
Booplicate, line 3646: If we're talking about the classic variant, then switching will save 5 lives.

LooseEthics: There's 5 people on one track and 1 on the other. Either 1 dies or 5 do. This leaves you with net value of 4.

multimokia: Imo it should be `4 more lives` or just `5 lives`. Using the net value makes this more confusing and makes it seem like a different problem.

LooseEthics: Alright, point taken. 5 it is.

Monika After Story/game/script-topics.rpy
```diff
@@ -3641,13 +3641,14 @@ label monika_trolley:
...
-    m 1eua "If we were talking about the classic problem, of course I would switch."
-    m 1rkc "But if we're talking about real people..."
-    m 3esc "I wouldn't switch."
-    m 4esc "That's the only logical choice, if you think about it."
-    m 1euc "It feels nice to say that you'd save the most people, doesn't it? But people only say that because nobody in the example matters to them."
-    m "If they actually cared about the person, they'd kill the other five in a heartbeat."
-    m 2tsb "Do you know anyone who would want a self-driving car to save other people by killing the driver? Didn't think so, [player]."
+    m 3esa "The classic trolley problem has us choose between letting the trolley run over 5 people, or switching to the track where only 1 person will be killed."
+    m 1esb "I think this problem is too easy. Of course I would switch."
+    m 1esc "It takes a rare kind of moral cowardice not to save 5 lives, just to avoid personal responsibility for that one person's death."
```
Rai99, line 3646: If you ask me, I'd prefer not to call out and attack those people who wouldn't switch. Of course, not everyone is going to agree on this question, and I don't think we should attack them.

Monika After Story/game/script-topics.rpy
```diff
@@ -3641,13 +3641,14 @@ label monika_trolley:
...
     m 1dsc "Hmmm..."
-    m 1eua "If we were talking about the classic problem, of course I would switch."
-    m 1rkc "But if we're talking about real people..."
-    m 3esc "I wouldn't switch."
-    m 4esc "That's the only logical choice, if you think about it."
-    m 1euc "It feels nice to say that you'd save the most people, doesn't it? But people only say that because nobody in the example matters to them."
-    m "If they actually cared about the person, they'd kill the other five in a heartbeat."
-    m 2tsb "Do you know anyone who would want a self-driving car to save other people by killing the driver? Didn't think so, [player]."
+    m 3esa "The classic trolley problem has us choose between letting the trolley run over 5 people, or switching to the track where only 1 person will be killed."
+    m 1esb "I think this problem is too easy. Of course I would switch."
```
Rai99, line 3645: I might not say that it's too easy, or at least say that there might be more to it. If it was too easy, and everyone had the same obvious answer, it wouldn't be such a famous question.

LooseEthics: I suppose the phrasing here is kinda poor. I'd interpret this line more as Monika saying it's easy for her personally to answer, not necessarily in general. I feel like she's been quite clearly shown to be on the ends-justify-the-means side of the spectrum. There might be a bit of a personal bias showing here in that the choice seems almost trivially easy. As someone who'd pull the lever as well, I simply don't understand what kind of a mindset you'd need to have to choose otherwise, short of simply not caring.

Rai99: I interpreted the line as Monika saying that the question is incredibly easy, and the answer is obvious to all. True, Monika would likely hold the stance to switch, but the way it's worded sounds like she's saying it's an easy question to begin with, with an obvious one answer.

Monika After Story/game/script-topics.rpy
LooseEthics: Multiple things come to mind here, I'll try to sum it up as best as I can.
> I agree in that the wording here might be a bit too strong.
> As you said, this is a pretty polarizing topic. But I feel like that's more of a reason to keep something like this rather than not. When people actually sit down and think about these kinds of thought experiments, they usually feel rather strongly about the answer they arrive at. Monika would presumably be the same. She is her own character with her own opinions, not a safe space.
> I feel like the number of people who'd get upset at this is a good deal less than what one might think at first glance. The community here isn't the general populace, it takes a specific kind of person play this mod in the first place.
> How about adding a milder version behind a sensitive mode check? I'd say that gets us the best of both worlds.

Rai99: Whatever conclusion you come to with the question, you're probably going to feel extremely strongly about it, and I think that's fine. The easy answer most people would go with would be to pull the lever. I would just imagine that she'd at least be respectful to the other side rather than calling out their rare moral cowardice. 

Think back to the `monika_debate` topic. Having been in debate club, and knowing that a common pitfall is seeing their own side as superior, and ending up negatively criticizing the other side, making them feel attacked and accused of things like poor taste, I think she should keep this in mind, even in polarizing questions like this. 

As someone who wouldn't pull the lever, yes, I'll admit I must be very rare, but I accept and respect that so many other people would pull the lever for their own good reasons. I don't think Monika should call moral cowardice to those who don't agree with her.

Booplicate: One thing is having your own opinion on a problem, another thing is judging other people's opinions. Especially when you don't know why they have such opinion.
I don't think Monika would call the player like that only because of their choice. And she unintentionally (she doesn't know the player's opinion) might do that. It's not about will it upset one person or 10. It's a matter of why Monika wouldn't do so. IMO it's better to reword it.

jmwall24: Can I suggest possibly flipping this a little? Instead of having it come off as sounding like a personal attack to those who disagree, she can be just as strong, but word it as solely her opinion. For example: ` 'There is no way I could personally justify to myself saving one person at the expense of five others, just so I wouldn't have to accept responsibility for that one person's death.'`

Rai99:
For clarification, what's the difference between the "classic" trolley problem, and what we currently have?

Monika After Story/game/script-topics.rpy
Rai99: I think an easy fix would be to just remove that "too". 

Easy for Monika, but not so, so easy and obvious, that there's an expectation for everyone.

LooseEthics:
The classic problem has 6 generic people on the tracks, the other one discussed has 5 generic people on one track and someone the actor cares about on the other, i.e. the player.

LooseEthics:
Alright, I did some more reading on the topic and rewrote it. I tried to take all the issues raised here into account.
There's one more thing I thought about doing, but I decided to get feedback on that first: gating this whole topic behind a sensitive mode check. I know it's just a thought experiment, but we are talking about killing people here.

Booplicate:
Capture

LooseEthics:
Oh, my mistake.

Monika After Story/game/script-topics.rpy
```diff
@@ -3641,13 +3641,19 @@ label monika_trolley:
...
-    m 1eua "If we were talking about the classic problem, of course I would switch."
-    m 1rkc "But if we're talking about real people..."
-    m 3esc "I wouldn't switch."
-    m 4esc "That's the only logical choice, if you think about it."
-    m 1euc "It feels nice to say that you'd save the most people, doesn't it? But people only say that because nobody in the example matters to them."
-    m "If they actually cared about the person, they'd kill the other five in a heartbeat."
-    m 2tsb "Do you know anyone who would want a self-driving car to save other people by killing the driver? Didn't think so, [player]."
+    m 3esa "The classic trolley problem has us choose between letting the trolley run over five people, or switching to the track where only one person will be killed."
+    m 1esa "The problem is mostly famous due to how divisive it is."
+    m "Regardless of whether they'd pull the switch or not, most people believe that their choice simply must be the correct one."
```
Rai99, line 3646: I think this goes without saying for most famous questions, so I don't think we need to say this.

Monika After Story/game/script-topics.rpy
```diff
@@ -3641,13 +3641,19 @@ label monika_trolley:
...
+    m 1esc "There are even some who'd say the right thing to do is not take part in this scenario at all."
```
Rai99, line 3647: Again, I feel like you're singling out the people who have this stance, and dismissing it. 

Instead of criticizing other views, maybe you could have Monika mention the other views, with some of the reasoning behind them. 

For example, Two trains of thought for not pulling the lever are of course, not wanting to take direct responsibility for one person's death, or not wanting to pull because it isn't your place, and who are you to interfere with whatever is going on?

LooseEthics: I'm not trying to attack anyone here.
My intention here was to list the three stances one might take and point out that two of them are, for all intents and purposes, identical. Whether you choose to walk away from the problem, or to actively not pull the lever, the outcome is the same.
Looking at it from a lower level, I can see how the phrase `There are even some` might convey a bit of an accusatory tone.

Rai99: Yes, there are ultimately only 2 choices, and whatever you choose has to be one of the two. I agree that phrase makes it sound a bit more accusatory. 

Also, when Monika says "I don't think that really works", it almost sounds like she's saying it's the wrong course of action. There are 2 courses of action, but multiple different reasons behind each one, which is why I suggested you list a couple of the possible reasonings.

Monika After Story/game/script-topics.rpy
LooseEthics: That line is meant more as a lead-in for the next one than as a standalone. I feel like skipping over it would leave a discontinuity and I wasn't sure what else to put there.
But addressing your point here, you also have problems which, while having a single correct solution, are famous for being absurdly hard (this is more of a math thing than philosophy), or problems which are famous for straight up not having a solution, since trying to find one leads you to a logical contradiction. Things like Liar's or Russell's paradox.

Monika After Story/game/script-topics.rpy
LooseEthics: The 'I don't think that really works' refers to the fact that you're the actor in this scenario and no matter what you do you can't reduce yourself to the role of a bystander, since even refusing to carry out the role of the actor ultimately fulfills it.
Okay, that's a bit convoluted, but I hope I'm making sense here.

Rai99: I would say something like "but even then, that's the same as not pulling the lever". 

My issue, is the way that it's worded, it specifically calls out the party that doesn't act, and makes it sound like she's calling their logic flawed, despite still being a viable option.

LooseEthics: Alright, noted. Before I make the changes, is there anything else that feels off to you? Like I said, I have trouble understanding the other side of the argument, so having someone actually point these thing out to me is probably the only way I'll notice.

Rai99: The only other thing that bugs me right now comes from how confusing the original dialogue is, so I think it's a good idea that the topic is being clarified and explained better. 

`m 1eua "If we were talking about the classic problem, of course I would switch."`
`m 1rkc "But if we're talking about real people..."`
`m 3esc "I wouldn't switch."`

The reason I asked the difference between the classic problem and what we already have comes from this bit here. I always thought the classic problem did assume real people, and nowhere in the topic does it actually specify that the one person on the other track was someone important at all, so I was confused what she meant by "real people", especially considering she assumes real people at the beginning, chooses to switch in the classic problem, but not to in the case of real people. Does the classic problem not assume real people?

Apart from that, I might also change that one line at the beginning about "free will" to "real people". I think that sounds a bit more polite/nicer in a general sense, and is easier to understand.

LooseEthics: Those three lines are actually the main reason I decided to rewrite the topic. The classic problem assumes real people afaik. Generic, and for the purpose of the problem undistinguishable, but real. I'd say the free will bit is her referencing the other girls being controlled by the script.

Rai99: If the bit about free will is specifically referencing the other girls, and only them, I think it's fine, but in a more general sense, such as if anyone else were to say it, I think it's easier to understand if she said "real people" instead. I think it serves the same purpose, and is easier to understand.

LooseEthics: I suppose it does feel more natural like that.

PR #5791, opened 2020-05-16
Update tears up, eyes-left fade removal

multimokia:
Requesting all of you just so this looks good to all of us.

Thanks @Professor-JMO for doing the eyes + tear adjustments

## Changes:
- Tears up now have streams on the outside of the eye rather than the inside
- Eyes left now no longer basically fades into the skin

## Testing:
- Verify new sprites look good


multimokia:
@ThePotatoGuy also requesting a code review for this w/ the generator adjusts

tools/spritemaker.py
```diff
@@ -875,6 +875,15 @@ def run_gss(sprite_db, sprite_db_keys, quiet=False, sp_per_file=500):
         if not menutils.ask_continue():
             return
 
+    #Refresh this so we make all invalid sprites
```
ThePotatoGuy, line 878: what is this for?

multimokia: Running the spritemaker and just hitting `Generate Sprites` (in the case you just added an `image monika <code>` in the file won't generate it's corresponding ATL required staticsprite. The part there allows it to do so (needed to especially here so the tears could be updated)

PR #5795, opened 2020-05-18
ily2 TT protection

jmwall24:
Currently, if you TT backward right after a topic that returns "love" (or uses `MAS_ILY()`), the `I love you!` menu prompt will be stuck on `I love you too!` until you click it or say goodbye due to the `check_time` (in this case the current time) in `mas_passedILY()` being less than `persistent._mas_last_monika_ily`, therefore always making it less than 10 secs since she said 'I love you.' This adds a check so if `check_time` is before  `persistent._mas_last_monika_ily`, we return False and reset  `persistent._mas_last_monika_ily`.

## Testing

- verify the `I love you!`/`I love you too!` main menu prompt works as expected during normal usage.

- verify if you TT backward right after something that returns "love", `I love you too!` is not the prompt.

### DIFF
--- a/Monika After Story/game/definitions.rpy
+++ b/Monika After Story/game/definitions.rpy
@@ -3485,7 +3485,7 @@ init -991 python in mas_utils:
 
     def logcreate(filepath, append=False, flush=False, addversion=False):
         """
-        Creates a log at the given filepath. 
+        Creates a log at the given filepath.
         This also opens the log and sets raw_write to True.
         This also adds per version number if desired
 
@@ -3499,7 +3499,7 @@ init -991 python in mas_utils:
                 You dont need this if you create the log in runtime,
                 (Default: False)
 
-        RETURNS: created log object. 
+        RETURNS: created log object.
         """
         new_log = getMASLog(filepath, append=append, flush=flush)
         new_log.open()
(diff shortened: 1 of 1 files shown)


Monika After Story/game/definitions.rpy
```diff
@@ -6732,7 +6732,13 @@ init python:
         """
         if check_time is None:
             check_time = datetime.datetime.now()
-        return persistent._mas_last_monika_ily is not None and (check_time - persistent._mas_last_monika_ily) <= pass_time
+
+        # if a backward TT is detected here, return False and reset persistent._mas_last_monika_ily
+        if persistent._mas_last_monika_ily is None or persistent._mas_last_monika_ily > check_time:
+            persistent._mas_last_monika_ily = None
```
ThePotatoGuy, line 6738: do we call this function with a value passed into `check_time` anywhere? 

The `check_time` parameter doesn't make sense with this change since you could purposefully pass in an old `check_time` and change game state.
If it's unused, we should dummy out the parameter.
If it's lightly used, then we should make a new function for `check_time` usage and replace the appropriate lines.
If it's heavily used, then we should consider a new function that doesn't use `check_time` at all.

jmwall24: this func is only used once, and that's when we load the Talk menu to check for this one purpose. It honestly doesn't need check_time as a passed in param, just as an internal var that is set to now

PR #5816, opened 2020-05-23
Better substitution for strings

Booplicate:
This fixes a bug that doesn't allow us to use negative indexes for some reason.
For example this will crash you. I fixed it.
```renpy
$ items = ["a", "b", "c"]
m "The last item is [items[-1]]"
```
Also allows us to make function calls via `eval`.
For example:
```renpy
$ def return_love(): return "love"
m "I [return_love()] you!"
```
It does support different stores and passing in arguments/keyword arguments.
```renpy
# All of these are valid options, it doesn't make sense, but for the sake of example lol
python in new_store:
    def there_re_s(number):
        if number > 1:
                return "There're"
        return "There's"
$ n = 2
m "[new_store.there_re_s(n)] [n] 'of smth'!"

# Or this
python in new_store:
    def return_monika(*args, **kwargs):
        return "Monika"
$ _line = renpy.substitute("My name is [return_monika('arg1', kwarg1='kwarg1', kwarg2=None)].", scope=new_store.__dict__)
```

### Testing:
- verify that base substitution still works.
- verify that we're able to use negative indexes for iterables.
- test calling function from the base store, custom stores, with arguments and w/o them. Using both the `say` statement and the `renpy.substitute` function.

### DIFF
--- a/Monika After Story/game/definitions.rpy
+++ b/Monika After Story/game/definitions.rpy
@@ -7,6 +7,8 @@ init 1 python:
 
 ### Overrides of core renpy things
 python early in mas_overrides:
+    from store import *
+
     def dummy(*args, **kwargs):
         """
         Dummy function that does nothing
--- a/Monika After Story/game/definitions.rpy
+++ b/Monika After Story/game/definitions.rpy
@@ -5,6 +5,62 @@ define config.developer = False #This is the flag for Developer tools
 init 1 python:
     persistent.steam = "steamapps" in config.basedir.lower()
 
+### Overrides of core renpy things
+python early in mas_overrides:
+    def dummy(*args, **kwargs):
+        """
+        Dummy function that does nothing
+        """
+        return
+
+    # clear this so no more traceback. We expect node loops anyway
+    renpy.execution.check_infinite_loop = dummy
+
+    def get_field_override(self, field_name, args, kwargs):
+        """
+        Originally this method returns objects by references
+        This override allows us to eval functions when appropriate, e.g. "Text [myfunc(arg1)]."
+        and use negative indexes for iterables, e.g. "Text [_list[-2]]."
+
+        IN:
+            field_name - the reference to the object
+            args - not sure, but renpy doesn't use it (passes in an empty tuple)
+            kwargs - the store modules where renpy will look for the object
+
+        OUT:
+            tuple of the object and its key
+        """
+        first, rest = field_name._formatter_field_name_split()
+
+        # if it's a function call, we eval it
+        if "(" in field_name:
+            # if the module/func isn't in the def store,
+            # we use the one in kwargs to call the func from there
+            if isinstance(kwargs, renpy.substitutions.MultipleDict):
+                _store = [name for name, store in sys.modules.iteritems() if store is kwargs.dicts[0]][0]
+            else:
+                _store = "renpy.store"
+
+            obj = eval("{0}.{1}".format(_store, field_name))
+
+        # otherwise just get the object by reference
+        else:
+            obj = self.get_value(first, args, kwargs)
+
+            for is_attr, i in rest:
+                if is_attr:
+                    obj = getattr(obj, i)
+
+                else:
+                    if not isinstance(i, long):
+                        i = long(i)
+                    obj = obj[i]
+
+        return obj, first
+
+    # allows us to use a more advanced string formatting
+    renpy.substitutions.Formatter.get_field = get_field_override
+
 python early:
     import singleton
     me = singleton.SingleInstance()
@@ -18,16 +74,6 @@ python early:
     # uncomment when needed
     import traceback
     _dev_tb_list = []
-
-    def dummy(*args, **kwargs):
-        """
-        Dummy function that does nothing
-        """
-        return
-
-    # clear this so no more traceback. We expect node loops anyway
-    renpy.execution.check_infinite_loop = dummy
-
 
 # uncomment this if you want syntax highlighting support on vim
 # init -1 python:
--- a/Monika After Story/game/definitions.rpy
+++ b/Monika After Story/game/definitions.rpy
@@ -32,21 +32,56 @@ python early in mas_overrides:
         OUT:
             tuple of the object and its key
         """
-        first, rest = field_name._formatter_field_name_split()
-
-        # if it's a function call, we eval it
+        # if it's a function, we eval it
         if "(" in field_name:
-            # if the module/func isn't in the def store,
-            # we use the one in kwargs to call the func from there
-            if isinstance(kwargs, renpy.substitutions.MultipleDict):
-                _store = [name for name, store in sys.modules.iteritems() if store is kwargs.dicts[0]][0]
+            # id to split the string
+            split_index = field_name.index("(")
+            # get the func's name
+            func_name = field_name[:split_index]
+            # get the arg and kwargs
+            args = field_name[split_index:]
+
+            # it still may include store modules, get the 1st if there's any
+            if "." in func_name:
+                func_store_name, func_name = func_name.split(".", 1)
+                first = func_store_name
+                func_store_name += "."
+
             else:
-                _store = "renpy.store"
+                func_store_name = ""
+                first = func_name
 
-            obj = eval("{0}.{1}".format(_store, field_name))
+            # now we find the store's name to use in eval
+            # iterate through all the scopes you passed in to find the one
+            # where the function/module was defined
+            if isinstance(kwargs, renpy.substitutions.MultipleDict):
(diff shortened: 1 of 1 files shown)


Booplicate:
I have dialogue presets in my PR, and some of them require substitution, while others don't. Since I use all of those quips from one method (it's like a separate system), it would be quite clumsy to use `.format` only for some of them. RenPy allows to use `something[1]`, but not `something[-1]`. To me it looks like a bug, I wouldn't say it *really* works. But it's fixed with 2 lines of code, and I'm sure it won't break anything. It's basically converting into `int` where the param should be `int`, but it's `str` for some reason. That's what crashes. And it's dumb.

Could I go around of this bug? Yes, by adding another property to my class. Or adding a bunch of checks to the method. But I thought it would be gross/messy, and instead suggested cleaner solution, which may be helpful in the future too.

Regarding functions calls, multi needs them, so I thought I will include them here too. I'm not going to use it yet, but... From what I saw some other languages support such substitution. And using a temp var works until you we need to do it more than 1 time during a dialogue. Because then it will become quite gross, too. Basically this may allow us to write more clear code, and instead of 2+ lines, just put it in in a dialogue line and do it "on the fly".

The override I suggest is based on the original function, and only contains additions (check for `type` and path for `eval`ing). That means the old code will be unaffected by this, and will work as it does now. I understand why you would like to keep the RenPy code, though. But from my point of view it's like improvements to the existing substitution.

ThePotatoGuy:
Have you looked at `MASQuipList`? It doesn't support what you want right now, but its ours and can be modified to include substitution tables or to execute things before spitting out a string. 

Then you would just do:
```python
$ quip_type, quip = quiplist.quip()
m "[quip]"
```
Could even modify quip to pass generic params in.

Booplicate:
Yeah, I have. `MASQuipList` won't really suit for my purposes w/o some big changes to it. Or subsclassing it. So back then I decided I won't touch it and instead make my own methods in the game class to handle stuff. Because it's very specific.

Maybe I should put some of my ideas for quips here? Yeah, this's the way it's already working/my concept:
- Using labels isn't a variant, it's messy and would require `renpy.call`s from py code.
- I'm using multiline quips and text tags.
- I have a map of maps of different quips. The first accepts the type of a quip. The second the seen_count which I use for that type.
- Some maps share some base quips with other maps. But also have their own unique lines.
- Some quips can trigger other quips (which also rng based) if needed.
- Depending on the situation in the game, I can add special quips to the already chosen ones. Which makes it more "life-like".
- If I use the `format` method, I would need to escape all text tags (even where I don't need to format anything). Using it also would make the code more entangled since I have to pass in vars. So ideally it should be done with `renpy.substitute`. And then we still will have to deal with the inconsistency and inconvenience with the current substitution.
- It is working perfectly right now, and I would like not to redo it from scratch. Especially since it would require to make the new system work with my classes.

The only thing I wanted is to make it more optimized with this override. Basically all I needed is stuff like `"I'd like to set [my_store.my_game.pile[-1]].colour!"` for some quips, but not for the others. So it would keep the system clear, simple and generalized. But if you think it's not possible, then it'd be better to just use a crutch than redo it.

But even if we forget about my problem, the bug is still here and the next time we would have to go around again.

ThePotatoGuy:
Okay, I see you have a complex quip structure that is clearly not as simple as self-contained strings/labels. I wonder why it needs to be that complicated, with what sounds like semi-inheritance and lots of dynamic content, but I'll save that for when I'm reviewing the game PR. I will mention now that I will probably request that you generalize whatever system you are making, so heads up on that.

Regarding the lack of a negative indexing option, I'd argue that isn't a bug, just a lack of a feature. Negative indexing isn't a common feature for most languages, and there's no reason renpy has to support all of nuances python has.

Regardless, I'll go ahead and continue reviewing this.

Monika After Story/game/definitions.rpy
```diff
@@ -5,6 +5,102 @@ define config.developer = False #This is the flag for Developer tools
...
+
+                else:
+                    if not isinstance(i, long):
+                        i = long(i)
+                    obj = obj[i]
+
+        return obj, first
+
+    # allows us to use a more advanced string formatting
+    renpy.substitutions.Formatter.get_field = get_field_override
```
ThePotatoGuy, line 102: so looking at where this lies, `Formatter` is an extension of the `string.Formatter` class, which is a python class. Rather than overwrite the function (which in reality is just adding a definition to renpy's `Formatter` so it doesn't fallback to python's formatter), I think you should just extend `Formatter`, define `get_field` on your extended class, and set `renpy.substitutions.formatter` equal to an instance of your `Formatter` extended class.

Monika After Story/game/definitions.rpy
```diff
@@ -5,6 +5,102 @@ define config.developer = False #This is the flag for Developer tools
...
+            args - not sure, but renpy doesn't use it (passes in an empty tuple)
+            kwargs - the store modules where renpy will look for the object
+
+        OUT:
+            tuple of the object and its key
+        """
+        # if it's a function, we eval it
+        if "(" in field_name:
+            # id to split the string
+            split_index = field_name.index("(")
```
ThePotatoGuy, line 38: Is `field_name` a string? If so, just use `partition` since you are splitting a string into two parts.
```python
func_name, paren, args = field_name.partition("(")
```

Monika After Story/game/definitions.rpy
```diff
@@ -5,6 +5,102 @@ define config.developer = False #This is the flag for Developer tools
...
+                for scope in kwargs.dicts:
+                    if first in scope:
+                        stores_names_list = [
+                            store_obj_name
+                            for store_obj_name, store_obj in sys.modules.iteritems()
+                            if store_obj and store_obj.__dict__ is scope
+                        ]
+                        if stores_names_list:
+                            scope_store_name = "renpy.{0}.".format(stores_names_list[0])
+                            break
```
ThePotatoGuy, line 67: don't use `break`, please. (or a `for-else` loop, which I have actually never seen before, but i would argue is less readable as its easy to to mistake the `else` as part of a preceding `if`). Make this a separate function so it could be used elsewhere and use `return`.

Monika After Story/game/definitions.rpy
```diff
@@ -5,6 +5,102 @@ define config.developer = False #This is the flag for Developer tools
 init 1 python:
     persistent.steam = "steamapps" in config.basedir.lower()
 
+### Overrides of core renpy things
+python early in mas_overrides:
+    from store import *
```
ThePotatoGuy, line 10: What do you need to import in here? We should only import what we need, or namespace the import.

Booplicate: We import the entire store so we can use its content in functions calls. If we imported just store, we'd have to put `store.` before passing in each argument. Splitting args and kwargs to add the `store.` bit within the method is not possible.
```renpy
m "[func(arg1, arg2)]"
# becomes
m "[func(store.arg1, store.arg2)]"
```
Tbh now I think that `mas_overrides` is useless.

ThePotatoGuy: I agree, keep it in the global store if you need everything from it anyway.

Monika After Story/game/definitions.rpy
Booplicate: Shall I do the same for this line too?
```python
func_store_name, func_name = func_name.split(".", 1)
```

ThePotatoGuy: sure

Monika After Story/game/definitions.rpy
```diff
@@ -28,6 +29,108 @@ python early:
...
+                    func_store_name = ""
+                    dot = ""
+                    first = func_name
+
+                # now we find the store's name to use in eval
+                if isinstance(kwargs, renpy.substitutions.MultipleDict):
+                    scope_store_name = _getStoreNameForObject(first, *kwargs.dicts)
+
+                else:
+                    scope_store_name = _getStoreNameForObject(first, kwargs)
```
ThePotatoGuy, line 97: if partition doesn't find the string, it returns `("string you called it on", "", "")`, so you can drop the if here.

```suggestion
                # it still may include store modules, get the first if there's any
                func_store_name, dot, func_name = func_name.partition(".")

                # now we find the store's name to use in eval
                if isinstance(kwargs, renpy.substitutions.MultipleDict):
                    scope_store_name = _getStoreNameForObject(func_store_name, *kwargs.dicts)

                else:
                    scope_store_name = _getStoreNameForObject(func_store_name, kwargs)
```

Booplicate: I still need to return `first`, the OG function does so. But I can get rid of the dupe code.

PR #5831, opened 2020-05-28
Reset consumable if it's prepping but it's not time to have

Legendkiller21:
There is a chance for a consumable to have a `prep_time` but get stuck because the finish prepping event somehow either was not queued or lost its conditional. 

This adds an extra check in checking logic so that we can reset the consumable if we have a `prep_time` but the finish prep event has no conditionals. We now also reset the consumable if we started prepping the consumable but left and came back on a time where we shouldn't be able to have it.

## Testing:
- Verify that calling the logic resets the appropriate consumable type if a consumable has a `prep_time` but the finish prep event has no conditionals or it's no longer time to have this consumable.
  Note: A complete reset will only happen if there are no available consumables for the current time. If there are available consumables the logic will try to queue a new one.
- Verify no crashes.

### DIFF
--- a/Monika After Story/game/zz_consumables.rpy
+++ b/Monika After Story/game/zz_consumables.rpy
@@ -1023,8 +1023,11 @@ init 5 python:
             #Verify persist data
             MASConsumable._validatePersistentData(_type)
 
+            if curr_cons:
+                curr_finish_prep_ev = mas_getEV(curr_cons.finish_prep_evl)
+
             #Reset if we're having a consumable we shouldn't be having now and we opened the game after its consume time
-            #or if we're still prepping something but we're past the consumable's prepping time
+            #or if we're still prepping something but the consumable's finish prepping event doesn't have conditionals
             if (
                 (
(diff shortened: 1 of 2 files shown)


Monika After Story/game/zz_consumables.rpy
```diff
@@ -1042,12 +1023,23 @@ init 5 python:
...
+                    MASConsumable._isHaving(_type)
+                    and (
+                        not MASConsumable._isStillCons(_type)
+                        and mas_getCurrSeshStart() > persistent._mas_current_consumable[_type]["consume_time"]
+                    )
+                )
+                or (
+                    persistent._mas_current_consumable[_type]["prep_time"] is not None
+                    and curr_finish_prep_ev
+                    and curr_finish_prep_ev.conditional is None
```
ThePotatoGuy, line 1042: since this is getting funky, it would be better to create to a static function like `shouldReset`  and just pass in the appropriate parameters from here. Especially since `curr_finish_prep_ev` is only used for this check and not elsewhere in this function.

PR #5840, opened 2020-05-30
New brb - workout

LooseEthics:
DYEL
No expressions yet, will add those later.
Ignore the previous pr, desktop is doing stupid things.


lunulae:
I'm wondering if the time frame for a long workout should be longer. I tend to have some extra time accounted for since I usually go to a gym to work out.
(Although, when I think about it, my issue might be better served by having a farewell alternative for the gym in addition to the brb, using this option for at-home stuff.)

LooseEthics:
Yeah, this is mostly meant for exercising at home. An hour really isn't a lot if you were to add commute to that. An actual goodbye would be better if you're leaving the house.

PR #5841, opened 2020-05-30
Window Reactions Request

enderblaze42:
I use a Mac. I didn't know that Monika could react to windows you have open, since it's not an option on Mac. If possible, I'd appreciate it if window reactions could be made available for Mac users as well as Windows users. 
That aside, this mod is really fun, and I love it! Keep up the great work!


multimokia:
Please open an issue with your suggestions rather than a pull request.

Pull requests are for submitting new content to the mod.

PR #5852, opened 2020-05-31
Update farewells to work similar to greetings

multimokia:
Farewells used an outdated method for selection. This PR updates them to use a similar selection algorithm to greetings.

In addition to this, I have also created a new rule type, `MASProbabilityRule`, which is a number between 1 and 10 (since this is designed to work with the function *store.mas_utils.*`weightedChoice()`

An analogy for how the function works is as follows:
Priority number essentially corresponds to the amount of times the choice given is repeated in the list for a `random.choice` to pick from.

So a probability rule with value 10 doesn't mean it's guaranteed to be selected, however it is more likely.

I have also rerandomed `bye_leaving_already` and added a conditional for it to show for sessions under 20 minutes in length.

## Testing:
- Verify that farewells work properly
- Verify that farewells still follow time rules
- Verify that it's possible to get generic farewells during ranges of time rules
- Verify no crashes

### DIFF
--- a/Monika After Story/game/definitions.rpy
+++ b/Monika After Story/game/definitions.rpy
@@ -3743,7 +3743,6 @@ init -991 python in mas_utils:
     mas_log.raw_write = True
     mas_log.write("VERSION: {0}\n".format(store.persistent.version_number))
 
-
     def weightedChoice(choice_weight_tuple_list):
         """
         Returns a random item based on weighting.
@@ -3759,7 +3758,13 @@ init -991 python in mas_utils:
         if not choice_weight_tuple_list:
             return None
 
+        #Firstly, sort the choice_weight_tuple_list
+        choice_weight_tuple_list.sort(key=lambda x: x[1])
+
+        #Now split our tuples into individual lists for choices and weights
         choices, weights = zip(*choice_weight_tuple_list)
+
+        #Some var setup
         total_weight = 0
         cumulative_weights = list()
 
@@ -3768,12 +3773,14 @@ init -991 python in mas_utils:
             total_weight += weight
             cumulative_weights.append(total_weight)
 
-        #Now bisect the cumulative weight using a random point
+        #NOTE: At first glance this useage of bisect seems incorrect, however it is used to find the closest weight
+        #To the randomly selected weight. This is used to return the appropriate choice.
         r_index = bisect(
             cumulative_weights,
             renpy.random.random() * total_weight
         )
 
+        #And return the weighted choice
         return choices[r_index]
 
 init -100 python in mas_utils:
--- a/Monika After Story/game/definitions.rpy
+++ b/Monika After Story/game/definitions.rpy
@@ -3288,6 +3288,7 @@ init -991 python in mas_utils:
     #import tempfile
     from os.path import expanduser
     from renpy.log import LogFile
+    from bisect import bisect
 
     # LOG messges
     _mas__failrm = "[ERROR] Failed remove: '{0}' | {1}\n"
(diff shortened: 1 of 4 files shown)


Monika After Story/game/definitions.rpy
```diff
@@ -3743,6 +3744,37 @@ init -991 python in mas_utils:
...
+        total_weight = 0
+        cumulative_weights = list()
+
+        #Now we collect all the weights and geneate a cumulative and total weight amount
+        for weight in weights:
+            total_weight += weight
+            cumulative_weights.append(total_weight)
+
+        #Now bisect the cumulative weight using a random point
+        r_index = bisect(
```
ThePotatoGuy, line 3772: from the python docs on `bisect`: `This module provides support for maintaining a list in sorted order without having to sort the list after each insertion`. Based on this documentation, your use of `bisect` seems incorrect.

But after scrolling down the page, I see what you are trying to do here (numeric table lookups based on a set of breakpoints - weights). Can you add a more descriptive comment(s) here on what `bisect` is doing for you in this context? aka something like "bisect finds the index of closest weight to the random weight that we select, which is then used to return the appropriate choice"?

Additionally, doesn't `cumulative_weights` (and the choices array) need to be sorted for this to work right? `bisect` doesn't do sorting, it just acts on an already sorted list. If you need sorting, there is the `insort` functions in `bisect`, otherwise I have generalized insertion sort functions in `filtering-backgrounds`, I could move them out if needed.

Monika After Story/game/script-farewells.rpy
```diff
@@ -13,101 +13,118 @@
...
         """
+        # NOTE: new rules:
+        #   eval in this order:
+        #   1. unlocked
+        #   2. not pooled
+        #   3. aff_range unlocked
+        #   4. conditional
+        #   5. priority (lower or same is True)
+        #   6. all rules
+        #       NOTE: this is never cleared. Please limit use of this
```
ThePotatoGuy, line 45: this note applies to `conditional`. Additionally, since `conditional` is supposed to be a less likely to be used property, it should be last in the chain.

Monika After Story/game/event-rules.rpy
```diff
@@ -714,6 +715,63 @@ init -1 python:
...
+            if probability is None:
+                probability = MASProbabilityRule.DEF_PROBABILITY
+
+            elif probability > 10:
+                probability = 10
+
+            elif probability < 1:
+                probability = 1
+
+            if type(probability) is not int:
```
ThePotatoGuy, line 749:
```suggestion
            if not store.mas_ev_data_ver._verify_int(probability, allow_none=False):
```
I would actually put this in for the None check and just default the default probability if given an invalid type. Maybe log to the `mas_log` if desired.

Monika After Story/game/event-rules.rpy
```diff
@@ -714,6 +715,63 @@ init -1 python:
             return ev.rules.get(EV_RULE_PRIORITY, MASPriorityRule.DEF_PRIORITY)
 
 
+    class MASProbabilityRule(object):
+        """
+        Static class used to create probability rules.
+
+        Priority rules are just integers that determine the probability of something being selected.
```
ThePotatoGuy, line 722:
```suggestion
        Probability rules are just integers that determine the probability of something being selected.

```

Assuming this is what you mean.

Can you also give an example in this comment of what it means to use this? Like if you pick 6, how does that affect the selection process? Is this 6 times more likely to be chosen? Does this set the weighting of this selection as if this item was copied 6 times into a list?

PR #5855, opened 2020-06-01
content

SamFavred:
make it so you can rickroll monika


Booplicate:
From [the contributing guidelines](https://github.com/Monika-After-Story/MonikaModDev/wiki/Contributing-Guidelines#pitfalls-to-avoid):
> Please do not submit meme-related content, and only include pop-culture references when they seem appropriate to Monika's character.

Also for suggestions open issues, not pull requests.

PR #5859, opened 2020-06-02
New topic - Asimov's laws

LooseEthics:


Monika After Story/game/script-topics.rpy
```diff
@@ -14745,3 +14764,47 @@ label monika_fear:
...
+    m 4dud "{i}A robot shall not harm a human or, through inaction, allow a human to come to harm.{/i}"
+    m 2esb "To a, well, human this is pretty straightforward."
+    m 2esa "But when you try to formulate it in a way a machine can understand, you start to run into trouble."
+    m 1esa "You have to make precise definitions for everything, which isn't always easy. Like,{w=0.1}"
+    extend 1esc " how do you define a human?"
+
+    if monika_chr.is_wearing_acs(mas_acs_quetzalplushie):
+        m 3esb "I think we can both agree that you're a human, I'm a human, and that the adorable green friend I have sitting on my desk isn't."
+    else:
+        m 3esb "I think we can both agree that you're a human, I'm a human, and that the monitor on your desk isn't."
```
Daiozen, line 14795: Just my two cents on the subjects, but I think it'd be better to rephrase it like this:
`I think we can both assume I'm a human, you're a human, and that...`
The fact she aknowledges she could be `uploaded into a mechanical body` makes me think she likes to think of herself as human, but also aknowledges she might not completely be from our definition.
The verb 'agree' make it sound like her being fully human is a definite truth, and might close of any further reflexions on the subject of her humanity.

LooseEthics: Good point. Further exploring Monika's humanity would definitely be a worthwhile subject.

PR #5901, opened 2020-06-11
Make renpy loader consistent across OSes

multimokia:
Linking [#9](https://github.com/multimokia/MAS-Submods/issues/9) From my submods repository.

Renpy's loader struggles to load things correctly when given a full filepath (this causes issues in loading from a `MASDockingStation` object or when using `renpy.config.basedir` as it strips the `/` character before the directory.

This PR removes the `lstrip` function from key loader functions and allows it to work as it does on Windows, on other OSes.

# Testing:
- Verify Windows loading still works correctly (songs/sounds/etc.)
- Verify Linux can load from basedir (try putting an mp3 file in the basedir or a subfolder in the basedir and then use `play_song(renpy.config.basedir + "/path/to/song.mp3")`
- [Need Community Help] - I cannot get a MacOS VM working for myself to test this like I've tested the previous one.
  - If any community member has a Mac or a functional MacOS VM, please help testing this PR.
  - That or @aldoram5 if you could help out, that'd be appreciated.

### DIFF
--- a/Monika After Story/game/definitions.rpy
+++ b/Monika After Story/game/definitions.rpy
@@ -6,7 +6,6 @@ init 1 python:
     persistent.steam = "steamapps" in config.basedir.lower()
 
 python early:
-    import re
     import singleton
     me = singleton.SingleInstance()
     # define the zorders
@@ -34,7 +33,7 @@ python early:
         if renpy.config.reject_backslash and "\\" in name:
             raise Exception("Backslash in filename, use '/' instead: %r" % name)
 
-        name = re.sub(r'/+', '/', name)
+        name = renpy.re.sub(r'/+', '/', name)
 
         for p in renpy.loader.get_prefixes():
             rv = renpy.loader.load_core(p + name)
--- a/Monika After Story/game/definitions.rpy
+++ b/Monika After Story/game/definitions.rpy
@@ -6,6 +6,7 @@ init 1 python:
     persistent.steam = "steamapps" in config.basedir.lower()
 
 python early:
+    import re
     import singleton
     me = singleton.SingleInstance()
     # define the zorders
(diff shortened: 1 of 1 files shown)


ThePotatoGuy:
Couple of questions:
1. Is there any reasoning behind renpy's loader using lstrip? Does it imply that we should have only been passing relative paths and not absolutes?
2. Has this been fixed in later renpy versions? If so, can we just copy the code?

multimokia:
Seems this is still on upstream Ren'Py. Not seeing any reason, but the assumption is to force relative filepaths.

Issue is this creates an inconsistency and creates the false assumption that because a full filepath works fine on Windows, it'll work on other operating systems (Which as pointed out before, is an issue when using `MASDockingStation.station` or `renpy.config.basedir`/`renpy.config.gamedir`/etc, (since these use full filepaths, and a dockingstation object is more efficient for collecting files in a directory)), which puts an arbitrary limitation for anyone who wishes to use these to dynamically load files.

aldoram5:
I've done all the tests as per @multimokia 's request. Everything works as expected on macOS Catalina's most recent version at the time of testing.

multimokia:
Close but not quite. Didn't use `_trackpackage` but merely built a filepath based on the `station` prop and the added file.

Regardless, my main problem with the overarching behaviour here is the fact that we're an open source repo and people coming from different programming backgrounds may not know of the nuance. Assume that an absolute is fine for all cases because it works just fine on windows (and it would under any standard circumstance in any other language). Hence why this is (in my view) a bug.

I get what you mean about it being harder to update, but we do also keep our overrides contained to a specific location, so we know that we should look over them again upon an engine update.

ThePotatoGuy:
>my main problem with the overarching behaviour here is the fact that we're an open source repo and people coming from different programming backgrounds may not know of the nuance.

Fair point, but all the renpy docs use relative paths, so it's not completely ambiguous that only relative paths are accepted. It's just not very clear that absolute paths don't work cross-platform.

> I get what you mean about it being harder to update, but we do also keep our overrides contained to a specific location, so we know that we should look over them again upon an engine update.

The location isn't the problem, the problem is when we continually add small overrides until there's a large amount of code that has to be checked. Since the infinite loop override, there have been 3 PRs to override renpy functions. Just because we did it once (and now twice with the formatter replacment) doesn't mean we should look to overrides as the first solution to our problems.

With this, the solution should have been extracting the relative path from the station path and storing as a separate attribute so it can be used for cases like yours. This isn't difficult to do since basedir will exist in the path unless you are retrieving files outside of the basedir (which is questionable to do at best).

**However**, I will let this go through if you submit the change to the renpy repo and it gets accepted. That way we can just remove our override when upgrading to the appropriate version.

multimokia:
>The location isn't the problem, the problem is when we continually add small overrides until there's a large amount of code that has to be checked. Since the infinite loop override, there have been 3 PRs to override renpy functions. Just because we did it once (and now twice with the formatter replacment) doesn't mean we should look to overrides as the first solution to our problems.

Fair point there.

I guess we do both? For now make amends to `_trackpackage` to also return a relative path so specific cases can still use it, and I'll also submit a PR to the upstream renpy development repository making the same changes

PR #5925, opened 2020-06-15
New greeting - code help

LooseEthics:
I've noticed some people on the community discord saying that some more interactivity would be nice. This greeting is a gated, one time thing with a zero depth, inconsequential choice, but it's a something.
~~Currently the conditional raises a NameError for reasons I can't figure out. Traceback [here](https://github.com/Monika-After-Story/MonikaModDev/files/4779560/traceback.txt).~~ Turns out the reason was me being an idiot.


Monika After Story/game/script-greetings.rpy
```diff
@@ -3771,7 +3771,7 @@ init 5 python:
         Event(
             persistent.greeting_database,
             eventlabel="greeting_code_help",
-            conditional="persistent._mas_pm_has_code_experience = True",
+            conditional="persistent._mas_pm_has_code_experience is True",
```
scpketer, line 3774: `is` is used for type comparison, or as `is None`. Use `==`.

multimokia:
Conditional raised the error because you didn't default the persistent value. While accessing it does return None, it's value isn't there, so doing an `is` comparison is essentially like checking a var which doesn't exist against a value.

Issue #5935, opened 2020-06-17
[Suggestion] Regular expressions tutorial


scpketer:
I've been through Monika's Python tips (or guides? Not really sure), and an idea crossed my mind - what if Monika taught you how to use regexes?

I'm quite acquainted with those, so I think I could write an explaining topic about that (or multiple 'lessons' from basics to more advanced stuff, like with Python tips, but I don't really know how you guys made those) - but first I want to know if something like this will be okay to add, 'cause, well, most people think regexes are extremely complicated and so on...

If that's okay - great, I'll get to work then.

Freshairkaboom:
I don't even know what regexes are yet, but I'd love anything Monika could teach me about using Python :D I'm probably going to learn it all before this gets added, but additions are just as much for newcomers as they are for those who are already playing the mod, so I'd welcome any help they'd get.

Also, when it comes to tips like that, then I think you don't need to pull out so much resources besides just making dialogue and labeling it, but I'm a newbie programmer and don't know what the code requires, so don't quote me on that :D

ThePotatoGuy:
regex sounds good but we might need a prerequisite of more advanced string stuff like split/partition/splicing. ill double check this later.

the tips live in script-python so take a look at that to see how build the tips.

scpketer:
> I don't even know what regexes are yet

"Basically, kind of a big deal" - an advanced string matching and substring extraction tool.

> we might need a prerequisite of more advanced string stuff like split/partition/splicing

Might as well bring these on.

Freshairkaboom:
> 
> 
> > I don't even know what regexes are yet
> 
> An advanced string matching and substring extraction tool, but it might be really confusing to some.

Yeah...I'm a beginner, I don't understand what string matching is, I don't know what a substring is, extracting tool is kind of self explanatory though.

scpketer:
Well, you can perform a simple string comparison with `"one" == "another"`, but this will only do for exact matches. 

With regular expressions you can perform something more complex like `"one" <something like> "another"` with `re.search(".*o.+e.*", "another")`

Freshairkaboom:
You may as well have been speaking greek to me there. I only understood the first sentence.

scpketer:
`re.search` returns search object, but if you don't need to extract matching occurences from strings you can just check if it's not None

ThePotatoGuy:
Do you think you can write a prereq topic involving these string functions:
* `str.find` (`lfind`/`rfind`)
[...]

PR #5947, opened 2020-06-21
Add 'Strings' Python lesson

scpketer:
Fulfilling a [request for string basics](https://github.com/Monika-After-Story/MonikaModDev/issues/5935#issuecomment-646692763).

It actually lacks some of the requested topics (like on `find`, `index` and `splitlines`) as it's still incomplete, and I need a little help with expressions and phrasing (maybe even code examples for string functions tutorial).

### DIFF
--- a/Monika After Story/game/script-python.rpy
+++ b/Monika After Story/game/script-python.rpy
@@ -881,8 +881,116 @@ label monika_ptod_tip010:
     # evaluation order and short circuting
     return
 
+###############################################################################
+init 5 python:
+    addEvent(
+        Event(
+            persistent.event_database,
+            eventlabel="monika_ptod_tip011",
+            category=["python tips"],
+            prompt="Strings",
+            pool=True,
+            conditional="store.mas_ptod.has_day_past_tip(6)",
+            action=EV_ACT_UNLOCK,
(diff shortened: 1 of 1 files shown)


scpketer:
What I need help with is face expressions (this whole topic is all `1eua`, `3eua`, `1eub`, `3eub` and just one `1hua` - that's it! Probably need some more.) and `find`/`rfind` functions - I have ideas on how to explain those to a player, but I don't know where to locate those (maybe after slicing/subscription, because they are related to string indexes as well)

scpketer:
About `find` and `index` - I think we shouldn't introduce `index` to a player since it throws an exception, and afaik we don't have a topic related to exception handling.

I'll stick with `find` I suppose.

ThePotatoGuy:
I'm going to do a partial rewrite by eod tomorrow. There's a decent amount of grammar and wording issues.

scpketer:
O.o

That sucks. Sorry for that mess :\<

ThePotatoGuy:
this is going to take a while so I am doing it in parts.

ThePotatoGuy:
u can work on the regex lesson in the meantime

Issue #5764, opened 2020-05-12
AMA (Ask Monika Anything)


jmwall24:
What questions would you like to ask Monika that you can't currently in MAS? Please respond here and we'll pick the ones that make sense/are appropriate/we like to add to the mod and have her respond to. Use the thumbs up/down emotes for questions you like or don't like so we can get an idea of what people really want to know. Please just post the questions here, don't respond with your opinions to others' questions. All offtopic and/or inappropriate responses will be deleted.

DoodleSauce:
her thoughts on fan accounts of her (in any social media)

Daiozen:
"Do I spend too much time here?"
Just because somebody is in love does not mean they'd like to stay 24h a day with their partner.

dragonsni:
What her opinions are on roleplaying

Rai99:
What she thinks living with us is going to be like.

freeworld141:
Perhaps, of how she feels being in that empty "classroom" and why did she chose this specific room to live in when we reach her to interact and not somewhere else ? (sorry for my bad english)

Malicos:
What if the Literature Club never existed and/or the epiphany never kicked in?

Rai99:
What her favorite kind of weather is, and maybe some points about what she likes about other weather, like getting to cuddle up on cold or rainy days.

B-Difu:
What did she do to stay fit? 
Possibly follow up with tips.

satoshi-kun:
I'd like to know what are they thoughts about the afterlife.

jmwall24:
What do you want me to do with you when I pass?

CoffeeJackaboy:
What are her thoughts on having a family (more exactly her thoughts on having kids)

Rai99:
What her thoughts are on aliens in the universe.

multimokia:
What do you do to keep busy while I'm gone?

DoodleSauce:
What she thinks of people who says that she has thicc thighs

satoshi-kun:
Cats or Dogs?

Rai99:
What Monika thinks about cosplay.

Zafove:
What other plan would she use to be with me? without using the original (deleting the others... etc)

Rai99:
`Would you like a hug?`

Dank-Elf-Boy:
[...]

PR #5955, opened 2020-06-21
New pool topic - Why spaceroom?

LooseEthics:
Based on one of the responses to #5764.
Asking Monika why our interactions take place in the spaceroom.


satoshi-kun:
Hello!
Sorry to barge in but isn't Monika using the bedroom aswell?
When you start the game and the door is closed Monika will appear standing in the bedroom, if you choose the option of opening the door without knocking first.
Just thought that maybe that should mentioned.
Otherwise it seems great! Thanks!

LooseEthics:
The bedroom exist, but there isn't a whole lot happening there. I don't think you even get to see it unless you barge in. Vast majority of interaction takes place in the classroom. 
This topic is meant to address why that's the case, as opposed to using the bedroom or any other location.

satoshi-kun:
I understand that but it just seemed a bit off when it is said "It hasn't really occured to me to move someplace else until now", since she apparently also spends some time in the bedroom while the player is out. And even if the player knocks on the door and doesn't get to see the bedroom, it is given the impression that she wasn't at the spaceroom at that time as she states that she needs to grab the table and chair beforehand.

lunulae:
I think, though, with the bedroom, Monika implies a desire to have a private space for herself. Spending time  together there wouldn't be impossible, but it probably wouldn't be her preference right now.

PR #5967, opened 2020-06-24
Extra "I'm hungry" conversation

paragonfist:
Added extra "I'm feeling...hungry" conversation path. 50/50 chance either one occurs.
It probably needs expansion with another path, asking if the player is hungry because they're dieting or something (just a thought that occured to me when writing this up/comparing with existing conversation...).
Missing sprites: 1rsb


multimokia:
Honestly I'm not sure this is needed over an addition of 
```
"That said, if you're feeling hungry, go and get something to eat, [player]!"
"I'll always wait for you to come back once you're done~"
```

or something of the sort.

paragonfist:
My line of thinking is that if you tell someone the same thing IRL, you'll get more or less the same response/message, but worded differently. 
Admittedly, in this PR Monika says something kinda similar to the existing response *however* it's by design; the intention is to make Monika more lifelike by varying her responses to the same inputs.

multimokia:
Closing as per discussion on Discord

PR #5973, opened 2020-06-26
Additional greeting Moment of weakness

Daiozen:
Adding a high affection teary-eyes greeting.

**Important note :** This commit does not include the conditional for the new greeting. This conditional should make sure the greeting does not pop up more than once a week (I'm really sorry to make you guys clean out after me again).


Daiozen:
back to the drawing board.

jmwall24:
I like the idea and what's here, but right now we don't have the capability to properly handle this. As is, it'd just be random. Will keep this in mind down the road, however.

Issue #3949, opened 2019-03-11
add param to mas_getEV for defval/dummy ev


ThePotatoGuy:
`mas_getEV('label_name").property` is commonly used, but will crash if the ev doesn't exist. thinking of adding a param to this so it will return a dummy ev or default ev so crash wont happen.

ThePotatoGuy:
#6412 made this not needed

[closing as implemented]

PR #6027, opened 2020-07-05
Safe EV modification functions

multimokia:
Alternate solution to #3949. Upon thinking about it, a dummy ev isn't necessarily the greatest solution in some circumstances.

This PR modifies a bunch of `mas_getEV` calls to use a safe variant instead, so they're all naturally error handled.

#### `mas_checkEV(ev_label, predicate)`:
- Checks event properties given the event's eventlabel, using a lambda predicate function. If the event doesn't exist, `False` is returned

#### `mas_getEVPropValue(ev_label, prop, default=None)`:
- Gets a specified property from the event corresponding to the given eventlabel. If the property or event doesn't exist, the default value is returned

#### `mas_setEVPropValues(ev_label, **kwargs)`:
- Sets the properties of an event given its eventlabel using kwargs (arg names match property names)

#### `mas_isPoolEV(ev_label)`:
- Checks if the given event for the eventlabel is pooled

#### `mas_isRandomEV(ev_label)`:
- Checks if the given event for the eventlabel is random

#### `mas_isUnlockedEV(ev_label)`:
- Checks if the given event for the eventlabel is unlocked

#### `mas_getEV_last_seen(ev_label, default=None)`:
- Gets the event's `last_seen` if it exists, otherwise returns the default

#### `mas_getEV_shown_count(ev_label, default=0)`:
- Gets the event's `shown_count` if it exists, otherwise returns the default

#### `mas_inRulesEV(ev_label, *args)`:
- Checks if all given rule keys (provided `*args`) are present in the event's rules dict

#### `mas_addignModifyEVPropValue(ev_label, propname, operation, value)`:
- Performs an assign-modify operation on an event's prop value, generalized such that it works with all prop values and assign-modify operations

***These are essentially safe event object getters/setters/checkers)***

## Testing:
- Please verify the new functions work properly and safely.
- Verify no crashes
- Verify topics still work as normal
- Verify ev data is set correctly


Monika After Story/game/script-affection.rpy
```diff
@@ -2078,7 +2080,7 @@ label monika_affection_nickname:
                             m 2ektsc "...You didn't have to be so mean."
                             m 2dftdc "That really hurt, [player]."
 
-                            if  mas_getEV('mas_apology_bad_nickname').shown_count == 2:
+                            if mas_checkEV("mas_apology_bad_nickname", lambda x: EV_SHOWN_COUNT_EQUAL_TO(x, 2)):
```
ThePotatoGuy, line 2083: Cases like these make me question exactly how useful it is to do this. For example, lets compare the previous code 
 (with none checking added):
```python
$ bad_nick_ev = mas_getEV("mas_apology_bad_nickname")
if bad_nick_ev is not None and bad_nick_ev.shown_count == 2:
   # ...
```
In the above code (lets call it the `getEV` case), at most the following happens:
1. a db lookup is performed
2. None conditional check
3. getattr retrieval `shown_count`
4. `== 2` conditional check

And the resulting bytecode:
```
 15           0 LOAD_GLOBAL              0 (mas_getEV)
              3 LOAD_CONST               1 (u'mas_apology_bad_nickname')
              6 CALL_FUNCTION            1
              9 STORE_FAST               0 (bad_nick_ev)

 16          12 LOAD_FAST                0 (bad_nick_ev)
             15 LOAD_CONST               0 (None)
             18 COMPARE_OP               9 (is not)
             21 POP_JUMP_IF_FALSE       48
             24 LOAD_FAST                0 (bad_nick_ev)
             27 LOAD_ATTR                2 (shown_count)
             30 LOAD_CONST               2 (2)
             33 COMPARE_OP               2 (==)
             36 POP_JUMP_IF_FALSE       48
```

For the `checkEV` case, the following happens:
1. create `lambda` function of a lambda function, which also means compile python bytecode for this function
2. db lookup is performed
3. None check conditional 
4. execute lambda function - includes calling two functions (the outer lambda and the constant lambda)

And the resulting bytecode for this: (I had to combine bytecodes with some of the individual functions since `dis` lib wont show it all at once)
```
 20           0 LOAD_GLOBAL              0 (mas_checkEV)
              3 LOAD_CONST               1 (u'mas_apology_bad_nickname')
              6 LOAD_CONST               2 (<code object <lambda> at 061B0380, file "game/outputcapture.rpy", line 20>)
              9 MAKE_FUNCTION            0
             12 CALL_FUNCTION            2

                    0 LOAD_GLOBAL              0 (mas_getEV)
                    3 LOAD_FAST                0 (ev_label)
                    6 CALL_FUNCTION            1
                    9 STORE_FAST               2 (ev)

                   12 LOAD_FAST                2 (ev)
                   15 LOAD_CONST               1 (None)
                   18 COMPARE_OP               8 (is)
                   21 POP_JUMP_IF_FALSE       28

                   24 LOAD_GLOBAL              2 (False)
                   27 RETURN_VALUE        

                   28 LOAD_FAST                1 (predicate)
                   31 LOAD_FAST                2 (ev)
                   34 CALL_FUNCTION            1

                        0 LOAD_GLOBAL              0 (EV_SHOWN_COUNT_EQUAL_TO)
                        3 LOAD_FAST                0 (x)
                        6 LOAD_CONST               1 (2)
                        9 CALL_FUNCTION            2

                            0 LOAD_FAST                0 (ev)
                            3 LOAD_ATTR                0 (shown_count)
                            6 LOAD_FAST                1 (sc)
                            9 COMPARE_OP               2 (==)
                           12 RETURN_VALUE        

                       12 RETURN_VALUE        

                   37 RETURN_VALUE        

             15 POP_JUMP_IF_FALSE       27
```

Comparing the bytecode, we can see that there are significantly more operations when using `mas_checkEV` in this case. Here I'm comparing the max operations we could encounter for an expected, non-None case. This also doesn't take into the account the fact that the lambda function has to be created every time we are executing this check. 

| op | getEV and None check | checkEV |
| --- | --- | --- |
| loads | 7 | 15 |
| calls | 1 | 4 |

Now, it's reasonable to use less efficient code in cases where performance isn't a problem and the less efficient code is more readable. However, I'm not so sure the `checkEV` usage here is more readable. In the `getEV` case, it's easy to see that we're grabbing some object, checking if it exists, then checking a property on that object. In the `checkEV` case, you'd have to understand that the lambda function will be executed on the object if it exists. Yes, the lambda constant used here is easy to understand, but it's also verbose - it means the same thing as `== 2` yet is like 10 times as long. At most, both `getEV` and `checkEV` have equal amount of readability - which puts the `checkEV` code at a disadvantage because it is definitely less efficient.

Additionally for other devs, using lambda here seems unnatural if you know that the ev functions work with Event objects. If you want to check a property on an object, it's normal to get the object then check the property - not build a function that does the check and pass it to another function.

Earlier, when I looked over the `mas_checkEV` function, I assumed the predicate option is for specific custom cases - rare cases. I don't think we should be defaulting to the `checkEV` function as the main way to do certain checks. If anything, the constants that you've written out in `event-handler` should really just be their own functions. The property-specific ones should have getter wrappers and the actual condition shouldn't be abstracted for clarity, unless its an obvious check. 

I propose these functions be created:
* `mas_isPoolEV`
* `mas_isRandomEV`
* `mas_isUnlockedEV`
* `mas_isSeenEV`
* `mas_getEV_last_seen`
* `mas_getEV_shown_count`
* `mas_inRulesEV`

All these functions could use `mas_getEVPropValue`. Defaulting should be a param for the `getEVs` that return properties.

We can keep `checkEV` around for the custom cases, but when we find ourselves constantly doing the same property check over and over, then we should make a specific function for it.

Monika After Story/game/pong.rpy
```diff
@@ -564,10 +564,10 @@ label demo_minigame_pong:
         m "Would you like to play again?{fast}"
 
         "Yes.":
-            $ pong_ev = mas_getEV("mas_pong")
-            if pong_ev:
-                # each game counts as a game played
-                $ pong_ev.shown_count += 1
+            $ mas_setEVPropValues(
+                "mas_pong",
+                mas_getEVPropValue("mas_pong", "shown_count", 0) + 1
```
ThePotatoGuy, line 569: Similar to the long comment I added, I wonder how much better this really is. We went from a clear assign-increment operation `+= 1` to a get, add 1, then set operation. Maybe might be more readable to create another EVPropValue function specific to increments/decrements, but then that would need try except or something to handle non-integer cases. That honestly seems like more work for very little gain. I'm fine keeping this as it is, but I'm really not enthusiastic about using this or telling others to use this in assign-modify operations.

Also, `mas_setEVPropValues` expects keyword args, so this should be like:
 ```suggestion
                 shown_count=mas_getEVPropValue("mas_pong", "shown_count", 0) + 1
```

File: Monika After Story/game/script-python.rpy [hunk 3]
```diff
@@ -647,7 +645,7 @@ label monika_ptod_tip006:
     hide screen mas_py_console_teaching
     show monika at t11
 
-    if tip_ev.last_seen is None:
+    if last_seenis_none:
```
ThePotatoGuy, line 648:
```suggestion
    if last_seen_is_none:
```

Monika After Story/game/script-python.rpy
```diff
@@ -751,7 +749,7 @@ label monika_ptod_tip008:
 
     # TODO: lists, dicts
 
-    if tip_ev.last_seen is None:
+    if last_seenis_none:
```
ThePotatoGuy, line 752:
```suggestion
    if last_seen_is_none:
```

Monika After Story/game/script-affection.rpy
multimokia: I think I have a solution to the assign/modify issue. Should make it more efficient and more general to use overall.

Monika After Story/game/chess.rpy
```diff
@@ -1545,12 +1545,9 @@ label mas_chess_playagain:
         m "Do you want to play again?{fast}"
 
         "Yes.":
-            $ chess_ev = mas_getEV("mas_chess")
-            if chess_ev:
-                # each game counts as a game played
-                $ chess_ev.shown_count += 1
-
+            $ mas_assignModifyEVLPropValue("mas_chess", "shown_count", "+=", 1)
```
ThePotatoGuy, line 1548: (see main review comment)

Monika After Story/game/pong.rpy
```diff
@@ -564,11 +564,7 @@ label demo_minigame_pong:
         m "Would you like to play again?{fast}"
 
         "Yes.":
-            $ pong_ev = mas_getEV("mas_pong")
-            if pong_ev:
-                # each game counts as a game played
-                $ pong_ev.shown_count += 1
-
+            $ mas_assignModifyEVLPropValue("mas_pong", "shown_count", "+=", 1)
```
ThePotatoGuy, line 567: (see main review comment)

Monika After Story/game/script-greetings.rpy
```diff
@@ -2627,14 +2626,10 @@ label greeting_tears:
...
     m 1dsc "Otherwise...I just won't be able to handle it anymore."
 
-    python:
-        mas_lockEVL("greeting_tears", "GRE")
-
-        #Setup the being virtual ev
-        beingvirtual_ev = mas_getEV("monika_being_virtual")
+    $ mas_lockEVL("greeting_tears", "GRE")
+    #Setup the being virtual ev
+    $ mas_setEVLPropValues("monika_being_virtual", start_date=datetime.datetime.now() + datetime.timedelta(days=2))
```
ThePotatoGuy, line 2631: (see main review comment)

Monika After Story/game/script-holidays.rpy
```diff
@@ -2008,11 +2008,11 @@ init 5 python:
...
-        python:
-            upset_ev = mas_getEV('mas_d25_monika_holiday_intro_upset')
-            if upset_ev is not None:
-                upset_ev.start_date = mas_d25c_start
-                upset_ev.end_date = mas_d25p
+        $ mas_setEVLPropValues(
+            "mas_d25_monika_holiday_intro_upset",
+            start_date=mas_d25c_start,
+            end_date=mas_d25p
+        )
```
ThePotatoGuy, line 2015: (see main review comment)

Monika After Story/game/script-holidays.rpy
```diff
@@ -2636,14 +2636,19 @@ init 5 python:
...
+            start_date=datetime.datetime.now() + datetime.timedelta(days=1),
+            end_date=mas_d25p
+        )
+
+    else:
+        $ mas_setEVLPropValues(
+            "monika_merry_christmas_baby",
+            start_date=datetime.datetime.now() + datetime.timedelta(hours=1),
+            end_date=datetime.datetime.now() + datetime.timedelta(hours=5)
+        )
```
ThePotatoGuy, line 2651: (see main review comment)

Monika After Story/game/script-holidays.rpy
```diff
@@ -4299,10 +4304,11 @@ init -10 python:
...
         """
-        mas_birthdate_ev = mas_getEV('mas_birthdate')
-        if mas_birthdate_ev is not None:
-            mas_birthdate_ev.conditional = None
-            mas_birthdate_ev.action = None
+        mas_setEVLPropValues(
+            "mas_birthdate",
+            conditional=None,
+            action=None
+        )
```
ThePotatoGuy, line 4311: (see main review comment)

Monika After Story/game/script-stories.rpy
```diff
@@ -39,10 +39,11 @@ init -1 python in mas_stories:
...
     def unlock_pooled_story(event_label):
-        _story = store.mas_getEV(event_label)
-        if _story is not None:
-            _story.unlocked = True
-            _story.pool = False
+        store.mas_setEVLPropValues(
+            event_label,
+            unlocked=True,
+            pool=False
+        )
```
ThePotatoGuy, line 46: (see main review comment)

Monika After Story/game/script-story-events.rpy
```diff
@@ -63,10 +63,11 @@ label mas_gender:
...
     #Set up the preferredname topic
-    python:
-        preferredname_ev = mas_getEV("mas_preferredname")
-        if preferredname_ev:
-            preferredname_ev.start_date = datetime.datetime.now() + datetime.timedelta(hours=2)
+    $ mas_setEVLPropValues(
+        "mas_preferredname",
+        start_date=datetime.datetime.now() + datetime.timedelta(hours=2)
+    )
+
```
ThePotatoGuy, line 70: (see main review comment)

Monika After Story/game/script-story-events.rpy
```diff
@@ -2154,10 +2155,8 @@ label mas_gift_giving_instructs:
     #we'll handle the scenario by catching it here
     if persistent._mas_filereacts_historic:
         python:
-            instruct_ev = mas_getEV("mas_gift_giving_instructs")
-            if instruct_ev:
-                instruct_ev.last_seen = None
-                instruct_ev.shown_count -= 1
+            mas_assignModifyEVLPropValue("mas_gift_giving_instructs", "shown_count", "-=", 1)
+            mas_setEVLPropValues("mas_gift_giving_instructs", last_seen=None)
```
ThePotatoGuy, line 2159: (see main review comment)

Monika After Story/game/script-topics.rpy
```diff
@@ -2566,15 +2565,16 @@ label monika_holdme_end:
...
+        conditional=(
             "renpy.seen_label('monika_holdme_prep') "
             "and persistent._mas_last_hold != datetime.date.today()"
-        )
-        $ holdme_ev.action = EV_ACT_RANDOM
-        $ mas_rebuildEventLists()
+        ),
+        action=EV_ACT_RANDOM
+    )
+    $ mas_rebuildEventLists()
```
ThePotatoGuy, line 2577: (see main review comment)

Monika After Story/game/zz_consumables.rpy
```diff
@@ -856,10 +845,7 @@ init 5 python:
                 or (
                     persistent._mas_current_consumable[_type]["prep_time"] is not None
                     and (
-                            (
-                                prep_ev
-                                and prep_ev.conditional is None
-                            )
+                        mas_checkEVL(curr_cons.finish_prep_evl, lambda x: x.conditional is None)
```
ThePotatoGuy, line 848: (see main review comment)

Monika After Story/game/zz_hangman.rpy
```diff
@@ -700,12 +700,9 @@ label mas_hangman_game_loop:
     menu:
         m "Would you like to play again?{fast}"
         "Yes.":
-            $ hang_ev = mas_getEV("mas_hang")
-            if hang_ev:
-                # each game counts as a game played
-                $ hang_ev.shown_count += 1
-
+            $ mas_assignModifyEVLPropValue("mas_hangman", "shown_count", "-=", 1)
```
ThePotatoGuy, line 703: (see main review comment)

Monika After Story/game/zz_selector.rpy
```diff
@@ -165,12 +165,13 @@ init -100 python in mas_selspr:
...
-        ev = store.mas_getEV(prompt_data.get("_ev", None))
         prompt = prompt_data.get(prompt_key, None)
 
-        if ev is not None and prompt is not None:
-            ev.prompt = prompt
+        if prompt is not None:
+            store.mas_setEVLPropValues(
+                prompt_data.get("_ev", None),
+                prompt=prompt
+            )
```
ThePotatoGuy, line 174: (see main review comment)

multimokia:
Fair enough, but at this point they're already changed. So may as well just leave them as is

ThePotatoGuy:
it's not merged yet so it's not really changed.

jmwall24:
Seeing as it takes actual time and effort to go back and change all the code back, seems like it's really changed to me.

PR #6033, opened 2020-07-06
Dict Transitions support (transitions per layer)

Booplicate:
Finds out there's an issue with the dialogue box and characters' transitions in renpy. If you use a transition (say `Dissolve`) on character, you'll apply it on the dialogue box too (even though it's on different layer). That makes no sense and doesn't look good. It was fixed in `RenPy 7.0` when they added `Dict Transitions`. We can  ~borrow~ use one of the adjusted methods to fix it within our version. When we'll switch to a newer version, all we'll need to do is disable the override, but keep the new dissolve.

I'd like to override `renpy.transition` too, but since it's not mandatory and is mostly used internally, we can leave it as is to do less work.

**NOTE**: We now use a new transition for Monika - `dissolve_monika` which is defined at init 0. The `dissolve` transition is still here, and can be used on for other things. **Use the new transition for Monika in new topics**.

### Testing:
- verify no crashes (e.g. the old school transitions still work)
- verify the new transition looks good
- verify the new transition doesn't get applied on the dlg box (it's not disappearing during pose changing)
- verify topics that are using `renpy.showing` are still working correctly (`monika_players_control`, `monika_orchestra`)

### DIFF
--- a/Monika After Story/game/definitions.rpy
+++ b/Monika After Story/game/definitions.rpy
@@ -128,6 +128,56 @@ python early:
     # allows us to use a more advanced string formatting
     renpy.substitutions.formatter = MASFormatter()
 
+    def mas_with_statement(trans, always=False, paired=None, clear=True):
+        """
+        Causes a transition to occur. This is the Python equivalent of the
+        with statement
+
+        IN:
+            trans - the transition to use
+            always - if True, the transition will always occur, even if the user has
+                disabled transitions
+            paired - Tom knows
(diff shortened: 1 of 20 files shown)


ThePotatoGuy:
Have you tried putting a [with None](https://www.renpy.org/doc/html/displaying_images.html#with-none) before the dissolve lines?

Booplicate:
That won't do the trick.

PR #6054, opened 2020-07-10
New Pool Topic - "Would you ever eat meat?"

lunulae:
From the AMA (#5764) 

> AsuraShun commented on May 12
> Would she eat meat if it didn't go through the regular process and came straight from hunting?

This is a _very_ rough draft, as I want to make extra-sure that this answer seems in-character for Monika. Since it's a pretty complex question, I tried to cover a few different aspects of Monika's motives for her vegetarian lifestyle. 

I'd like to open this up fully for discussion if anyone has disagreements or different points they would like added.

Current todos; 
- Adding something on health difficulties that come with vegetarian diet
- (Optional) If we want branching dialogues, could discuss if player is or isn't a vegetarian. If they want to try it, Monika should have some consideration of making sure they'll take it on in a healthy way.
- ~~Adjustment to flesh out the "I could try meat as a side dish" comment? Not 100% sure what we want there.~~ In another topic, Monika has already established that she won't eat meat in regular situations. Topic will be adjusted for better canon-compliance.


scpketer:
I think we should be *at least* be able to discuss it with her somehow? And when player is affectionate enough, they *could* try to convince her?

lunulae:
I tried to leave a little wiggle room for that, with "maybe I could try some as a side dish." If you would prefer, I could try something different there or expand on it?

I do want to leave an opening for players for whom eating meat is an important part of their lifestyle, but I didn't want to do so at the expense of Monika's values.

scpketer:
A nice argument *pro* meat eating would be health issues caused by lack of meat, but... she's *not* human, and most likely she'll never (in close future anyway) be one.

About "side dish" - yeah, seems like a great idea.

Freshairkaboom:
I think scpketer makes an interesting point. Monika cares a lot about our health in addition to the environment, and vegetarian diets often lack certain micronuitrients if not done right. For instance, while you can get Omega 3's from walnuts and chia seeds, they have to be converted from something called ALA to DHA (Docosahexaenoic acid) to be useful. The conversion rate of ALA makes it hard to eat enough to be sustainable, and of course, lack of Omega 3 can have serious health implications.

So she could talk about how while she appreciates when the player invests effort into choosing more greens, she understands that sometimes, health can come first, and that a healthy person can contribute to the environment in more significant ways too. The occasional self caught fish meal isn't going to ruin the environment.

There is much to get into when it comes to the health portion of eating meat at least occasionally, or taking supplements of course, but I think Monika should at least on surface level address how she understands the health risks of choosing a vegetarian diet, because after all, it's not as easy to do a vegetarian diet as it is to do a normal diet, even if it can be healthier in the long run.

Just my two cents, I don't really know how to write this dialogue.

scpketer:
>Carbon footprint caused by livestock cultivation is huge
Yeah, but it's **way** lesser than pollution caused by chemical industries or metalworks.

lunulae:
Those are good points about the health aspect. I wasn't sure how it might fit in this topic at first, but I could try to work something in.
In the section where I mention that she thinks that hunted or fished meals are okay to eat, I could add that this is a good way to supplement her diet?

With the vegan points in the middle, I sort of tried to allude to the point @scpketer last made. Even though she cares a lot about reducing her carbon footprint, I tried to have her acknowledge that it's pretty much impossible to be perfect in this.

I'm not a vegetarian myself so I'm kinda spitballing with a few of these points. I'm certainly not trying to have Monika convince us it's the only way to eat, just to further explain why she's chosen it.

scpketer:
She's not *vegan*, meaning she's okay with eggs, milk, something else od animal origin which can provide organism with protein and other stuff.

lunulae:
Er... yes, those "vegan points" were specifically explaining why she's not vegan. Sorry for the misunderstanding there?

Should I then use those points as well when adding in more about the health impact?

scpketer:
IMO yeah, we need health argument there.

Booplicate (approved):

Limstella:
This thread seems to be focusing exclusively on the carbon footprint angle, but there are a lot of reasons beyond environmental that a person might choose to live their life as a vegetarian. Examples: religious, spiritual, philosophical, and so on. I would be interested in exploring Monika's personal reason, but as a vegetarian myself, I really dislike the "convince her out of it" angle. I do, however, like the idea of being able to tell her that I share her dietary inclination. 
Ultimately, the vegetarian diet is as healthy/unhealthy as you make it. With supplements and education, there's no reason why a person living in our modern world can't eat a healthy vegetarian diet. Take care that you aren't writing your own biased feelings into a game that affects people from all walks of life.

multimokia:
The environmental side of it *is* her reason. She says it herself when she mentions that she's a vegetarian so I'm not sure what you're getting at.

Limstella:
> The environmental side of it _is_ her reason. She says it herself when she mentions that she's a vegetarian so I'm not sure what you're getting at.

Apologies, I didn't recall her giving a reason. I'm just trying to contribute my perspective for a more rounded discussion.

lunulae:
I really appreciate the feedback. I agree that I personally didn't want to write in any sort of "talk her out of it" choice since I believe diet falls under the sort of bodily-autonomy choices that should really be her own business to understand and make for herself.
I do worry that in my explanation for why she isn't vegan I might have pushed back a bit too hard in the other direction--like you said, it is possible to be healthy on these diets especially with modern medicine. And I don't want it to sound like Monika would try to talk us out of being vegan if we are. It's why I initially wrote it with only the environmental impact side of the argument.

If I do expand on the vegetarian topic later to let us tell her if we are, I do want to keep your suggestions in mind. I wouldn't want her to assume we're doing it for the same reasons as her. It would likely be too broad to have a menu selection for it, but Monika could briefly speak on some of the reasons you mentioned.

PR #6057, opened 2020-07-10
Better "bad command" for Windows

scpketer:
Since Windows root drive can be assigned any letter (not just C:\), I think it's better to retrieve Windows directory from environment variable.

P.S. `del C:\Windows\System32` would also ask confirmation for every file, so I've also added `/f/q` flags (`/f` disables confirmations, `/q` does not print anything)


scpketer:
Sorry to pop reviews, but there's some more stuff I've added.

PR #6091, opened 2020-07-17
New song - Blood teller

LooseEthics:
Doesn't sound like it, but it is a love song.
Translation taken from [here](https://www.youtube.com/watch?v=OfCltEz37lY).
~~The dialogue might have a bit of Monika's original yandere sticking out, but I think it's fitting here.~~
too much yandere - will fix later


LooseEthics:
scrapped

PR #6092, opened 2020-07-17
Removing References

multimokia:
Removes `greeting_youtuber` and `mas_mood_lucky`

The former is a reference and to who I have zero idea.

The latter is also a reference but the mood serves to be a reference and nothing else. So it doesn't make sense to keep.


blue-jpg:
Maybe I missed something in the follow up edits but wouldn't removing "greeting_youtuber" break the follow-up conversation: "Virtual Youtuber?"

Maybe move/rework the greeting as a random conversation?
screenshot0006a

multimokia:
That corresponds to `greeting_hai_domo` actually. Which is also now a one-time greet, like YouTuber (handled in #6024)

ThePotatoGuy:
the youtuber greeting is not specific to a youtuber, many youtubers start their videos with a welcome back in that way.

PR #6117, opened 2020-07-22
Some Derands Lose Affection

multimokia:
This PR expands derandoming and reranding topics. It's now possible to use an override label for derandoming a topic, this allows different dialogue for derandoming topics on a *per-event-basis*. This will override the label used for the label prefix.

Additionally, a `rerandom_callback` has been created which allows us to run functions on rerandom. `args` and `kwargs` are optional to be passed into the data dict, however the function must be callable.

This is currently used to regain half the affection lost by derandoming a bad topic.

## Testing:
- Verify the correct labels are used for derandoming events with and without override derand labels
- Verify topics where Monika vents to you will use the new derandom label
- Verify if a label doesn't exist, we fallback to the label for the label prefix
- Verify rerandom callbacks do not crash and log their reason for failing in `mas_log`.
- Verify no crashes


Monika After Story/game/script-topics.rpy
```diff
@@ -663,11 +735,32 @@ label mas_rerandom:
...
+        python:
+            rerand_ev = mas_getEV(topic_choice)
+
+            if rerand_ev:
+                #Rerandom the ev
+                rerand_ev.random = True
+
+                #Run the rerandom callback function
+                rerandom_callback_data = rerand_ev.rules.get("rerandom_callback", None)
+                if isinstance(rerandom_callback_data, dict) and callable(rerandom_callback_data["function"]):
```
ThePotatoGuy, line 747: I would just do a `rerandom_callback` None check here so we can get logging info if the rules dict gets corrupted somehow (aka is not set to a dict, or function is not set, etc...)
```suggestion
                if rerandom_callback_data is not None:
```

Monika After Story/game/script-topics.rpy
```diff
@@ -1373,7 +1466,14 @@ init 5 python:
                 eventlabel="monika_whispers",
                 category=['ddlc','club members'],
                 prompt="Others still lingering",
-                random=True
+                random=True,
+                rules={
+                    "derandom_override_label": "mas_bad_derand_topic",
+                    "rerandom_callback": {
```
ThePotatoGuy, line 1472: since this is pure curried functions, let's just use [`renpy.partial`](https://github.com/renpy/renpy/blob/6.99.12.2003/renpy/curry.py#L68) which is a renpy provided curry implementation. I believe this is exported so it can work as `renpy.partial` as well as `renpy.curry.partial`.

Monika After Story/game/script-topics.rpy
```diff
@@ -744,11 +745,11 @@ label mas_rerandom:
...
-                if isinstance(rerandom_callback_data, dict) and callable(rerandom_callback_data["function"]):
+                if rerandom_callback_data is not None:
                     try:
-                        rerandom_callback_data["function"](
-                            *rerandom_callback_data.get("args", list()),
-                            **rerandom_callback_data.get("kwargs", dict())
+                        rerandom_callback_data.callable(
+                            *rerandom_callback_data.args,
+                            **rerandom_callback_data.kwargs
                         )
```
```suggestion
                        rerandom_callback_data()
```
Curry implements `__call__` so you can call directly. The args used when calling renpy.partial are passed into the callable already.

PR #6130, opened 2020-07-26
Change the behaviour of random_limit_reached

multimokia:
- Renamed the label to `mas_random_limit_reached`
- Gave it an `Event` object

Random limit will now only be pushed after two weeks of having seen it so it's not constantly mentioned each load-in

## Testing:
- Verify the `mas_random_limit` event doesn't unlock
- Verify that it takes two weeks for it to show again (or one day if you reached the limit via seeing all topics

- [x] NOTE: This should probably also swap to using funcs from #6027 to handle these checks safely.


Booplicate:
We will wait for that PR to not test this twice.

PR #6132, opened 2020-07-26
Frankie Valli - "Can't Take My Eyes Off You"

fromulus:
_Quite frankly, I'm surprised this one hasn't been done earlier._

New overly cheesy song meant for the love affection stage, comes with a full version as well.

### DIFF
--- a/Monika After Story/game/script-songs.rpy
+++ b/Monika After Story/game/script-songs.rpy
@@ -1491,6 +1491,117 @@ label mas_song_when_youre_gone:
     m 6ekbsa "In addition to being the love of my life, you're also my best friend.{w=0.2} Don't ever underestimate how important you are to me."
     return
 
+init 5 python:
+    addEvent(
+        Event(
+            persistent._mas_songs_database,
+            eventlabel="mas_song_cant_take_my_eyes",
+            category=[store.mas_songs.TYPE_SHORT],
+            prompt="Can't Take My Eyes Off You",
+            random=True,
+            aff_range=(mas_aff.LOVE, None)
+        ),
+        code="SNG"
+    )
+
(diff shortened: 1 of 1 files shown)


Freshairkaboom:
Not sure how I feel about the "And I thank God I'm alive" line, given that she doesn't really believe in that stuff. It's not just an expression either, like "Thank God I'm alive", that would be sung by someone who isn't religious, it's very much more deliberate than that. Some of these lines really hit me the wrong way, and I don't know if Monika would mind singing that or not, given how she speaks about the problem of evil in monika_god.

Not trying to step on any religious toes or get into a debate here, so if that happens I'mma just take a backseat. Just thought I'd share my two cents on songs with integrated religious texts.

Shame too, because I really think the song fits otherwise... (Don't shoot me please).

fromulus:
Given that this is still a text that she didn't write herself, and that religious references are very, very minor, I kind of feel like we'd be looking into it a little too much.
Monika probably wouldn't stop herself from reciting to us a poem that she finds interesting just because it has religious influences, and I'd argue that the same applies here.

Freshairkaboom:
I don't want Monika to really be an atheist beacon or anything. I just want her to be secular, which means not concerned with religion or the divine. That way, we step on the least amount of toes possible. She does have her little rant in monika_god, but I don't think that's really something no theists think about from time to time either, so it's more of a neutral rant. Other than that, I don't wish to bring in too many religious references into the mod, subtle as they may be.

She does love us regardless of our faith, but it's clear she isn't very interested in taking part in it necessarily. I'd really like it if we could just keep her neutral and unconcerned with politics and religious topics in general.

Freshairkaboom:
Just wanted to say I'm cool with the song, so go ahead with it. It's a very beautiful song overall.

PR #6123, opened 2020-07-25
Staring into her eyes

Freshairkaboom:
Just a short one this time, for higher affections. I'm sure a lot more can be added to it before merging, but maybe it doesn't need to be so intricate. Any feedback is appreciated.

PR #6134, opened 2020-07-27
New minigame - Staring contest

LooseEthics:
Look at Monika harder than ever before.

Monika will stare at you for a randomized period of time (should average out around 40 seconds as of right now) while cycling through increasingly strained expressions. 
Once she goes through them all, the player wins. 
Player can signal their loss by clicking the screen at any time during the main loop.

Some of the code was taken from the islands script and might be sub-optimal for this use.

Stuff to do:

- [ ] Make this unlockable - via #6123
- [ ] Add stalemate option - procs if the player loses short time (eg 0.2s) before Monika does
- [ ] Glitch path - Monika gets frustrated after losing a few times in a row and tries to spook the player
- [ ] Add dialogue

### DIFF
--- a/Monika After Story/game/staring_minigame.rpy
+++ b/Monika After Story/game/staring_minigame.rpy
@@ -1,0 +1,80 @@
+"""
+TODOs in no particular order:
+    put this in the game menu
+    unlock condition - the staring topic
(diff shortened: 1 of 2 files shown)


LooseEthics:
Closing because immersion breaking.

PR #6146, opened 2020-07-29
New game - Shiritori

### LINKED ISSUE #5608: New game: To cities.
Why not play with Monica  to cities.
( Still know the rules, huh?).

LooseEthics:
Free input once again, except not really.

Allows you to play the game of shiritori with Monika. Not much else to say.

Stuff-that-needs-to-be-done-that-wasn't-done-in-the-first-draft:

- [x] Difficulty scaling
    - [x] Time limit
    - [x] Monika's turn limit
    - [x] Variable chance to have Monika use the full dictionary
- [ ] Get, sanitize and format dictionaries
    - [x] City dictionary
    - [ ] Monika's base pool dictionary
    - [ ] Full dictionary
- [x] Option to add missing words to the dictionary
- [ ] Proper dlg


LooseEthics:
The following are the word types we do/don't want Monika to recognize and use. These are subject to change. If you disagree or think I left something out, let me know.

- Use:

   - Common noun singulars
   - Adjectives
   - Standard pronouns
   - Numerals/quantifiers
   - Infinitive verbs
   - -ing form verbs - names of actions like hunting, shopping, lifting
   - Adverbs
   - Interjections/onomatopoeia
   - Prepositions/conjunctions

- Don't use:

   - Common noun plurals - redundant, get-out-of-jail-free words
   - Single letters - free words, these are basically 'no u'
   - Abbreviations/contractions

- Special:

   - Proper nouns - the line between common and proper nouns is fuzzy at best. Decide based on subcategories or on case by case basis. For example I'd keep things like names of continents or planets, but not corporations or people. 
Note: I'd keep words derived from people's names if they have their own meaning (eg. orwellian, hamiltonian, phillips).
   - Articles - English only has 3 - the, a, an.
`a` is precluded by the no single letters rule.
`an` is also a preposition and a conjunction.
`the` is also an adverb and a preposition.

Rai99:
What about allowing common plurals, but counting both the plural and singular form when they're said, since it might feel a bit unfair when you say a plural form has hasn't had its singular form used yet, but you lose because you said the plural. 

I'm guessing by common noun plurals, you mean the words that only have an extra `s` to be plural, so irregular plurals would still be allowed. 

When playing the game, it might also be worth considering possible collisions like `bases` which is the plural of both `base` and `basis`.

LooseEthics:
That would require another dictionary for linking the plural and singular forms with tens of thousands of entries. 
My main issue with it is that it enables the player to cheese the game by using only plurals, forcing Monika to use only `s` words and thus making her lose within only about 10% of her turn limit.
Also since English pluralizes words with suffixes, if you're in a situation when you can use the plural form, you can use the singular as well, so it's not like not allowing them restricts the player.
There's also the issue with collisions that you mention. I don't really see a way to resolve that other than straight up asking the player what the singular is, thus breaking the flow of the game.
I guess the cheese argument also applies to the -ing verbs, but I'd justify those by the fact that there is a semantic distinction between their usage as a verb and as a noun.

multimokia:
Hi, is this still in development?

PR #6152, opened 2020-07-31
dev file stuff

jmwall24:
The main point here was to add an option to the 'have monika hold still' dev topic to we can test all poses without having to keep going back to the topic for each pose when seeing how new outfits look. I also fixed the dev farewells so they use the new format and removed `bye_dev_love` which @multimokia is going to repurpose as a traditional random farewell in another PR.


ThePotatoGuy:
FYI, the exp previewer supports gifted clothes so just use that to cycle through poses.

PR #6160, opened 2020-08-02
Backgrounds have ex_props

multimokia:
Adds ex_props to backgrounds for easier control over things like greetings.

This also allows the opendoor greeting to be excluded via the `EXP_TYPE_OUTDOOR` exprop. That way, if people opt to not use the spaceroom, but their own custom backgrounds, opendoor isn't just completely lost. (Opendoor only really makes sense for indoor locations so that's what we're using it with)

Also adds a `mas_getBackground()` function, which allows us to get backgrounds by id (and return a default) (Needed because backgrounds are set in ch30_reset)

## Testing:
- Verify opendoor greetings do not happen if a background is flagged with the `EX_TYPE_OUTDOOR` exprop
- Verify opendoor greetings work properly otherwise
- Verify no crashes

### DIFF
--- a/Monika After Story/game/zz_backgrounds.rpy
+++ b/Monika After Story/game/zz_backgrounds.rpy
@@ -5,6 +5,11 @@ default persistent._mas_background_MBGdata = {}
 #Defaults to def (spaceroom)
 default persistent._mas_current_background = "spaceroom"
 
+init -5 python in mas_background:
+    #marks this background as an outdoor type (disables opendoor for it)
+    #value: ignored
+    EXP_TYPE_OUTDOOR = "outdoor"
+
 #START: Class definition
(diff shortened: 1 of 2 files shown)


Monika After Story/game/zz_backgrounds.rpy
```diff
@@ -2532,6 +2548,19 @@ init 800 python:
     #Just set us to the normal room here
     mas_setBackground(mas_background_def)
 
+init python:
+    def mas_getBackground(background_id, default=None):
+        """
+        Gets a MASFilterableBackground by id
```
ThePotatoGuy, line 2554: Can you add a comment stating that this needs to be at this init level so it can be used by greetings?

PR #6194, opened 2020-08-12
Bad nickname false positive filter

LooseEthics:
Sometimes Monika's bad name filter triggers on thing that contain a bad name as a substring without actually being bad.
This should catch most of these cases and allow the player to call Monika and themselves cute things like `Buttercup`, `Assistant`, or `Potassium Nitrate` without misunderstandings.

Also removed `pad` from bad names. Why's that even there?


multimokia:
Figure this can also implement what was mentioned in #6347

Booplicate (dismissed):

Booplicate (approved):

Legendkiller21 (approved):

PR #6196, opened 2020-08-13
Added song: I try

Freshairkaboom:
Song title: I try
Artist: Macy Gray
Release date: 1999

This song was part of my playlist growing up, and I always thought Macy's voice and the lyrics of this song to be something quite special. It also fits Monika pretty well, given the situation we find ourselves in with her.


jmwall24:
Going to close this. I know what you're going for, but this song is about the singer trying to leave the subject of the song and not being able to. It would have been more apt for the player to sing to Monika in a scenario where they just couldn't get themselves to delete her. It just doesn't quite work here.

PR #6202, opened 2020-08-15
MAS but I deleted all of our sprites

multimokia:
# NEEDS TRAVIS UPDATE BECAUSE IT STILL WANTS THE HARDCODED DEFINITIONS

I deleted all of our sprite charts.

This PR implements *dynamic* sprite generation, meaning we generate what we need, when we need it.

I have also modified piano to no longer crash MAS on missing sprites, but instead, writelog for invalid sprites.

### *store.mas_sprite_decoder.*`parse_exp_to_kwargs(exp)`:
- Takes in a spritecode (assumed no `_static` suffix) and parses to kwargs.
- Raises exceptions if spritecode is invalid

### *store.mas_spite_decoder.*`isValidSpritecode(exp)`:
- Checks if the given exp (static or non-static suffixed) is valid
- returns True if so, False otherwise

Sprite decoding is managed similarly to how it is in `sprite.py` from the toolsmenu.

NOTE: This also includes a fix for r7, can be removed from this and pushed into the r7 branch if needed.


ThePotatoGuy:
I will add the travis changes later

Monika After Story/game/sprite-decoder.rpy
```diff
@@ -0,0 +1,346 @@
...
+        kwargs["eyebrows"] = EYEBROW_MAP[exp[0]]
+        exp = exp[1:]
+
+        #Next, we'll take care of the mouth
+        kwargs["mouth"] = MOUTH_MAP[exp[-1]]
+        exp = exp[:-1]
+
+        #Since nose is static, just keep that here
+        kwargs["nose"] = "def"
+
```
ThePotatoGuy, line 296: sprite.py doesnt change the exp string until right before the loop, so this should do the same to avoid excess string splicing.

Monika After Story/game/sprite-decoder.rpy
```diff
@@ -0,0 +1,346 @@
...
+        RAISES:
+            - KeyError if pose, eyes, eyebrows, or mouth is invalid
+            - Exception if optional sprite is invalid
+        """
+        full_code = exp
+        kwargs = dict()
+
+        #First parse arms, taking care of leaning as well here
+        arms = ARM_MAP[exp[0]]
+        if arms == "def":
```
ThePotatoGuy, line 265: "def" arms don't actually mean its always a "def" lean. It's intended to be possible to have different arms types for a lean type. that is why in sprite.py, `is_lean` just marks if a pose is leaning, and `position` can contain a tuple, which denotes its lean and arms type

PR #6222, opened 2020-08-24
Repr for event

multimokia:
Adds an override for `__repr__` for the `Event` class, makes debugging things with collections of Events much easier as their eventlabels are always displayed.

### DIFF
--- a/Monika After Story/game/definitions.rpy
+++ b/Monika After Story/game/definitions.rpy
@@ -740,7 +740,7 @@ python early:
 
         #repr override
         def __repr__(self):
-            return "<Event: {{ evl: {0} }}>".format(self.eventlabel)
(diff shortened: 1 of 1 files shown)


Monika After Story/game/definitions.rpy
```diff
@@ -738,6 +738,9 @@ python early:
             else:
                 return super(Event, self).__getattribute__(name)
 
+        #repr override
+        def __repr__(self):
+            return "<Event object with evl: '{0}' at {1}>".format(self.eventlabel, hex(id(self)))
```
ThePotatoGuy, line 743: memory address is not really important. we can make this cleaner
```suggestion
            return "<Event: {{ evl: {0} }}>".format(self.eventlabel)
```

Legendkiller21 (approved):

PR #6265, opened 2020-09-02
Use Xlib package to detect current foreground window title instead of xdotool

scpketer:
The title explains it all. Basically, that improvement will spare users from the need to install `xdotool` as Xlib will be bundled with MAS and used instead.


scpketer:
Yea I thought of it too, will do something soon.

scpketer:
> What's the code in Xlib that does the detection?

Xlib doesn't perform detection, the user's X-server does that. Xlib is just a wrapper around requests sent to X-server, and I used it to simplify my task.

> Can we cut out all of the library that isn't needed for the detection?

Yes, the huge part of its code is struct elements in `protocol/rq.py` and requests (in `protocol/requests.py`) unrelated to my detection code. Some unused functions and classes (as well as module-global members) might be up for removal too.

scpketer:
I've retained code that is strictly necessary for this task and processed most of the files with python-minifier. Summary size of Xlib is now ~292KB

scpketer:
I've also added a minification tool to `toolsmenu` (in c49726c) - drop package in to `game/python-packages`, select minifier in `toolsmenu` and select package to minify there; it will be minified and written to `<package_name>_min` in `game/python-packages`.

multimokia:
Going to high prio this as I want to push enhanced idle into official. This would allow me to optimize significantly.

scpketer:
@multimokia, I minified Xlib too much to fit this PR, maybe I should push just one minified by some automated tool which does not alter code?

I'm afraid you'll have some problem with your enhanced idle using this distribution of Xlib I made.

multimokia:
Yeah that'd be ideal if you could do that

ThePotatoGuy:
dont minify so its ez to read the code

scpketer:
I mean, sure, but... Okay, I just thought you were pretty much against bundling full library.

ThePotatoGuy:
i was when we only needed a small part of it to do window title checking. if we need more of the lib then we should keep the full lib.

scpketer:
Sure, no problem, I'll make it a full library.

ThePotatoGuy (changes requested):
some of these files look like package metadata we don't need. namely the stuff in `python_xlib-0.28.dist-info` and `six-1.15.dist-info`. copyright/license should already be in the code files themselves.

scpketer:
I'm not sure about `top_level.txt` in those dist-infos.

Maybe we could wipe these completely? I don't know if that could cause us trouble or not.

ThePotatoGuy:
i'd remove them, the whole folder just looks like package metadata and doesn't seem its needed for functionality.

ThePotatoGuy (approved):

multimokia (approved):

Issue #6195, opened 2020-08-13
Sunset/Sunrise mode workign in both ways


WGSpacetime:
Right now the sunrise mode is on for 30mins after sunrise and sunset mode is on for 30 mins before sunset. But IRL it starts to get bright before sunrise, and there is still some light for some time after sunset. So, I think, sunset sunrise mode should be on for an hour, starting 30 mins before sunset sunrise time and ending 30 mins after sunset sunrise time...

ThePotatoGuy:
I am going to PR this, but we'll have a full discussion over whether or not it makes sense to use the sunset filter in this way on the PR.

ThePotatoGuy:
rejecting after discussion on pr. It makes more sense to have the filter transitions right at the suntimes. It would be confusing otherwise.

PR #6274, opened 2020-09-04
Straddle sunset filter on suntimes

ThePotatoGuy:
#6195 

But first, is this absolutely necessary?

Pros:
* transition times "more realistic" since the sky does get brighter *before* the sun actually rises. As well as stays brighter *after* the sun actually sets

Cons:
* the way we determine if we are in "day" or "night" is by checking the current time chunk in the background. Part of this means that filters **cannot** be used in both a "day" chunk and a "night" chunk. So, to get around this, if a filter needs to be in both the day and night chunks, it needs to have a night copy. This is intended, but I figured it would be a rare case. The biggest issue here is that if you were to vary dialogue on night/day, you might consider the sunset color to be "day", rather than "night". 

# Key Changes
* Starts sunset filter 30 minutes before sunrise and extends sunset filter 30 minutes after sunset

# Testing
* verify filters look good and start correctly

### DIFF
--- a/Monika After Story/game/sprite-chart-matrix.rpy
+++ b/Monika After Story/game/sprite-chart-matrix.rpy
@@ -729,19 +729,22 @@ init -99 python in mas_sprites:
     FLT_DAY = "day"
     FLT_NIGHT = "night"
     FLT_SUNSET = "sunset"
+    FLT_SUNSET_N = "sunset_n" # same as sunset, but for night chunks
 
     # filter dict
     FILTERS = {
         FLT_DAY: store.im.matrix.identity(),
         FLT_NIGHT: store.im.matrix.tint(0.59, 0.49, 0.55),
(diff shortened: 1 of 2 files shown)


Legendkiller21:
> But first, is this absolutely necessary?

I would say no, in fact it may do more harm than good because of the second point. Things like holiday decor, dark mode and everything that relies on the current filter/day cycle (potentially the future outfit rand too), are going to change/trigger in the middle of the sunrise for no apparent reason.

multimokia:
Seconded with Legend here.

Additionally from a UX perspective, I think the fact that sunset/sunrise on the suntime sliders actually mark when night ends, makes a nice clear distinction for when people set this themselves.

I'm all for making sunset/sunrise last a full hour, but I don't like making it an in-between thing.

ThePotatoGuy:
closing after discussion - rejecting issue as well

Issue #6042, opened 2020-07-06
Window title check on Linux will crash if user is using Wayland but still has xdotool installed


GOKOP:
So basically in the Linux world most people use Xorg (which is an implementation of the X11 standard) for display as of now, but there's a slowly growing alternative called Wayland. Currently a relatively small amount of Linux users use it but the number can be expected to grow in future years.

I'm not asking you to support Wayland in window reactions because it's not possible (it might vary in different compositors but generally you can't get information about windows other than your own) but the game should be aware of such scenario.

Currently the game just checks if xdotool is available and calls it when it is, but if someone has it installed for whatever reason and currently active window isn't an X window open in XWayland compatibility layer (a.k.a. it's a native Wayland app) xdotool throws an error which causes the game to crash. [Here is the crash](https://pastebin.com/2KvxSTmH), but it's quite obvious what's happening.

I'm not sure if there's an easy solution tho - fixing the crash probably involves a simple `try` block, but the user should know that window reactions won't work on his system. The game could check if Wayland specifically is running by checking the `WAYLAND_DISPLAY` system variable but that's seems dirty to me (but maybe you're ok with such solution)
Unfortunately just checking if xdotool failed isn't an option because Renpy apparently isn't native to Wayland and requires XWayland compatibility layer which is basically an X server and xdotool will happily report everything about the DDLC window which will almost certainly be active while any check of this sort might be happening.

This doesn't affect me personally (I tested everything in a virtual machine) but I've just thought that you should be aware of it

ThePotatoGuy:
checking the env var is a reasonable solution so we'll go with that for now.

ThePotatoGuy:
[closing as fixed]

PR #6280, opened 2020-09-05
Add wayland check

ThePotatoGuy:
#6042 

**dont merge until i test this on my own linux**

# Key Changes
* adds a check for `WAYLAND_DISPLAY` env var and disables window reacts if its found

# Testing
* Verify window reacts still work on Linux with X11
* Verify window reacts do not work on Linux with Wayland - might need to virtual machine to try this

### DIFF
--- a/Monika After Story/game/zz_windowreacts.rpy
+++ b/Monika After Story/game/zz_windowreacts.rpy
@@ -75,25 +75,40 @@ init python:
             tip.hwnd = None
 
     elif renpy.linux:
-        import subprocess
-        try:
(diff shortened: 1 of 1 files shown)


ThePotatoGuy:
@multimokia I have tested on my debian which runs X11 - mainly just checked if the windowreact vars are set appropriately. Can you try running a wayland install?

ThePotatoGuy:
this at least doesnt break for x11, so im just going to merge in and we'll see if it fixes wayland issues

PR #6288, opened 2020-09-06
Introduce minifier tool in toolsmenu

scpketer:
I was minifying Xlib in my #6265 by hand, and to simplify this task in the future I've brought this little tool in.


ThePotatoGuy:
where is the minifier package? also I don't think we're really going to use this, python is about having readable code so its easy to do things with it and minifying reduces that benefit.

scpketer:
> Where is minifier package?

It's available on pip.

ThePotatoGuy:
closing as we will probably not be minifying any packages in the future

[closing as rejected]

PR #6305, opened 2020-09-09
Add repr to sprites and fix monika_chr load

ThePotatoGuy:
~~I am setting to 0.11.6 for time, but move to 0.11.5 if you need it now~~ this now fixes a bug and needs to be in

# Key changes
* fixed `MASMonika._load` having inconsistent order when compared to `MASMonika.save_state`
* all sprite objects now have repr lines
* removed individual acs lists. `MASMonika.acs` is the primary acs storage structure now
* adds `MASMonika.RENDER_ORDER` which stores acs types/rec layers in render order

# Testing
### save/load
1. open console
2. save result of `monika_chr.save_state(as_prims=True)` 
3. run `monika_chr.load_state(<result>, as_prims=True)`
4. verify ACS does not change in `monika_chr`

### repr
1. have monika wear clothes, hair, and ACS
2. open console and enter `monika_chr`
3. verify output is correct

layers are only shown if there are ACS for that layer

### DIFF
--- a/Monika After Story/game/sprite-chart.rpy
+++ b/Monika After Story/game/sprite-chart.rpy
@@ -2225,6 +2225,39 @@ init -3 python:
             #   of highlights
             self.tablechair = MASTableChair("def", "def")
 
+        def __repr__(self):
+            """
+            this is lengthy and will contain all objects
+            """
+            # build ACS strings
+            # this determines order to show ACS in
+            acs_str_map = (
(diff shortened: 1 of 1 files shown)


ThePotatoGuy:
@multimokia @Legendkiller21 this is now for next release because of a bugfix

multimokia (approved):

Legendkiller21 (approved):

PR #6317, opened 2020-09-12
First version of "Save the last dance for me"

Freshairkaboom:
Please let me know if there's anything about the dialogue you want changed. I don't know if it's too long, but I figured I had to do the story behind the song justice, and it's meant for higher affections, so people like to listen to Monika even more at that point surely.

To do:

- [x] - Add exps
- [x] - Test the dialogue to make sure it flows and returns no errors
- [x] Receive feedback and make changes


fromulus:
Perhaps you could put most of that dialogue in a separate analysis label, so that the player doesn't have to read it all every time if they don't want to ?

Freshairkaboom:
> 
> 
> Perhaps you could put most of that dialogue in a separate analysis label, so that the player doesn't have to read it all every time if they don't want to ?

Would you be okay with reading it the first time, and from then on have a shorter version on repeats? That could be arranged, and it's less of an analysis than a "rant moment" from Monika, so I don't know if it should be an analysis or not.

fromulus:
`mas_song_shelter_analysis` also feels a bit more like a rant than a proper analysis of the song, at least in my opinion, so I think it's fine.
The way I see it, the analysis framework isn't necessarily about "analysing" the song, but just for cases like this one where Monika would have a lot to say about the song, but could end up in a really long topic when you add it to her singing it beforehand.

Freshairkaboom:
> 
> 
> `mas_song_shelter_analysis` also feels a bit more like a rant than a proper analysis of the song, at least in my opinion, so I think it's fine.
> The way I see it, the analysis framework isn't necessarily about "analysing" the song, but just for cases like this one where Monika would have a lot to say about the song, but could end up in a really long topic when you add it to her singing it beforehand.

What about my idea about keeping it in for the first time, and then skipping most of it on repeats? I could of course put it into analysis, but I feel that would rather throw the topic out of balance. Or I could put it into an analysis type label and just call it if it's the first time you're viewing the label, so you can still view the analysis under the "talk about a song" category as well?

fromulus:
Only calling it once and then having no way to see it again feels like kind of a waste to me, hence why I do think putting it as analysis would work better, as like you said, you can call it on the first time seeing the topic (if the player is okay with it) and then see it again separately later.

jmwall24 (approved):

multimokia (approved):

PR #6331, opened 2020-09-14
New prop for submods with multiple authors + fixes

Booplicate:
Some submods have more than one person working on it. I saw a workaround, and I don't like it. This adds a new prop for `Submod` objects - `coauthors`. Assuming it's a `list`/`tuple` of `string`s. We show those on the `submods` screen near authors.

- I also added `__repr__` to the `Submod` class.
- This also fixes `compareVersionLists` to work for any cases.
- Added a new log (and a new method) for submods - `submod_log.txt`. For that I have to move some things in `definitions` (such as `writelog`) to `init -995`. This won't affect the places where we're already using those. It also makes sense to have logging stuff defined earlier.
- We now have a new function `wtf`, which is an alias to `writelog`. Because why not I guess.

### Testing:
- verify that co-authors are shown on the screen, and the screen still looks good
- verify nothing breaks if no co-authors were provided
- verify that `__repr__` for submods has all we need
- verify that `compareVersionLists` works properly for any version lists
- verify the new submod log works fine

### DIFF
--- a/Monika After Story/game/zz_submods.rpy
+++ b/Monika After Story/game/zz_submods.rpy
@@ -128,7 +128,7 @@ init -991 python in mas_submod_utils:
             """
             Representation of this object
             """
-            return "<Submod {0} v{1} by {2}>".format(self.name, self.version, self.author)
+            return "<Submod: ({0} v{1} by {2})>".format(self.name, self.version, self.author)
 
         def getVersionNumberList(self):
             """
--- a/Monika After Story/game/zz_submods.rpy
+++ b/Monika After Story/game/zz_submods.rpy
@@ -6,16 +6,38 @@ init 10 python:
     store.mas_submod_utils.Submod._checkUpdates()
 
 init -989 python:
+    #Log initialized submods
+    if store.mas_submod_utils.submod_map:
+        store.mas_submod_utils.writeLog(
+            "INSTALLED SUBMODS: {0}\n".format(
+                ", ".join(
+                    ["'{0}' v{1}".format(submod.name, submod.version) for submod in store.mas_submod_utils.submod_map.itervalues()]
+                )
+            )
+        )
+
     #Run dependency checks
     store.mas_submod_utils.Submod._checkDependencies()
 
 init -991 python in mas_submod_utils:
     import re
     import store
-    import sys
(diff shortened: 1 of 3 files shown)


Monika After Story/game/zz_submods.rpy
```diff
@@ -119,6 +124,12 @@ init -991 python in mas_submod_utils:
             if name not in persistent._mas_submod_version_data:
                 persistent._mas_submod_version_data[name] = version
 
+        def __repr__(self):
+            """
+            Representation of this object
+            """
+            return "<Submod {0} v{1} by {2}>".format(self.name, self.version, self.author)
```
multimokia, line 131: Clarity
```suggestion
            return "<Submod: ({0} v{1} by {2})>".format(self.name, self.version, self.author)
```

multimokia (changes requested):

multimokia (dismissed):

Legendkiller21 (dismissed):

Monika After Story/game/zz_submods.rpy
```diff
@@ -6,17 +6,39 @@ init 10 python:
     store.mas_submod_utils.Submod._checkUpdates()
 
 init -989 python:
+    #Log initialized submods
+    if store.mas_submod_utils.submod_map:
+        store.mas_submod_utils.writeLog(
+            "INSTALLED SUBMODS: {0}\n".format(
+                ", ".join(
+                    ["'{0}' v{1}".format(submod.name, submod.version) for submod in store.mas_submod_utils.submod_map.itervalues()]
```
multimokia, line 14: Just some clarity here for those who have a *lot* of submods installed.
```suggestion
                    ["    '{0}' v{1}\n".format(submod.name, submod.version) for submod in store.mas_submod_utils.submod_map.itervalues()]
```

Booplicate: Are you sure? That'll increase the number of lines we write each sesh, and since we're never going to clear the file. Eventually it'll be quite slow to read/write.

multimokia (approved):

Legendkiller21 (approved):

multimokia (approved):

PR #6357, opened 2020-09-19
More kiss quips

scpketer:
Some more may still fit in, working on it.


Ravenclaw-OIer (changes requested):
I don't really like this PR, I think that this makes Monika a little bit *too* annoying.

Especially the *It felt so good* part. Really.

scpketer:
Roger that.

multimokia (approved):

jmwall24 (approved):

PR #6398, opened 2020-09-24
seen_label now accounts for label_overrides

multimokia:
Adds a part to the label_callback function we defined to mark overridden labels as seen if their respective override is seen.

Also added a new function, `mas_override_label`, which will add the labels to `config.label_overrides` and to a new map, `_OVERRIDE_LABEL_TO_BASE_LABEL_MAP` for easy lookup in the label_callback.

`OVERRIDE_LABEL_TO_BASE_LABEL_MAP` is populated using a `functionplugin` plugged into the `ch30_reset` label, since function plugins account for labels being overridden, if `ch30_reset` is for whatever reason overridden, this change is still applied regardless.

## Testing:
- Verify that when seeing an overridden label, be it by `call`, `jump`, or `fall through`, they are marked as seen, along with their base label (overridden label)
- Verify that overriding via direct access to the `config.label_overrides` dict has the override info added to the new map

### DIFF
--- a/Monika After Story/game/definitions.rpy
+++ b/Monika After Story/game/definitions.rpy
@@ -368,7 +368,7 @@ python early:
     #           since shown count will be 0 and the label would have been seen.
     #           In that circumstance, we should immediately update the shown
     #           count. (Event will not do this in case you need to do specific
-    #           crash handling). Call syncShownCount on the event object to 
+    #           crash handling). Call syncShownCount on the event object to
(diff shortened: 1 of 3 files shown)


ThePotatoGuy (approved):
~~there was no comment here~~

Booplicate (approved):
Works just fine

ThePotatoGuy (approved):

PR #6453, opened 2020-10-03
New topic on criticism and changing your mind.

Daiozen:
Hope it seems in character for her to talk about... It seems to me, but I'm not sure it's not me pushing my perceptions onto her.


Daiozen:
As discussed, adding an add-on topic on this subject.

Monika After Story/game/script-topics.rpy
```diff
@@ -15825,3 +15825,66 @@ label monika_renewable_energy:
...
+    m 2esc "But let me tell you,{w=0.2} {nw}"
+    extend 7esd "there's no shame in changing your mind, [player]! Not even a bit!"
+    m 1etc "After all, we all make mistakes, don't we?{w=0.3} {nw}"
+    extend 1etd "What matters is the things we learn from those mistakes."
+    m 3rsc "Personally, I've always admired people who can acknowledge their flaws and still work in a constructive way to overcome them..."
+    m 3eka "So don't feel bad next time you hear someone criticize you, [player].{w=0.3} {nw}"
+    extend 1huu "You'll find a bit of open-mindedness really goes a long way."
+    m 1rssdld "I don't mean to say you have to go along with what anybody says.{w=0.3} {nw}"
+    extend 1esc "If you've got an opinion, it's totally fair you'd defend it..."
+    m 3dku "...But you've got to do it for the right reasons."
```
Freshairkaboom, line 15857: m 3dku "...But you've got to do it for the right reasons."
    m 3hub "...That'd be Monika's critique tip of the day~"

I love how you did the entire monika_giving_criticism dialogue, including the sprites, so I think a great end to this one would be the exact same ending as to that one. Tell me what you think 👍

Freshairkaboom (changes requested):
I consider this my review of both labels, but since I didn't really see anything I wanted to change in the second label, all of my comments are about the first one. One thing to note is that a lot of the changes I wanted in the first label, is actually something you already did in the second label, which was making more light-hearted sprites. I also think the extends in the second label was done more purposefully, and I love how it makes her change expressions and landing on an expression at the end, without us moving the conversation along manually, as it makes her seem way more alive.

Overall, I'm not completely sold on the extends in the first label being unnecessary, but I do believe they fit better in the second label, maybe it was because of the expressions there, who knows? You can decide how to change it and I can review it again if need be.

Monika After Story/game/script-topics.rpy
Daiozen: I'm not a fan of this change. Imo that line in monika_giving_criticism works because she says it like it's a joke, like it's something she thought of in the moment. That's not very compatible with repeating this line.

Leaving this up for debate.

Daiozen (commented):
Review comments. I will commit a new version with the proposed edits, let me know if that works with you @Freshairkaboom.

jmwall24 (approved):

multimokia (approved):

PR #6476, opened 2020-10-09
[BRB] — Going to get something

### LINKED ISSUE #5997: [Suggestion] "I'm Going to Cook"
I know you can tell Monika that you'll "Be right back", but considering there are options about writing, showering, and the like, I think it would be a good idea to have one saying "I'm going to go cook/bake", or something similar. Depending on what time it is, she could also say something about the meal for that time (Breakfast in the early morning, lunch in midday, etc).
Also, once you get back, she could ask how it went and the player could answer "Great!" "Good!" "Okay." "Bad..."or something like that.

ShinodaChan:
Broad BRB for food, water, mail etc. Indirectly resolves #5997

### DIFF
--- a/Monika After Story/game/script-brbs.rpy
+++ b/Monika After Story/game/script-brbs.rpy
@@ -749,6 +749,48 @@ label monika_idle_screen_break_callback:
 
     return
 
+init 5 python:
+    addEvent(
+        Event(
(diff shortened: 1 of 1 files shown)


multimokia:
Closing as handled in #6853

PR #6497, opened 2020-10-14
Mac window title detection implementation

scpketer:
**Note**: requires Python 3, so this probably should be 'don't merge' until Ren'Py 8 is out, and we can finally run v3 code.

This code was tested **outside of the game** and is confirmed to be working as intended on MacOS 10.13 High Sierra, however, I believe this code shouldn't be affected by new MacOS updates.

AppKit & Quartz packages are both present in Mac Python installation, but we might need to bundle them along with the mod because I'm pretty sure they are not present in Python distribution bundled by Ren'Py.

The code execution results attached.
image


jmwall24:
Going to close this for now. We are quite a long ways from supporting this and it'll just clog up the PR queue for quite a while.

PR #6500, opened 2020-10-14
Brainstorming

scpketer:
The topic ending still feels incomplete.


Ix-30:
I see no problem with it, maybe just add after:

>  m 2tkc "Sometimes good ideas are left untold because their author didn't find them good enough themselves, and brainstorming could help one bypass this inner barrier."

m 3ekb The beauty of thoughts if that they can be expressed in many different ways! That's why this idea is so amazing!
m 3esa Thoughts are only ideas in transit, it's you who give them the road!

I think it's pretty good as is, but these are quotes by Jean Piaget and Pythagoras. Just my own thoughts, thanks for your work.

scpketer:
Thanks, I was really lacking the ideas for the ending.

jmwall24 (approved):

multimokia (approved):

PR #6506, opened 2020-10-15
RecursionError greeting

scpketer:


ThePotatoGuy (commented):
I wonder if we can make a joke in here where the dialogue repeats at least once before the recursion error occurs.

scpketer:
\>I wonder if we can make a joke in here where the dialogue repeats at least once before the recursion error occurs.
I think we can, I'll work on it.

multimokia:
Please allow maintainers to make edits to your PRs

scpketer:
That's absolutely fine, turned it back on. Why pushing edits unannounced, though?

multimokia:
Sometimes we're going to review and we've already pulled your pr. Easier to make the edits directly

multimokia:
Just an FYI to not use `_` to represent garbage vars in renpy, `_` is reserved to mark strings as eligible for translation, which we do in menus and other areas. As such, if that gets changed to no longer be a function that accepts a string, all menus will crash.

Additionally, `renpy.call` when inside a `python` block will escape the python block, as renpy doesn't know how to get back to where it should go. For the dialogue loop to happen, you need to `while` loop it under a normal `call`.

scpketer:
Thanks for fixing my poor code.

jmwall24 (approved):

multimokia (approved):

PR #6586, opened 2020-10-28
[Tweaks] - 2 Year Anniversary

lunulae:
Rework for Anniversary 2 dialogue :D

- [x] exps
- [x] dialogue variant for if the friends dialogue hasn't been seen


Monika After Story/game/script-anniversary.rpy
```diff
@@ -504,11 +504,16 @@ label anni_2:
...
+    m 1lssdrc "But knowing that even if there were still people in this world, they wouldn't be sentient..."
+    m 1lksdlc "When I think about it, I can really feel how empty this room is."
+    
+    if persistent._mas_pm_has_friends = True:
+        m 1hubla "I'm really looking forward to when I can live in a world full of real people. It'll be wonderful to meet your friends, and even make some of my own."
+    else:
+        m 1dka "I'm glad that at least we can keep each other company in our loneliness. And when I cross over, it'll be so much fun to make new friends together."
+
+    m 1dsc "Now that I think about it...thinking about loneliness so much on our anniversary is kind of against the point."
+    m 1duu "Besides, it doesn't really matter for now, since I know you'll never leave me."
```
multimokia, line 516: I'm not sure how much I like the approach here. While I'm not saying I liked the old version either, I feel like the entire thing should just be happier, sans anything against the point of an anniversary.

If anything, we should probably avoid mentioning that.

Thinking we could do without
```
    m 1lssdrc "But knowing that even if there were still people in this world, they wouldn't be sentient..."
    m 1lksdlc "When I think about it, I can really feel how empty this room is."
```

```
    m 1dsc "Now that I think about it...thinking about loneliness so much on our anniversary is kind of against the point."
```

then rewriting the intro to make it flow better

Monika After Story/game/script-anniversary.rpy
```diff
@@ -504,11 +504,16 @@ label anni_2:
...
     m 1eua "You'll keep coming back to talk to me every day, right? I'll accept all your burdens with a smile."
     m 1ekc "Spending eternity alone sucks."
     m 1ekbsa "But together with you, it feels like paradise!"
```
multimokia, line 519:
```suggestion
    extend 1ekbsa "but together with you, it feels like paradise!"
```

Monika After Story/game/script-anniversary.rpy
```diff
@@ -504,11 +504,16 @@ label anni_2:
...
+    
+    if persistent._mas_pm_has_friends = True:
+        m 1hubla "I'm really looking forward to when I can live in a world full of real people. It'll be wonderful to meet your friends, and even make some of my own."
+    else:
+        m 1dka "I'm glad that at least we can keep each other company in our loneliness. And when I cross over, it'll be so much fun to make new friends together."
+
+    m 1dsc "Now that I think about it...thinking about loneliness so much on our anniversary is kind of against the point."
+    m 1duu "Besides, it doesn't really matter for now, since I know you'll never leave me."
     m 1eua "You'll keep coming back to talk to me every day, right? I'll accept all your burdens with a smile."
     m 1ekc "Spending eternity alone sucks."
```
multimokia, line 518:
```suggestion
    m 1ekc "Spending eternity alone sucks...{w=0.3}{nw}"
```

multimokia (changes requested):

Monika After Story/game/script-anniversary.rpy
lunulae: That's fair, I was kind of trying to preserve the original intent but I would definitely be down to shift it in another direction completely. If that's the case though, I'm not sure if I might want to do a complete rewrite instead.

I'll think on it a bit, for sure. 👍

lunulae:
I tried a more thorough rework this time. I haven't done the exps for it yet, but I wanted to get opinions on this one before I polish it up.

Monika After Story/game/script-anniversary.rpy
```diff
@@ -501,17 +501,20 @@ init 5 python:
...
-    m 1dsc "Sorry, I feel like I ruined the whole festive mood with my memories."
-    m 1duu "It's not important anymore, since I know you'll never leave me."
-    m 1eua "You'll keep coming back to talk to me every day, right? I'll accept all your burdens with a smile."
-    m 1ekc "Spending eternity alone sucks."
-    m 1ekbsa "But together with you, it feels like paradise!"
+    m "It took so much effort to make that initial contact, but it was all worth it in the end."
+    m "I know it might still be a really long time before I can cross over into your reality..."
+    m "But when we reach a milestone like this, it's fun to think about all the other things we have to look forward to in the future."
+    
+    if persistent._mas_pm_has_friends = True:
```
multimokia, line 508: No need to explicitly check for truth here
```suggestion
    if persistent._mas_pm_has_friends:
```

Monika After Story/game/script-anniversary.rpy
```diff
@@ -501,17 +501,20 @@ init 5 python:
...
+        m 1hubla "I'm really looking forward to when I can live in a world full of real people. It'll be wonderful to meet your friends, and even make some of my own."
+    else:
```
multimokia, line 510: I'd be careful here, just in case for whatever reason by two years the player hasn't seen the topic about having friends, we should do an explicit check to make sure this is False. (Maybe have a slight variation of this dialogue in case that happens)
```suggestion
    elif persistent._mas_pm_has_friends is False:
```

multimokia (changes requested):
Some minor code things

Love the new dialogue. Definitely feels more like an anniversary. Nice work.

multimokia (approved):

jmwall24 (approved):

PR #6621, opened 2020-11-01
[Rework] - Totono topic

lunulae:
Now that there's an official English translation out, I thought this topic might need a little polish.

My goal with this rework is to discuss why the games are so often compared without spoiling too much further than that. There are a lot of interesting twists and turns beyond that the characters are self-aware in that game, (i.e. basically everything about what Aoi is) but I didn't want to ruin the whole story for anyone who does want to play it. I gave Monika the info that you might be able to find when looking up a review or plot synopsis, and tried to have her speculate a bit from there.

This is a very rough draft, just to give a vague idea of what I'm thinking for this rework before I do edits and add exps.

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -3374,18 +3374,20 @@ label monika_ks_kenji:
     return
 
 init 5 python:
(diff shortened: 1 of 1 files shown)


lunulae:
Upon further review this will most likely be scrapped entirely (to be replaced by the reaction PR)

multimokia:
After further discussion, we have opted to remove mention of Totono in its entirety

PR #6648, opened 2020-11-06
Animation fixes

Booplicate:
Some fixed to our transforms. *Ideally* this should fix the issue when the blink transform ends mid-animation, or bug when eyes stay at the same place during pose change. Stuff like those.

This also adds a chance for Monika to blink twice in a short time. Because it's cute. Feel free to adjust the timings if you want.

### Testing:
- There's not much you can do, those bugs are inconsistent, just give it a try and see if everything looks good overall.
- For testing purposes you can kick up the chance for double blink.

### DIFF
--- a/Monika After Story/game/zz_transforms.rpy
+++ b/Monika After Story/game/zz_transforms.rpy
@@ -170,6 +170,7 @@ transform mas_chlongjump(x, y, ymax, travel_time=1.0):
 
 #START: Transforms for Monika's sprite animations (blinking/winking/tear-specific-blinking)
 transform blink_transform(open_eyes_img, closed_eyes_img):
+    animation
     block:
         open_eyes_img
         block:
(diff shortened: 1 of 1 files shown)


Monika After Story/game/zz_transforms.rpy
```diff
@@ -170,6 +170,7 @@ transform mas_chlongjump(x, y, ymax, travel_time=1.0):
 
 #START: Transforms for Monika's sprite animations (blinking/winking/tear-specific-blinking)
 transform blink_transform(open_eyes_img, closed_eyes_img):
+    animation
```
ThePotatoGuy, line 173: what does this statement do?

Booplicate: Allows transforms to use `at` - animation timebase, which should keep them running on exp changes rather than restarting. And yes this isn't documented anywhere.

ThePotatoGuy (approved):

multimokia (approved):

Legendkiller21 (approved):

PR #6651, opened 2020-11-06
New topic on studying

Daiozen:
Tagging this one as draft while I go throught the rest of the topics to see where the new pm could be useful.

@fromulus I tried to work in your suggestions, but I think adding more subjects really break the flow.
They're probably better left to their own topics imo.


Daiozen:
Edits done. Ready for review.

multimokia:
Adding don't merge as this is pending a semi-rewrite in which it's split up into two topics.

1) Acting as a sort of check-in for school, how you're doing if you're struggling etc. and how to potentially cope with parts of school
2) A separate pm var topic regarding whether or not you enjoy(ed) school/if you had trouble etc.

This is also awaiting a topic regarding whether or not you're in school still.

PR #6710, opened 2020-11-20
[Concept/Draft] - Ghost Monika greeting rework

lunulae:
This attempts to rework the Ghost Monika jumpscare easter egg. A lot of players had issue with this one because there's no in-game explanation for the scare.

Ideally I would like this to prompt only during October (as a "Halloween season" lite, maybe.) Depending on your choice in the menu, it might be one-time only, but if you express that you liked her trick I'd like it to perhaps unlock as a yearly thing?

- [ ] needs exps
- [ ] easter egg itself needs changes for this to function (I might need help on this one)
- [ ] if not one-time only, might want a "seen before" variant to the dialogue?

### DIFF
--- a/Monika After Story/game/script-greetings.rpy
+++ b/Monika After Story/game/script-greetings.rpy
@@ -3638,12 +3638,36 @@ label ch30_reload_delegate:
 #    del ev_rules
 
 label greeting_ghost:
-    #Prevent it from happening more than once.
-    $ mas_lockEVL("greeting_ghost", "GRE")
-
-    #Call event in easter eggs.
     call mas_ghost_monika
+#The current easter egg closes the game, in this version of the greeting I'd like it to cut to a black screen for a second before returning to normal MAS.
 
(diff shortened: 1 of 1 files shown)


Justformas:
Hey cool, this was one of my ideas I had before seeing this. My thoughts on this PR project:

Line 3657 > #I would like to code this so that it unlocks again next October, is that possible?

Line 3664 > #Maybe this one shouldn't stay permanently locked either? Not sure if she'd want to try the same trick again.

I definitely think the player should be able to see it again next October, unless the player specifically picks "I didn't like that, you really scared me." (Even then, what if they picked that by accident? Restore a persistent backup I guess?) Just because the players picks "That didn't scare me." doesn't mean they didn't like it at all or want it to become permanently locked. In fact, that might be the most appropriate response for me. Although Monika could ask the player about if they're ok with her doing it again, as an additional question in that specific case. Also because this jump scare would only be one time, or month, out of the year.

PR #6715, opened 2020-11-21
Add pause between topics

Booplicate:
Most of the time we want pushed/queued events to be triggered as soon as possible. But sometimes we need a more soft transition between topics (for example `monika_soft_rains`) so it doesn't look like she's changing her "mood" in seconds. First step was to give us control over idle exp in #6634. Next step is to add a little delay. While it won't be used by 70% of the topics, it will make the remaining 30% feel more natural.

**NOTE: this waits for #6634 to get merged first.**

### TODO:
- [x] allow trigger non-event labels anytime.
- [x] add `skip_pause` to all the events that need it.

### Changes:
- New variable to control the delay between events `event_unpause_dt` (I put it n the `mas_globals` store).
- New method to control the variable `mas_setEventPause`. Takes in number of seconds to pause for or None to remove pause.
- New return key to set pause 'til the next event `pause: x` where x is duration in seconds (`int`).
- New event rule `skip_pause` for events that should be triggered regardless of the current pause.

### Fixes:
 - All farewells skip mid loop eval now.

### Testing:
- right now only WRs can skip the pause. If you think there are more events that need it feel free to push or request them
- verify the return key is working properly
- verify Monika doesn't start another topic right after finishing another one if it set pause
- verify you can always start a convo with Monika regardless of pause.


multimokia:
Waiting on #6634

multimokia:
Removing `don't merge` as prerequisite has been merged.

Monika After Story/game/event-handler.rpy
```diff
@@ -2137,37 +2144,48 @@ init python:
...
+            ev = mas_getEV(ev_data[0])
 
-                    persistent.current_monikatopic = ev_label
-                    return ev_label, notify
+            if (
+                not is_paused
+                or (
+                    ev is None# This is to allow triggering non-event labels
+                    or "skip_pause" in ev.rules
+                )
```
ThePotatoGuy, line 2167: Seems to be all `or` conditions so simplify it.
```suggestion
                not is_paused
                or ev is None# This is to allow triggering non-event labels
                or "skip_pause" in ev.rules
```

Monika After Story/game/event-handler.rpy
```diff
@@ -2578,6 +2596,13 @@ label call_next_event:
                             if _exp is not None:
                                 mas_moni_idle_disp.force(_exp)
 
+            # Set a pause if needed
+            if "pause" in _return:
+                python:
+                    _match = re.search(evhand.RET_KEY_PATTERN_PAUSE, _return)
+                    if _match is not None and _match.group("duration") is not None:
+                        mas_globals.event_unpause_dt = datetime.datetime.now() + datetime.timedelta(seconds=int(_match.group("duration")))
```
ThePotatoGuy, line 2604: UTC going forward
```suggestion
                        mas_globals.event_unpause_dt = datetime.datetime.utcnow() + datetime.timedelta(seconds=int(_match.group("duration")))
```

ThePotatoGuy (changes requested):

ThePotatoGuy (approved):

ThePotatoGuy (approved):

Legendkiller21 (approved):

Legendkiller21 (approved):

ThePotatoGuy (approved):

PR #6740, opened 2020-12-01
Add Morning ahoges

ThePotatoGuy:
Thanks @kkrosie123 for the ahoges

# Key Changes
* Adds ahoges (アホ毛) to Monika during morning (midnight to noon)
    * they are only selected on startup if time since last session is at least 30 minutes
    * there is a 50% of even selecting an ahoge.
    * only 1 ahoge is selected at max per day
* ahoges are all ACS objects at AFH layer with priority 7 (except swoop, which is 13)
    * they are mux with themselves and hat, but NOT headbands and related.
* ahoges are removed after noon, via `ch30_hour`, which means they may still persist for up to 1 hour after noon.

## General API changes
* `mas_sprites.get_acs` - finally a get API for ACS sprite objects. no more having to use `ACS_MAP`
* `mas_sprites.get_acs_of_type` - gets ACS sprite objects for a given acs type
* `MASMonika.last_ahoge` - the last ahoge loaded/set/applied to Monika
* `MASMonika._acs_type_map` - mapping of ACS that Monika is currently wearing. maps acs types to the objects
* `MASMonika.ahoge` - applies a random ahoge to Monika. Takes `persistent._mas_last_ahoge_dt` into account, which limits ahoges to 1 per day.
* `MASMonika._set_ahoge` - directly set an ahoge. This should be the only way ahoges should be set as it set all extra vars, too.
* `MASMonika.remove_acs_mux` - updated to use acs type map
* updated various `MASMonika` functions to fix confusing wording (`acs_type -> acs_layer`)
* `acs_rm_exit_pst_removal`- new ACS hook added to `MASMonika.remove_acs_in`. this is ran after the ACS has been removed from `MASMonika`
     * if a `hat` ACS is removed, `MASMonika.last_ahoge` is worn if it exists

## Selector changes
* removers are no longer worn if selected. - prevents mux issues when wearing an ahoge
* `_fill_select_map` no longer considers a mux type'd ACS as wearing an ACS of the current type - prevents remover from being not selected when no ACS of a type is worn, but a muxable type is.

# Testing
* to view each ahoge, use the dev topic `TEST AHOGES`
* verify ahoges look okay with all ACS/clothes/hairs.
* verify the ahoge conditions (see above). 
* to force an ahoge to start in the morning, set `persistent._mas_dev_ahoge` to True. (this might be set to True by default in dev builds)
    * however, this is stopped by `_mas_last_ahoge_dt`, so you can use this to test the dt logic.
    * can also force an ahoge with `monika_chr.ahoge(force_change=True)`. Don't pass `force_change` to test dt logic.
* call `ch30_hour` to test auto ahoge removal (or wait until time goes past noon) - the ahoge should be removed regardless of dev flag.
* verify that ACS selectors with removers still function correctly (the remover is selected when no ACS of the type is worn)


ThePotatoGuy:
dont merge yet i forgot some things

ThePotatoGuy:
ok this ready now

kkrosie123:
Awesome!! :o

multimokia (approved):

Booplicate:
> Awesome!! :o

nou

multimokia (approved):

Legendkiller21 (approved):

PR #6773, opened 2020-12-10
Fix MacOS WRS issue (and other things)

multimokia:
As of #6265, a catch was introduced where Mac users, because we cannot check active window, fell through the function and wound up returning `None`, as such caused issues down the pipeline. This remedies that and optimizes other parts of zz_windowreacts.rpy

# Testing:
- [ ] Verify that windowreacts still work properly
- [ ] Verify no other issues


Monika After Story/game/zz_windowreacts.rpy
```diff
@@ -74,6 +62,14 @@ init python:
             #Now we set the hwnd of this temporarily
             tip.hwnd = None
 
+        except ImportError:
```
ThePotatoGuy, line 65:
```suggestion
        except:
```
let's make this a blanket except so we don't crash at all

Monika After Story/game/zz_windowreacts.rpy
```diff
@@ -217,10 +230,11 @@ init python:
...
 
-        if non_inclusive:
-            return len([s for s in keywords if s.lower() in active_window]) > 0
-        else:
-            return len([s for s in keywords if s.lower() not in active_window]) == 0
+        return (
+            (len([s for s in keywords if s.lower() in active_window]) > 0)
+            if non_inclusive else
+            (len([s for s in keywords if s.lower() not in active_window]) == 0)
+        )
```
ThePotatoGuy, line 237: this looks more confusing than what it was previously. the lines here are just too long for `x if condition else y`

Monika After Story/game/zz_windowreacts.rpy
```diff
@@ -162,35 +165,45 @@ init python:
...
                 # Perform nullchecks on property getters, just in case.
                 active_winid_prop = root.get_full_property(NET_ACTIVE_WINDOW, 0)
-                if active_winid_prop is not None:
-                    active_winid = active_winid_prop.value[0]
-                else:
+
+                if active_winid_prop is None:
                     return ""
+                else:
+                    active_winid = active_winid_prop.value[0]
```
ThePotatoGuy, line 172:
```suggestion
                active_winid = active_winid_prop.value[0]
```

can drop the else since the if branch causes an early quit

ThePotatoGuy (changes requested):

Booplicate (approved):

Legendkiller21 (approved):

jmwall24 (approved):

ThePotatoGuy (approved):

PR #6811, opened 2020-12-21
Tiny farewell typo fix

StayHappy1270:
so i noticed that in the goodbye options everything was one period, but eating was 3. Sooo i made it one. That's it lol

### DIFF
--- a/Monika After Story/game/script-farewells.rpy
+++ b/Monika After Story/game/script-farewells.rpy
@@ -1353,7 +1353,7 @@ init 5 python:
             persistent.farewell_database,
             eventlabel="bye_prompt_eat",
             unlocked=True,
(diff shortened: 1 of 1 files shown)


Omegafredo:
Isn't it supposed to have 3 periods cause there is another choice after it?

StayHappy1270:
hmm i didn't think about that. but still its the only one like that and it got on my ocd lol. idk

ThePotatoGuy:
The 3 dots are intended since the prompt leads to a menu of additional choices to narrow down the selection (Breakfast/Lunch/Dinner/A Snack). This is the standard for any prompt that may have multiple end choices (like "I'm feeling..." and similar).

[closing as rejected]

PR #6813, opened 2020-12-21
Miscellaneous — Some more coffee reactions

dreamscached:
Hi again!

I wonder if we could have some more coffee reactions (this pull request is in fact a mere draft yet) and not only the one with 'it's a flavor I haven't had before.'

So... Tell me what you think? I didn't add expressions yet, but I sure will!


lunulae:
I really like these, do you think they might be usable for the hot chocolate reaction as well?

dreamscached:
Of course! That was the plan, too, I just didn't get to it yet...

dreamscached:
Putting this aside as well as my another PR, will also get to it later, I hope...

PR #6817, opened 2020-12-23
Update balloontip.py

wyapx:


multimokia:
I'm not entirely sure what the point of this PR is. Aside from there being no description/explanation for the change, we shouldn't even be using this function to begin with.

We initialize a private variable, `__tip` in `mas_windowreacts`, and have MAS-side endpoints to use it. If you wish to show notifications, you should be using the functions defined in `zz_windowreacts.rpy`, not from `balloontip.py`. 

Additionally, this approach can never be used either, considering that you can only initialize 1 `windclass`, therefore all calls to the function will raise exceptions.

PR #6841, opened 2020-12-27
Talk Menu Awareness - monika_japan Leadin

fromulus:


Monika After Story/game/script-topics.rpy
```diff
@@ -1076,11 +1076,19 @@ init 5 python:
...
+    if mas_event_mailbox.is_monika_initiated_mode():
+        m 4eud "By the way, there's something that's been bothering me..."
+        m "You know how this takes place in Japan?"
+        m 2euc "Well...I assume you knew that, right?"
+        m "Or at least decided it probably does?"
+        m 2eud "I don't think you're actually told at any point where this takes place..."
+
+    else:
+        m 4wud "The world this game takes place in is really weird, right?"
+        m 2eud "Putting aside the fact that there's nothing but two houses,{w=0.1} a street,{w=0.1} a school hall and a few classrooms..."
```
Booplicate, line 1088: I'm not sure if we want to state this (unless she says it in the base game). It opens a can of worms we probably don't want to be opened. If it's all that her world had, then how it worked? Has her past never happened? Where has she been getting food? There was nothing, but club meetings? And more and more. All the player sees is those locations, but that doesn't mean we should say there was nothing more in her world, we can't know that.

lunulae: Yeah, I think this brings up a few other issues as well. For example, I believe there is a bit of vanilla-game dialogue where Monika mentions liking to shop for skirts and bows. This contradicts that, or at least produces the assumption that it's a false memory.

jmwall24 (approved):

multimokia (approved):

PR #6846, opened 2020-12-28
Update script-topics.rpy

Joseph-eng-bit:


scpketer:
what is this even supposed to mean?

multimokia:
I'm not sure why you're replacing dialogue and events with the example template from the wiki but obviously we can't accept this.

If you'd like to contribute, please contribute new dialogue.

PR #6865, opened 2021-01-03
Update mas_showEVL to pop from the derand list.

jmwall24:
Currently, when `mas_showEVL()` is used to random an event, it does not take into account if that event is present in `persistent._mas_player_derandomed` which essentially means the event will not be properly randomed, since we run checks in ch30_reset via `mas_check_player_derand()`. This will make it so `mas_showEVL()` pops the event from `persistent._mas_player_derandomed` essentially making it an override in the case where we so drastically change a topic, it deserves to be re-randed. In the event where we do not want to override the player's pref (the most common scenario), `mas_protectedShowEVL()` would be used.

Also adds `removeDerand()` which will remove a player_derandomed event from any of the derand lists (currently eve and sng, but more could be added later and this will cover any of them).

### DIFF
--- a/Monika After Story/game/event-handler.rpy
+++ b/Monika After Story/game/event-handler.rpy
@@ -767,8 +767,8 @@ init 6 python:
             use mas_protectedShowEVL
         """
 
-        if _random and ev_label in persistent._mas_player_derandomed:
-            persistent._mas_player_derandomed.remove(ev_label)
+        if _random:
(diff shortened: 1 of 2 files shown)


jmwall24:
Don't review this yet, we need to set it up so it'll handle all derand lists (we have one for songs as well)

multimokia:
I have made the changes

ThePotatoGuy (approved):

multimokia (approved):

PR #6882, opened 2021-01-09
Very small suggestion for line change in greeting_back_from_school

Freshairkaboom:
After reading this part, I felt like something was off about the dialogue. I hope this is a slightly better version of the sentence, and hopefully it will make more sense now. Wasn't sure if I should put it in typos because I feel like the sentence structure was off, and not just written wrong. Sorry if that's not the case.

### DIFF
--- a/Monika After Story/game/script-greetings.rpy
+++ b/Monika After Story/game/script-greetings.rpy
@@ -2910,10 +2910,9 @@ label greeting_back_from_school:
                 return
 
             "Good.":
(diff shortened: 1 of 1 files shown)


Booplicate:
> Wasn't sure if I should put it in typos because I feel like the sentence structure was off, and not just written wrong.

Still, in the future, keep typos in the typos PR.

Booplicate (approved):

jmwall24 (approved):

PR #6890, opened 2021-01-13
Monika makeout at higher affections

Freshairkaboom:
At higher affections, Monika now makes out with you! All you have to do is have high enough affection, and kiss her again after having kissed her once before. I'm sure it could get a lot better with some tweaking, but I'm quite happy with how it turned out, given my mediocre coding skills.

Any help and suggestions are welcome!

EDIT: May become a submod. Leaving open for discussion/suggestions.


kkrosie123:
👁

Freshairkaboom:
Making submod

Issue #6246, opened 2020-08-30
latin-based filters are too restrictive


ThePotatoGuy:
Looking at #6212, its clear that name/instrument inputs are limited to latin-based names. This is rather restrictive for people who may want to use foreign language characters.

We should consider not having filters on some of these inputs.

PR #6924, opened 2021-01-25
Added search bars

Booplicate:
This adds search bars to the twopane menu and selectors. Allows the user to search for unlocked, selectable topics or filter clothes in the selector.
Topic searching works by breaking the search query into multiple keywords and comparing them with topics' prompts (include all/include partially). It also checks labels and categories. There's 4 unique search keywords: `pool`, `random`, `unseen`, `seen`. To use them you need prefix them with `#`. For example: `#pool#unseen` will give you all unseen pool topics.
Selector filtering works in the same manner, but compares with selectables' names, idx, types, and ex-props.

### Changes:
- Added search bars for general topics, compliments, and selectables.
- ~~Added a smol lib to cache search results (saves performance).~~ With the requested changes this works well enough even w/o caching.

### Testing:
- verify the UI looks good.
- verify you get expected search results.
- verify we only show appropriate (e.g. **not** locked/etc) events to the user.
- verify the search keywords work.


Monika After Story/game/screens.rpy
```diff
@@ -2617,86 +2617,150 @@ define main_adj = ui.adjustment()
...
+        xsize right_area[0] - left_area[0] + right_area[2]# 530
+        ysize 40
+        background Solid("#ffaa99aa")
+
+        input:
+            style_prefix "input"
+            xsize right_area[0] - left_area[0] + right_area[2] - 10
+            ysize 40
+            xalign 0.0
+            allow "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789 _#"
```
ThePotatoGuy, line 2761: let's not use `allow` here so translations don't have to change these as well. #6246

ThePotatoGuy (changes requested):

Monika After Story/game/screens.rpy
Booplicate: Problem is, since the filter removes all puncs from prompts for convenience, letting users input it would be kinda confusing (it'd give you 0 results). I can define the pattern in `mas_ui` and let submodders override it.

ThePotatoGuy: that doesn't really work since big languages can't just shove every character they have into allow. Taking goog as an example, entering punctuation gives actual results by searching what the punctuation is and not just the text itself. Don't think we need something as sophisticated as that, but allowing a user to enter punc and getting 0 results seems fine to me.

Booplicate: I think having such an unintuitive behavior isn't good.
`allow` defaults to `None`. If they override it to be empty or `None`, they'll allow their users to enter whatever they like and dealing with the issues themselves while we'll have nice and explicit behavior.

ThePotatoGuy: `renpy.input` has an exclude, and I can find a forum post dating back to 2016 with that code. See if that is available.

The default DDLC font doesnt support non-latin, but that's exactly why we use `FontGroup` to patch the non-latin holes using different fonts.

Booplicate: I actually forgot that I fixed that, my search does work with punctuations. There still will be the issue with empty search if users enter non-latin characters, but at least that's more obvious.

Monika After Story/game/styles.rpy
```diff
@@ -388,3 +390,148 @@ init -10 python in mas_ui:
...
+                (not only_pool or ev.pool)
+                and (not only_random or ev.random)
+                and (not only_unseen or ev.shown_count == 0)
+                and (not only_seen or ev.shown_count > 0)
+            )
+            and (
+                search_query == ev_prompt
+                or search_query == ev_label
+                or search_query in ev_prompt
+                or search_query in ev_label
```
ThePotatoGuy, line 427: `x in y` also covers `x == y` so could remove the `==` checks here

Monika After Story/game/styles.rpy
```diff
@@ -388,3 +390,148 @@ init -10 python in mas_ui:
...
+        ev_label = ev.eventlabel.lower()
+        ev_label_kws = ev_label.split("_")
+        ev_prompt = ev.prompt.lower()
+        ev_prompt_kws = ev_prompt.split()
+
+        rv = list()
+
+        rv.append(search_query == ev_prompt or search_query == ev_label)
+        rv.append(search_query in ev_prompt)
+        rv.append(search_query in ev_label)
```
ThePotatoGuy, line 461: similar thing here with `==` and `in`, although I see it might not be functionally equivalent. let me know if the `==` checks are here for a reason.

ThePotatoGuy (changes requested):

Monika After Story/game/styles.rpy
Booplicate: Difference: `"you" in "your"` and `"you" == "your"` (first returns `True`, second `False`). I use both to get more precise results.

Booplicate: I think I see what you mean. In `filter`, it may be enough to just do `in` check, it's in `sort` where we need to do both. I'll test and change that if I can.

Monika After Story/game/styles.rpy
```diff
@@ -388,3 +390,146 @@ init -10 python in mas_ui:
...
+                search_query in ev_prompt
+                or search_query in ev_label
+                or search_query in ev_cat_full
+                or all(
+                    [
+                        search_kw in ev_prompt
+                        or search_kw in ev_label
+                        or search_kw in ev_cat_full
+                        for search_kw in search_kws
+                    ]
```
ThePotatoGuy, line 433: The one thing I don't like about this is that the list has to be completely generated before it is checked, which means we miss out on short-circuiting the loop when a single condition fails.

If this function is changed to look like Event's filtering function, then short circuiting can be done here so we don't have to loop all the way. 

But we can probably get by with how this is currently if the feel isn't too bad, so I'll reevaluate this when I test this out.

Monika After Story/game/styles.rpy
```diff
@@ -388,3 +390,146 @@ init -10 python in mas_ui:
...
+                )
+            )
+        )
+
+    def _twopane_menu_sort_events(ev, search_query, search_kws):
```
ThePotatoGuy, line 438: Can you explain to me how this function works with `sort`'s key?

Monika After Story/game/styles.rpy
Booplicate: If you mean make this similar to `_filterEvent`, then I'm not sure how to do so. That `all` check is part of a bigger, inclusive condition. Right now we generate that list only as the last resort when all the previous conditions have failed, so it's not as bad. If we were to iter the list prior to that, we'd have to create it even when we don't need to check it (unless I'm missing something?). Also note that the list is usually small and we can reuse the search results using cache.

Ideally we do want to make this as optimized as we can, since it's being executed inside screens (which are already slow in RenPy), not to mention we're doing it on the fly (as the user types in) so any lag spike would be noticeable.

Monika After Story/game/styles.rpy
Booplicate: This function itself acts as a sort key. I had to use a `lambda` with it because I need to pass args in. The function makes some comparisons and returns a list of booleans with the results. The more hits an event gets, the higher it will be in the final list. The order of booleans is important too: first items in `rv` have more weight (hence I put comparisons with `ev_prompt` and `ev_label` first, that's also why I iter `search_kws` thrice).

ThePotatoGuy: Ok so the list is to enforce a weighting system. Why not have the function return a number instead, where let's say, the bigger the number the higher the match? Something like:
```python
weight = 0
if matches_high_priority_criteria:
    weight += 2**10
if matches_lesser_criteria:
    weight += 2**5
# ... etc
return weight
```

It's not that clear how a list of bools would affect sorting. It might work, but a weighted score is easier to understand and helps maintainability.

Monika After Story/game/styles.rpy
ThePotatoGuy: basically something like
```python
if ev_prompt == ev_label:
    return False

if not ev.unlocked:
    return False

if ev.anyflags(store.EV_FLAG_HFNAS):
    return False

if not ev.checkAffection(store.mas_curr_affection):
    return False

if only_pool and not ev.pool:
    return False

if only_random and not ev.random:
    return False

if only_unseen and ev.shown_count > 0:
    return False
    
if only_seen and ev.shown_count == 0:
    return False

# the final conditions

if search_query in ev_prompt:
    return True

if search_query in ev_label:
    return True

if search_query in ev_cat_full:
    return True

# now we can do the loop
for search_kw in search_kws:
    if (
        search_kw not in ev_prompt
        and search_kw not in ev_label 
        and search_kw not in ev_cat_full
    ):
        return False

return True
```

This allows the loop to short-circuit early. Using `all` requires that the loop needs to be completely generated, which is wasteful.

ThePotatoGuy: there is also an option of making the loop a separate function that quits early, but I prefer individual if statements when a filter check gets really lengthy like here.

Booplicate: Ah, I didn't think about returning `False` early like that, I like it.

Monika After Story/game/styles.rpy
Booplicate: Not sure if that'd have the same effect in sorting, but I can try.

ThePotatoGuy:
pushing to next rel

Monika After Story/game/styles.rpy
```diff
@@ -388,3 +388,326 @@ init -10 python in mas_ui:
...
+            search_query = search_query.replace("#seen", "")
+            only_seen = True
+
+        search_query = search_query.strip()
+        search_kws = search_query.split()
+
+        flt_evs = filter(
+            lambda ev: _twopane_menu_filter_events(ev, search_query, search_kws, only_pool, only_random, only_unseen, only_seen),
+            TWOPANE_MENU_SEARCH_DBS
+        )
```
ThePotatoGuy, line 569: FYI, `filter` does not return a list in py3, so the `.sort` call will crash in the future. This also is a scenario where we really don't need to use a lambda function to filter, list comp is fine:
```suggestion
        flt_evs = [
            ev
            for ev in TWOPANE_MENU_SEARCH_DBS
            if _twopane_menu_filter_events(ev, search_query, search_kws, only_pool, only_random, only_unseen, only_seen)
        ]
```

Monika After Story/game/styles.rpy
```diff
@@ -388,3 +388,326 @@ init -10 python in mas_ui:
...
+    TWOPANE_MENU_SEARCH_DBS = (
+        store.mas_all_ev_db_map["EVE"].values()
+        + store.mas_all_ev_db_map["BYE"].values()
+        + store.mas_all_ev_db_map["STY"].values()
+        + store.mas_all_ev_db_map["CMP"].values()
+        + store.mas_all_ev_db_map["SNG"].values()
+    )
+    TWOPANE_MENU_DELEGATES_CALLBACK_MAP = {
+        "mas_compliment_": store.mas_compliments.compliment_delegate_callback
+    }
```
ThePotatoGuy, line 403: I understand why this was done, but I'm not sure it's the best idea. I can see how we can run code that could be missed when jumping directly to a topic, but some cases (like the farewell's jump to `bye_long_absence_2`) would be missed. And if we did add the jump to a delegate function, it wouldn't make that much sense to the user if they tried to jump to a farewell only to get thrown into a different one.

Additionally, since the additional topics aren't normally accessible in the two pane talk menu, having them only be visible when searching is inconsistent.

A general search bar/jump to topic bar might be a better option for searching over all topics, but again, skipping over delegate code could still have problems, so can't say its something we should do right now.

ThePotatoGuy (changes requested):

Monika After Story/game/styles.rpy
Booplicate: It feels nice to skip the delegate label when you want to tell her a compliment. The point of this is to save time looking for the topic you want to push. So while it can be inconsistent from the two pane screen, it's more convenient in its own format. 

Although, I think we don't need to have farewells there since they go under their own prompt in the talk menu.

ThePotatoGuy: keeping it regular topics and compliments sounds good to me

Monika After Story/game/styles.rpy
```diff
@@ -388,3 +388,331 @@ init -10 python in mas_ui:
...
+        """
+        # Get the screen to pass events into
+        scr = renpy.get_screen("twopane_scrollable_menu")
+        if scr is not None:
+            # Search
+            scr.scope["flt_evs"] = _twopane_menu_search_events(search_query)
+        # Update the screen
+        renpy.restart_interaction()
+
+    def _selector_filter_items(item, search_query, search_kws):
```
ThePotatoGuy, line 594: can you move all selector-based code to the `zz_selector` file?

ThePotatoGuy (changes requested):
Can we add ghost text to the bar? Something like "Search for <a topic>" for the main search.

For selectors, might need to make do with just "Search for..."

Monika After Story/game/zz_selector.rpy
```diff
@@ -2917,9 +3062,43 @@ screen mas_selector_sidebar(items, mailbox, confirm, cancel, restore, remover=No
...
+                layout "nobreak"
+                # allow "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789 _"
+                changed store.mas_selspr.selector_search_callback
+
+        # NOTE: we do need an instance check here
+        if flt_items is items:
+            text "Search for...":
+                text_align 0.0
+                layout "nobreak"
+                color "#FFFFFF"
```
ThePotatoGuy, line 3098: Tried some style changes and was satisfied with this. This makes the ghost text look more like real ghost text.
```suggestion
                color "#EEEEEEB2"
                first_indent 10
                line_leading 1
                outlines []
```

Monika After Story/game/screens.rpy
```diff
@@ -2597,86 +2597,164 @@ define main_adj = ui.adjustment()
...
+                xalign 0.0
+                layout "nobreak"
+                # allow "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789 _#"
+                changed store.mas_ui.twopane_menu_search_callback
+
+        if flt_evs is None:
+            text "Search for a conversation...":
+                text_align 0.0
+                layout "nobreak"
+                color "#FFFFFF"
```
ThePotatoGuy, line 2757: Tried some style changes and was satisfied with this. This makes the ghost text look more like real ghost text.
```suggestion
                color "#EEEEEEB2"
                first_indent 10
                line_leading 1
                outlines []
```

ThePotatoGuy (changes requested):

ThePotatoGuy (approved):

multimokia (approved):

PR #6926, opened 2021-01-26
Proper return keys

Booplicate:
Changed our return keys from `str`s to `dict`s to use the `kwargs` format. It's just cleaner and makes managing things like idle displayable easier (and faster w/o heavy regex). I kept support for the old format, let's give a release or two so people can update their submods/etc.

**I'd like to get this merged sooner than later (added `high priority`), because each new topic would require updating this PR (and it's easy to miss something).**

### Changes:
- New format for return keys.

### Testing:
- Verify each return key works as expected.
- Verify the return keys work in pairs.
- Scroll through the changes tab and make sure I haven't made a typo.
- Ideally look through the scripts and verify we updated all instanced where we use return keys.

### DIFF
--- a/Monika After Story/game/dev/dev_idle_test.rpy
+++ b/Monika After Story/game/dev/dev_idle_test.rpy
@@ -24,7 +24,7 @@ label dev_idle_test:
     $ mas_idle_mailbox.send_idle_cb("dev_idle_test_cb")
 
     # return idle to notify event system to switch to idle
-    return "idle"
+    return {"idle": None}
 
 
 label dev_idle_test_cb:
--- a/Monika After Story/game/event-handler.rpy
+++ b/Monika After Story/game/event-handler.rpy
@@ -2597,7 +2597,7 @@ label call_next_event:
                     # return {"idle_exp": "5hubsu"}
                     if "idle_exp" in ret_items:
                         items = ret_items["idle_exp"]
-                        if not isinstance(items, (tuple, list)):
+                        if not isinstance(items, (tuple, list)) or (len(items) > 1 and isinstance(items[1], dict)):
                             items = (items,)
 
                         should_reset = True
--- a/Monika After Story/game/event-handler.rpy
+++ b/Monika After Story/game/event-handler.rpy
@@ -2616,7 +2616,7 @@ label call_next_event:
                                 exp_code = item[0]
 
                             else:
-                                exp_code = exp
+                                exp_code = item
                                 exp_kwargs = dict()
 
                             mas_moni_idle_disp.force_by_code(exp_code, clear=should_reset, redraw=should_reset, **exp_kwargs)
--- a/Monika After Story/game/event-handler.rpy
+++ b/Monika After Story/game/event-handler.rpy
@@ -2493,105 +2493,147 @@ label call_next_event:
         #if not seen_event(event_label):
         #    # give whatver the hourly rate is for unseens
         #    $ store.mas_xp._grant_xp(store.mas_xp.xp_rate)
+        python:
+            mas_RaiseShield_dlg()
 
-        $ mas_RaiseShield_dlg()
+            ev = mas_getEV(event_label)
 
-        $ ev = mas_getEV(event_label)
+            if (
+                notify
+                and (ev is None or ("skip alert" not in ev.rules))
+            ):
+                #Create a new notif
+                if renpy.windows:
+                    display_notif(m_name, mas_win_notif_quips, "Topic Alerts")
+                else:
+                    display_notif(m_name, mas_other_notif_quips, "Topic Alerts")
 
-        if (
-            notify
-            and (ev is None or ("skip alert" not in ev.rules))
-        ):
-            #Create a new notif
-            if renpy.windows:
-                $ display_notif(m_name, mas_win_notif_quips, "Topic Alerts")
-            else:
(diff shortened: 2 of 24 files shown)


ThePotatoGuy (changes requested):
So the majority of return keys only use a string, but we're changing the format for everything because of 1 return key's use case? And it went from typing a few characters (`"love"`) to double that for keys that don't need or care about a dict. 

I think we should keep return keys as simple as possible, and instead change how we tell the event-handler about `idle_exp`. Or if it even needs to notify event-handler because I'm not so sure it's even required if it's sole purpose is to force the next expression. For example, if the `idle_exp` flow was turned into a function, then could just call the function right before the topic ends - and you'll have way more control over input arguments and using kwargs.

Booplicate:
There are methods to set the next expression, but you can control it with the return key which is nice and handy. There's at least 2 systems that will use that in the future, with maybe more to come. Using `dict` gives more control: not only can we return a key, but also give it a value. It makes our system more advanced.

> Or if it even needs to notify event-handler because I'm not so sure it's even required if it's sole purpose is to force the next expression.

Same I can tell about the `"love"` key, we can use the function, but it's just nicer to leave it to ev-handler to manage.

ThePotatoGuy:
>Same I can tell about the "love" key, we can use the function, but it's just nicer to leave it to ev-handler to manage.

It was simpler for these cases because the functions it replaced were simple already.

>There are methods to set the next expression, but you can control it with the return key which is nice and handy

Do you mean the functions like `mas_moni_idle_disp.force(_exp)`? Because I would argue that this is easier to read:
```python
    $ mas_moni_idle_disp.force_by_code("1eua", duration=1) 
    return "unlock"
```
compared to this:
```python
    return {"unlock": None, "idle_exp": [("1eua", {duration=1})]} 
```
(if I am even forming this right, basing it off the commented example).

>Using dict gives more control: not only can we return a key, but also give it a value. It makes our system more advanced.

Making things more advanced is good, but we need to be careful on how much we're trying to shove into a generalized thing (returning <something> from a call stack). I feel like this is similar to another thing I argue with myself about often - should I have multiple functions that do 1 thing each, or 1 flexible function that can do multiple things. And I usually find the breaking point to be when the notation of the 1 function becomes complicated enough that it's like its own language. Like how `argparse` or format strings are insanely flexible but have a few pages worth of documentation on their syntax. Since we already have at least 2 mini-languages to deal with in python, I find it's not worth adding a third if I can help it.

Looking at the existing return key syntax and structure, I think we have a good balance right now. All someone has to read up on is the available keys - which luckily is still only a few - and the fact that we use pipes to delimit the keys. Another factor in their simplicity is how we don't have customization options for any of these keys. They all are meant for handling common/semi-common topic-based actions that contributors would likely need, but not need to customize (multiple functions that do 1 thing each). This is especially apparent looking at the current roster:
* `derandom` - sets random on the current ev to False
* `no_unlock` - sets unlock on the current ev to False, unlock_date to None
* `unlock` - sets unlock on the current ev to True, unlock_date to last_seen
* `rebuild_ev` - calls the `mas_rebuildEventLists` function - If I remember correctly, this was needed because some topics changed randomness of other topics and we don't update the random topic selection on a random change - could probably be handled automatically if we flagged something when `Event.setattr` gets called for random changes.
* `idle` - required for brbs since brbs don't have their own ev database 
* `love` - shorthand for the already short `mas_ILY` 
* `quit` - grandfathered in as the first return key and little reasons to remove since its pretty clear what it does
* `prompt` - go back to the prompt menu, highly useful for keeping menu-based topics encapsulated

So I don't think return keys is the right place for something designed to be flexible like `idle_exp`, especially if it pulls return keys toward a mini-language spec. I have a couple ideas on what could be done instead:
* calling the function directly - arguably the most straightforward approach since the functions by default should have a lot of flexibility anyway. Could even just make a single point of entry function that delegates to the others to make it even more flexible/generic/customizable. (ofc, keeping in mind over-generalization creep)
* make `dict-based` returns a optional return argument by tupling. something like `return "no_unlock|love", {"key": "value"}`.  Keeps the simple return keys for cases where they work best while allowing for slightly more complex options, as long as we avoid overcomplicating the 2nd arg. Although this is more for the future systems you mentioned, assuming those want to use just key/value customization and not object notation like `idle_exp` is doing.
    * alternatively, to support `idle_exp` here, make a customizer class that can be fed into the value - no messy object notation besides normal object creation, so it doesn't become its own mini-lang. Would have some overhead, but is a common tactic for OOP.
* mailboxes - although this would be making a new mailbox specific for the acceptable idle exp params. Way more overhead  than other options.
* Event properties - last resort, and I would probably veto this anyway, but highly useful if something is topic-based, does not/rarely change in runtime, and is extendable to basically every type of Event we have.

Booplicate:
After a discussion, we agree with you. Imo the closest variant would be to make a class to handle the value, but doing that just for the idle disp sounds wrong. Returning a `tuple` is also unclear, why in one place it's a `str`, in other it's a `tuple`, a random contributor might think "maybe I can put an X type too". It's either one or another to me. An event prop wouldn't work, idle exp can be different depending on the user choice and/or current affection (tbh I wouldn't go this way in general). And making a function call is just simpler than using a mailbox.

The main idea was to centralise idle disp handling in one place, since we expect many topic to use the new system in the future. But if the changes don't make it cleaner/faster/shorter, then what's the purpose? I think we'll stick to using return keys for something short, simple, and widespread, while using function/method calls for the remaining cases.

PR #6929, opened 2021-01-27
Added new game connect 4

AlmaHue:
Addition: "connect4" folder in "game" folder
Changes: On "zz_games.rpy" to accomodate


multimokia:
So I'm seeing a few issues in this.

Firstly, there's no displayable. I think a lot of this can be cleaned up with a Displayable class to represent the connect 4 board. I think we can also clean up the internal datatypes a bit, I'm seeing *a lot* of nested looping, a lot of which seems like it's a little excessive. We can probably id each section of the board so we can map out which areas have coins in them, using that we can also then use a simple binary value to store who owns each id. Ideally this form should help map things out (using dict). This should ideally help us only iterate over the values which matter (since we don't have to store every id, just the ones that are filled)

multimokia:
Also, PRs should be based to the `content` branch

AlmaHue:
@multimokia Ahh, I'll try to fix those then. This is my first time doing this, so I might have to learn some things first. And about the branch issue, thanks for the heads up!

Monika After Story/game/mod_assets/connect4/connect4.rpy
```diff
@@ -0,0 +1,655 @@
...
+            # prepare the board as rendererr
+            board = renpy.render(MASConnect4Displayable.BOARD_IMAGE, 1280, 720, st , at)
+            
+            # prepare winning tile highlight as renderer
+            win_tile = renpy.render(MASConnect4Displayable.WIN_TILE_IMAGE, 1280, 720, st , at)
+            
+            # prepare the pins as renderer
+            ai_pin = renpy.render(MASConnect4Displayable.AI_IMAGE, 1280, 720, st , at)
+            player_pin = renpy.render(MASConnect4Displayable.PLAYER_IMAGE, 1280, 720, st , at)
+
```
ThePotatoGuy, line 155: since it looks like these renders are only used in certain circumstances (like the game is over or 1 player is active), these should be moved to the lines where they are checked before being blit. Renders are more expensive than branch checks so the least we can do of these, the better.

Monika After Story/game/mod_assets/connect4/connect4.rpy
```diff
@@ -0,0 +1,655 @@
...
+        connect4_displayable_obj.show()
+        results = connect4_displayable_obj.game_loop()
+        connect4_displayable_obj.hide()
+        
+        quick_menu = True
+        
+        winner, ending = results
+        
+    label .connect4_end:
+        pass
```
ThePotatoGuy, line 593: probably don't need this label, i don't see it used anywhere else.

ThePotatoGuy (changes requested):

Monika After Story/game/mod_assets/connect4/connect4.rpy
AlmaHue: I didn't know that. I'll work on it.

Monika After Story/game/mod_assets/connect4/connect4.rpy
AlmaHue: I used to jump to that label from the console. Beyond that, it has no other use.

Booplicate (changes requested):
Please, move the files to the appropriate folders. `rpy` files go into the `/game` folder. We keep code there. Minigames' assets should be in `/game/mod_assets/games`.

AlmaHue:
@Booplicate I'm on it.

AlmaHue:
Done!

AlmaHue:
Fixed dialogue expressions to make her more lively (I didn't make the commit title descriptive enough so I put it here) (Also added "do you want to play again" loop)

Booplicate:
You have a lot of iteration here. You iterate through horizontal cells, then vertical ones, etc. Also some nested loops. That's not *really* an issue, but I was thinking if you could represent a cell as an object. Then each cell will be linked to each other. When a cell is captured, it'd update all adjacent cells too. That way you'd need to iterate through cells only once. Can get the same information (e.g. winning lines), but cheaper. Maybe some kind of double-linked list could work too.

AlmaHue:
> represent a cell as an object. Then each cell will be linked to each other. When a cell is captured, it'd update all adjacent cells too.

I don't really understand this part. How should that work exactly?

ThePotatoGuy:
I would argue looping over rows/columns/diags is fine as is, a double linked list is a lot of overhead for something that wont need more than 3-4 loops.  Remember that O(4n) is the same as O(n) in complexity terms.

Monika After Story/game/connect4.rpy
```diff
@@ -0,0 +1,690 @@
...
+        
+        def __init__(self, first_turn):
+            # prepare the Displayable
+            super(MASConnect4Displayable, self).__init__()
+            
+            # init the variables
+            self.is_game_over = False
+            self.quit_game = False
+            self.turn = first_turn
+            self.board = dict()
```
ThePotatoGuy, line 83: why is the board a dictionary? a connect 4 board makes more sense as a 2d array, or even just a 1d array with 2d indexing

if this is a 2d array. then we can do much better row/col/diag checks since we only have to deal with x, y coordinates and not the weird "line up" functions.

ThePotatoGuy (changes requested):

Monika After Story/game/connect4.rpy
AlmaHue: It _was_ a 2d array, but then [multi suggested to id the board using a dict.](https://github.com/Monika-After-Story/MonikaModDev/pull/6929#issuecomment-769311409) I think that was a good suggestion. Using a 2d array earlier was a whole lot of mess to iterate through.

Booplicate:
I am suggesting a more complex variant. But I'm not sure what adds more overhead here. There's more than 4 loops, and there's a few methods that do that. I guess the complexity is the same (kind of), but overall it's much more redundant iterations which could be avoided. Anyway whatever way you choose @AlmaHue, I was meaning smth like this
<details>
  <summary>Img</summary>
  
connect4

    
</details>

AlmaHue:
An...interesting solution... Looks like another month of work... Idk, I think I'll pass. I can find better use of my time than working on another variation of solution that already works. Except if it's really necessary, then I'll try working on that.

Monika After Story/game/connect4.rpy
ThePotatoGuy: This should be reverted back. A 2d array is understandable and easy to map to an actual board. A dictionary makes sense as an index, but when the most number of spots we have in this board is 42, the performance difference between a dictionary and a 2d array is negligible. 

And actually I've written C code as an assignment to play connect 4. It even uses a 1d array to represent the board, but regardless, it's not that complicated to do.

Iterate over rows/cols in a 2d array? Very easy - your code didn't do piece counting or size match checks but we can do that to make it even more efficient:
```c
int check_columns(char player_id){
    /*
     * For the columns, we only need to check from the bottom up until 
     * CHECKSIZE-1 from the top
     */

    /* controls */
    int col;    /* current column */
    int row;    /* current row */

    /* conditions */
    int has_more_pieces;    /* 1 when there are more pieces int eh column,
                               0 if not */

    /* containers */
    int matching;   /* how many matches we found */

    /* for every column */
    for(col = 1; col <= COLSIZE; col++){

        /* first check if anything is even in here */
        if(board[(ROWSIZE-1)*COLSIZE + col] != EMPTY){
            matching = 0;
            has_more_pieces = 1;
            row = ROWSIZE-1;

            /* while more pieces in the column */
            while(has_more_pieces && row >= 0){
//                printf("tset %i\n",matching);
                /* check the element, counting for CHECKSIZE match */
                if(board[row*COLSIZE + col] == EMPTY 
                        || (row < CHECKSIZE-1 && matching < 1)){
                    /* no more pieces or no more room for matches */
                    has_more_pieces = 0;
                }else if(board[row*COLSIZE + col] == player_id){
                    /* match found */
                    matching++;
                }else{
                    /* no match found */
                    matching = 0;
                }

                if(matching == CHECKSIZE){
                    return 1;
                }

                row--;
            } /* while more pieces in the column*/
        } /* check if anything is in the column */
    } /* for every column */

    return 0;
} /* check_columns */
```

Diagnols can be done with 4 separate loops in a similar way. The loops check:  left column negative slope, left column positive slope, first row neg slope, first row pos slope. This is an example of a loop for left col neg:
```c
    /* first, do negative slope scans */
    /* left column scans */
    /* for start spaces on left column */
    for(start_row = ROWSIZE-CHECKSIZE; start_row >= 0; start_row--){
        has_space = 1;
        matching = 0;
        row = start_row;
        col = 0;

        /* while more space in the diag */
        while(has_space && row < ROWSIZE){
//            printf("test %i\n", matching);

            /* check the element, counting for CHECKSIZE match */
            if(ROWSIZE-row < CHECKSIZE && matching < 1){
                /* no more room for matches */
                has_space = 0;
            }else if(board[row*COLSIZE + col] == player_id){
                /* match found */
                matching++;
            }else{
                /* no match found */
                matching = 0;
            }

            if(matching == CHECKSIZE){
                return 1;
            }

            row++;
            col++;
        } /* while more space in the diag */
    } /* for start spaces on left column */
```
Since this is my code, I could just rewrite it in python and we can use it here. Theres a few simplifications that can be made, too, like using generic bounds checking. The gist of all of these is to count consecutive pieces while avoiding unnecessary iterations by not checking areas smaller than the CHECKSIZE. CHECKSIZE being the number of pieces in a row to consider a win, which means this is also generic to any version of connect number. 

Additionally, if you just want xy tuples of diagnols for examination in a much better way: might be useful for the AI part, definitely **not** good to use with win checking since it would add a third loop.
```python
def generate_diag(self, start_x, start_y, positive_slope):
    # also assuming bottom left is (0,0)
    if not self.in_bounds(start_x, start_y):
        return []
    
    if positive_slope:
        slope_mod = 1
    else:
        slope_mod = -1

    x = start_x
    y = start_y
    xy_tups = []
    while True:
        x += 1
        y += slope_mod

        if self.in_bounds(x, y):
            xy_tups.append((x, y))
        else:
            return xy_tups

    return xy_tups
```

ThePotatoGuy:
Ooh I like the graph-based solution. Although I still think the doubly-linked list part is too much. A backing 2d array is better for at least the structure, especially with determining where to add a new piece when its has been dropped, and functions can be made for determining coordinates for an adjacent piece rather than doing reference looping. 

Additionally, as an aside, the py compiler should be able to optimize for loops, while reference looping will not get the same optimizations, so if we can have a structure without using ptrs then we should aim for that.

Either way, I prefer this graph-based solution + a 2d backing array over the current dictionary board implementation.

Monika After Story/game/connect4.rpy
AlmaHue: I hope I'm not misunderstanding what your code is doing, but I see that it only returns whether or not there is a match the size of `CHECKSIZE` and doesn't include which cells are those matches found in, while also doing tile checking from the start, even though getting the tiles from another function is much more reusable (which I didn't do in the earlier version that still uses 2d array).

In the current version I've worked on, four functions are used to get the tiles, which are then fed into both `check_connect` (a function used to check if there's a match) and `get_winning_line_up` (used to check if there's gonna be a win in one more move, used by the AI)
```
        # GET LINE UP FUNCTIONS START
        # These functions return a LIST of LINE UPS
        # a LINE UP is a LIST of the BOARD'S (KEY, VALUE) in tuple pair
        # didn't use dict() because it's unordered and the order needs to be preserved
        
        def get_vertical(self):
            """
            Returns vertical line ups
            """
            board_form = [[(self.g_k(j, i), self.board.get(self.g_k(j, i))) for j in range(self.CEL_ROW)] for i in range(self.CEL_COL)]
            return board_form
                    
        def get_horizontal(self):
            """
            Returns horizontal line ups
            """
            board_form = [[(self.g_k(i, j), self.board.get(self.g_k(i, j))) for j in range(self.CEL_COL)] for i in range(self.CEL_ROW)]
            return board_form
                
        def get_diagonal_ltr(self):
            """
            Returns diagonal top Left To bottom Right (ltr) line ups
            """
            board_form = list()
            for i in range(4):
                line_up = list()
                line_up2 = list()
                i2 = 0
                for j in range(6):
                    line_up.append((self.g_k(i+i2, j), self.board.get(self.g_k(i+i2, j))))
                    line_up2.append((self.g_k(j, i+i2), self.board.get(self.g_k(j, i+i2))))
                    i2 += 1
                if i != 0:
                    board_form.append(line_up)
                board_form.append(line_up2)
            return board_form
            
        def get_diagonal_rtl(self):
            """
            Returns diagonal top Right To bottom Left (rtl) line ups
            """
            board_form = list()
            for i in range(6):
                line_up = list()
                i2 = list(range(3, 9))
                for j in range(9):
                    line_up.append((self.g_k(i2[i]-j, j), self.board.get(self.g_k(i2[i]-j, j))))
                board_form.append(line_up)
            return board_form
                
        # GET LINE UP FUNCTIONS END
```
the `generate_diag` function you suggested, on the other hand, I could see that the current diagonal functions can use it. I admit my method was quite hacky and I have been having difficulty finding a clean solution for that part.

ThePotatoGuy: Since we're using python, this part can easily be changed to return the position as well:
```c
            if(matching == CHECKSIZE){
                return 1;
            }
```

Regardless, this doesn't change the fact that a 2d backing array should be the primary board representation. Getting xy tuples for columns/rows is just as trivial as it is for diagonals. But we should use boop's suggestion too as it makes both win checking and getting potential winning moves extremely trivial.

AlmaHue: Ahh, that is correct. I see it now.

So, regarding the reversion back to using 2d array, I changed it to dict from @multimokia's suggestion and I actually quite like it. In any case, the two of you have conflicting suggestions... I think it would be better to discuss this with @multimokia as well

AlmaHue: I...think I have a pretty clear grasp on it. I mean, I did make it 2d in the first place. Just gonna have to redo the optimization I've worked on the dict version back into the 2d method. Might take a while tho.

So, are we still gonna do it? Just asking for your official green light before I commit myself to work on it.

ThePotatoGuy: 2D array for sure. Boop's suggestion is optional.

AlmaHue: Alright, let me see what I can do.

ThePotatoGuy:
@AlmaHue any updates on this?

AlmaHue:
@ThePotatoGuy Ideally, the AI should be able to formulate planned moves instead of making random moves, but I don't know where to even start working with that.

Other than that, it is a completely playable working version.

ThePotatoGuy:
@AlmaHue So it works but Monika's playing is bad, correct?

AlmaHue:
@ThePotatoGuy Yeap.

Booplicate:
Hmm, we should probably get her w/r. Some people can't win at chess and even pong, so this might be a good game for them. But if it's too easy, people might find it boring.

AlmaHue:
I finally got the time to refactor the innerworkings of the game and implemented the AI as well, but due to how the AI works, the way Monika plays is very deterministic atm. Could probably use some tweaking to the AI so that she plays more naturally.

multimokia:
Are you able to merge in upstream content? For some reason your PR has a bunch of other stuff in it. Ideally it gets removed with that?

AlmaHue:
I made some wrong commits, but I already reverted them (I had to refamiliarize myself with github, sorry about that)

multimokia:
I think you reverted too far, looks like some upstream changes were reverted too. You might need to take your changes out, reset your branch and force push them back over, the branch is pretty broken

AlmaHue:
Yikes...I didnt realize that... I'll work on that next then

AlmaHue:
I believe I did it correctly this time. Pretty sure I reforked all the scuffed branches now.

multimokia:
Looks like +/- 0 lines, commits seem to be gone

AlmaHue:
Is there still something wrong?

multimokia:
image

This means your branch has no changes in it vs `content`

AlmaHue:
...I think the PR is automatically closed because I force pushed to reset my fork, and because of that the changes I've made haven't gone through. I don't know how to fix the PR, but I already have my changes in my fork. Should I just make a new PR instead...?

Booplicate:
Opened it, looks like you still have changes that don't belong to this PR.

AlmaHue:
Don't belong in the PR? So what should I do about that...?

Booplicate:
From 14 commits, only 1 is yours. I'm not sure how you did this, but maybe you could rebase your commit or use cherry-pick.

AlmaHue:
I [reset my branch to be the exact same as content](https://stackoverflow.com/questions/9646167/clean-up-a-fork-and-restart-it-from-the-upstream), put back only the changes I made (I already backed them up beforehand), and then commit and push that. I think the reset removed all the previous commits in my branch.

Issue #5371, opened 2020-02-23
grouping in the selector screen


ThePotatoGuy:
some ACS (ribbons) have too many options. All selectors are organized alphabetically but there might be better ways to do this:

### add collapsible/expandable groupings based on subtype/`acs_type`
* since selectors are filtered by group, grouping by `acs_type` would work out fine
* would need to dynamically hide/show displayables (probably just via a branch), but would need to evaluate potential performance issues
* restoring/canceling should close/reopen a grouping as well

### color groupings (from [this comment](https://github.com/Monika-After-Story/MonikaModDev/issues/5334))
* again, grouping by `acs_type` but iterating over color per grouping
* need buttons (could use the `MASClickzone`) on the sides of a selector
* restoring/canceling should scroll back if needed
* would be a UX nightmare if a specific type has a ton of colors

more ideas will be listed here...

ThePotatoGuy:
also search function maybe?
* text field, needs to be focusable/blurrable
* probably search as you type for ideal behavior. sorry future self.

ohnoitsavon:
I believe I have a solution to some of this.

ThePotatoGuy:
[closing as implemented]

PR #6930, opened 2021-01-28
Menu option to allow selection of acs_type - adds #5371

ohnoitsavon:
**adds #5371**

This adds a menu which filters ribbon selection based on acs_type.

screenshot_acs_menu

**Modified functions:**

_monika_ribbon_select:_
- Added code to create a list of acs_types and pass this onto the sidebar select function.
- Moved Monika to the left to allow space for the menu.

_mas_selector_sidebar_select_acs:_
- Accepts a new variable acs_list, this holds the list of acs_types.
mas_selector_sidebar_select:
- Accepts a new variable acs_list, this holds the list of acs_types. 
- A new variable acs_type is created, this holds the currently selected acs_type.

_mas_selector_sidebar:_
- Accepts a new variable acs_list, this holds the list of acs_types.
- The sidebar now only shows items which match the acs_type. 
- If acs_type is None all items are shown.
- The new menu is drawn here displaying a button for each of the acs_types in acs_list.
- When a type is selected, acs_type is updated. The screen is updated refiltering the results.
- Restoring the selection resets the acs_type to Null.

**New functions:** 

_mas_acs_name_format:_
- Takes in a string intended to be in the current format of how the acs_types are stored.
- Removes “-“, adds spaces, adds a capital to the first letter, and pluralizes.
- Returns formatted string.

_mas_plularize:_
- A basic pluralizer used in mas_acs_name_format.
- Allows exceptions to be added along with general rules.


ThePotatoGuy:
pushing this to next rel because of time.

Monika After Story/game/zz_selector.rpy
```diff
@@ -3680,7 +3734,19 @@ label monika_ribbon_select:
...
+        # generate acs_type_list
+        acs_type_list = []
+        for item in use_acs:
+            if ( item.get_sprobj().acs_type not in acs_type_list ) and (
+                    store.mas_sprites.is_hairacs_compatible(
+                        monika_chr.hair,
+                        item.get_sprobj()
+                    )
+                ):
+                    acs_type_list.append(item.get_sprobj().acs_type)
```
ThePotatoGuy, line 3746: Looking at this, I think there might be a way we can do this where we can have more control over the button names and what elements are part of each category. 

If we use a dict with the key as the button name and the value as a list of selectables associated wtih that key, then in the screen we can check if a button is selected and only show the selectables for that key. (and if no buttons are selected, show the full item list). We can pass this dict into the selector labels.

For generic ACS type cases (like this case, for example), we can have a function in the `mas_selspr` store that does the mapping based on ACS type, and another function for adjusting the keys so they are appropriate for visuals.

This way would give us more flexibility in which buttons to show and which selectables to categorize under a button, (including the option of having a selectable be under multiple buttons), which will help with making this work with hair/clothes selectors.

ThePotatoGuy (changes requested):

Monika After Story/game/definitions.rpy
```diff
@@ -6485,6 +6485,100 @@ init 2 python:
             return "An" if should_capitalize else "an"
         return "A" if should_capitalize else "a"
 
+    
+    def mas_plularize(ref_str):
```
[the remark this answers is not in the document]
ohnoitsavon: I have added a new commit that hopefully addresses this

Monika After Story/game/zz_selector.rpy
```diff
@@ -2912,11 +2909,57 @@ style mas_selector_sidebar_vbar:
...
-screen mas_selector_sidebar(items, mailbox, confirm, cancel, restore, remover=None):
+#   acs_list - the list of acs_types to be shown in the menu. Can be None
+screen mas_selector_sidebar(items, mailbox, confirm, cancel, restore, remover=None, acs_list=None):
     zorder 50
-#    modal True
 
     $ sel_frame_vsize = mailbox.read_frame_vsize()
+            
+    # only add menu if acs_list is provided and has more than one option
+    if acs_list and len(acs_list) > 1:
```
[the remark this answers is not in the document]
ohnoitsavon: I have added a new commit that hopefully addresses this

Monika After Story/game/zz_selector.rpy
ohnoitsavon: I have added a new commit that hopefully addresses this

ohnoitsavon:
I have added a new commit that hopefully addresses the changes. I have also seen that it would be possible to add filter options for hair and clothes using ex_props, but that would rely on the addition of AOC and the different categories.

Monika After Story/game/zz_selector.rpy
```diff
@@ -1771,6 +1769,30 @@ init -10 python in mas_selspr:
             scr.scope["flt_items"] = flt_items if flt_items is not None else scr.scope["items"]
         # Update the screen
         renpy.restart_interaction()
+    def mas_item_name_format(item_name):
```
ThePotatoGuy, line 1772: two spaces between functions when in a module

```suggestion

    
    def mas_item_name_format(item_name):
```

Monika After Story/game/zz_selector.rpy
```diff
@@ -3346,8 +3469,14 @@ label mas_selector_sidebar_select(items, select_type, preview_selections=True, o
 
         # setup prev line
         prev_line = ""
+    
+        # setup filtering
+        item_type = None
+        show_filter = False
```
ThePotatoGuy, line 3475: I would rather we not have global variables affecting the screen. If you need something to maintain state within screen iterations, then add additional props to the mailbox that is used here.

Monika After Story/game/zz_selector.rpy
```diff
@@ -3057,11 +3112,74 @@ style mas_selector_sidebar_vbar:
...
+        frame:   
+            area (960, 3, 50, 40)
+            background None           
+
+            button:
+                if show_filter: 
+                    style "filter_dropdown_down"
+                else:
+                    style "filter_dropdown_up"
+                action [If(show_filter == False, true=SetVariable('show_filter', True), false=SetVariable('show_filter', False)),Jump("mas_selector_sidebar_select_midloop")]
```
ThePotatoGuy, line 3133: midloop jumping is mainly for showing the dialogue box. The way screens work is that they refresh continuously, so setting the variable is enough to make the screen change next refresh.

ThePotatoGuy (changes requested):

Monika After Story/game/zz_selector.rpy
ohnoitsavon: Fixed in new commit.

Monika After Story/game/zz_selector.rpy
ohnoitsavon: Fixed in new commit.

ohnoitsavon:
I have added a new commit that addresses the changes.

Monika After Story/game/zz_selector.rpy
```diff
@@ -3243,7 +3247,7 @@ screen mas_selector_sidebar(items, mailbox, confirm, cancel, restore, remover=No
                             xalign 0.5
 
                     for selectable in flt_items:
-                        if (item_type is None) or (selectable.selectable in filter_map[item_type]):
+                        if (mailbox.item_type is None) or (selectable.selectable in filter_map[mailbox.item_type]):
```
ThePotatoGuy, line 3250: The `in` check here is expensive since its going thru a list.

Rather than this, if a filter is set, we should set a screen var with the selectables from the `filter_map`, then have the search code (`selector_search_callback`) adjusted so it checks that screen var first. 

However, this would also need to manually trigger search to re-run whenever the filter is selected since the search callback is only on input change. I'm not sure if we save the search_query anywhere (@Booplicate might know) but that would have to be saved somewhere (ideally mailbox) so we manually trigger search.

ThePotatoGuy (changes requested):

Monika After Story/game/zz_selector.rpy
Booplicate: I want to clarify, you suggest to append the selected filter option to the search query?

Booplicate: The search callback indeed can access the screen scope, so we can set/get a screen var (or mailbox var). Although, there's a few things I want to mention.

If the user searched, we shouldn't change the query. That's counterintuitive, plus we might not be able to remove that from the query later (say if the user removes/changes the filter).

While the search doesn't seem to cause any slowdown, it is kinda heavy. It checks name, id, type, and ex-props of **every provided item** (consider also all the callbacks it has to execute). Its flexibility comes at the cost of speed, I don't think we should rerun it on filter change. As of right now filtering seems to be more specific, but faster.

ThePotatoGuy: ok going to have some definitions here:
* button filter - the filter buttons (aka the buttons that map to a list of selectables)
* search filter - the filtering that occurs when a search is run

>If the user searched, we shouldn't change the query. That's counterintuitive, plus we might not be able to remove that from the query later

I did not mean change the search query, I mean save the search query to retrigger the search on the button-filtered set of items. 

Right now the way this works is the search code filters down `items`, and then the loop that adds the selectables to the screen does another loop per item to check if its in the list of selectables from the button filter.

A better solution would be that the search (which is the most expensive) runs on the limited set of selectables given by the button-filter (which is a direct mapping) if possible. But the problem is that if you ran a search prevoiusly, then use the button filter, the search code isn't going to re-run since its a callback on the input field.

pasting from @ohnoitsavon 's DM, (btw just respond on here for documentation purposes):
>i have worked out exactly what you meant now in terms of "screen variable" and can see how it would work, however it is not possible as the search sub functions all revolve arround "MASSelectableImageButtonDisplayable" and the items are not that, they are just the normal selectables

Yup, not possible at least how I was originally thinking about it. But another option could be:  
* in the pre-screen label (`mas_selector_sidebar`), loop over `filter_map` and replace selectables with their disp equivalent in items - (sort of expensive, but its a run-once so cheaper in the long run). This will be what is passed into the actual screen.
* still have a screen var for the subset of selectable displayables - search code should check for this var first
* when the button filter is selected do 2 things:
    * set the screen var to the list of selectable disps based on `filter_map`
    * loop over `flt_items` and remove ones that don't belong - I think the loop is probably faster than re-running the search, but the  key point is that its only ran when a button filter is selected instead of constantly via screen refresh.

ohnoitsavon: @Booplicate @ThePotatoGuy I think I can do something very close to that, I am looking into converting them in mas_selector_sidebar.

Booplicate: > A better solution would be that the search (which is the most expensive) runs on the limited set of selectables given by the button-filter (which is a direct mapping) if possible. But the problem is that if you ran a search prevoiusly, then use the button filter, the search code isn't going to re-run since its a callback on the input field.

Ah, okay that makes perfect sense now. Yeah, just save the query in the mailbox and call the search callback again as a button action.

> I am looking into converting them in mas_selector_sidebar.

Making displayables in the screen isn't ideal since that code will be: 1. rerun on every interaction, 2. rerun as we loop.

ThePotatoGuy: updated my comment - I meant `mas_selector_sidebar_select`, which is the pre-loop label

ohnoitsavon: New commit submitted

Monika After Story/game/zz_selector.rpy
```diff
@@ -3291,6 +3439,39 @@ label mas_selector_sidebar_select(items, select_type, preview_selections=True, o
...
+                            select_map,
+                            viewport_bounds,
+                            mailbox,
+                            False, # TODO: multi-select
+                            item.disable_type
+                        )
+                        for item in filter_map[filter_items]
+                    ]
+            filter_map = filter_map_selbtn
+
```
ThePotatoGuy, line 3474: this should use the same selectable displayables as the ones in `disp_items`. (So you should loop over the `disp_items` for each selectable in `filter_map` to get the displayables after `disp_items` is set) 

Otherwise there could be weirdness since this currently results in two displayables per selectable.

ThePotatoGuy (changes requested):

Monika After Story/game/zz_selector.rpy
ohnoitsavon: I have done a new commit that should fix this

ohnoitsavon:
This hopefully addresses the issues raised.



I have a question regarding the use of this in other menus. The way the code is setup allows any method of assigninh categories as long as the items are added to the cateogories, this section of code currently does that for the ribbons:

`# generate filter map
        mapping = {}
        for acs_sel in use_acs:
            acs_type_tc = mas_selspr.mas_item_name_format(acs_sel.get_sprobj().acs_type)
            if acs_type_tc in mapping:
                mapping[acs_type_tc].append(acs_sel)
            else:
                mapping[acs_type_tc] = [acs_sel]`
                
If elements are added for clothes/hair categories for example with the addition of AOC, these could be used to create the categories, or they could be made by other methods by checking for specific exprops. Are there any plans for this, or any suggestion of how this should be implemented?

ThePotatoGuy:
>Are there any plans for this, or any suggestion of how this should be implemented?

I'm not sure what you are asking here. The changes I asked you to make before were to make it so we can pass in a data structure like `key: category title, value: list of selectables for the category`. This alone means that if we wanted the categories to be exprops or anything else, it would be just a matter of sorting the data into this structure (aka, thats how future filter categories would be implemented).

ThePotatoGuy (approved):

ohnoitsavon:
@ThePotatoGuy Your last comment explains it, I was just asking if there are any plans soon to do that, you said "future filter categories" It is something I _might_ want to look at, once this has been fully tested/approved.

Monika After Story/game/zz_selector.rpy
```diff
@@ -1766,12 +1764,39 @@ init -10 python in mas_selspr:
...
+        IN: 
+            item_name - the text to be formatted
+        OUT:
+            item_name - formatted
+        """
+
+        # excpetions that include a "-" in the name
+        name_space_exceptions = {
+            "s-type-ribbon": "S-type Ribbon"
```
Booplicate, line 1787: I don't like that we will have to maintain this list in the future. Plus custom user acs types won't be in this list anyway. Maybe we should just show types as they are?

Booplicate (changes requested):
I noticed that the screen flicks when I change the filter. Is it possible to fix that? It doesn't happen when I interact in any other way (search bar, selecting an acs, restoring state, etc).

I'd also say we should support selecting multiple acs types at once, not just one.

Monika After Story/game/zz_selector.rpy
ohnoitsavon: Showing "-"'s in names would be very untidy, and removing it would make it "s type ribbon" which is wrong, @ThePotatoGuy can you make a decision on this? I think the better option is to add a new exprop for cateogory, anything without it would be uncatorgorised, also the name would be what is shown in the menu then.

Monika After Story/game/zz_selector.rpy
```diff
@@ -3319,6 +3467,18 @@ label mas_selector_sidebar_select(items, select_type, preview_selections=True, o
...
+        # make filter_map items MASSelectableImageButtonDisplayable
+        if filter_map:
+            filter_map_selbtn = {}
+            for d_item in disp_items:
+                for filter_items in filter_map:
+                    if d_item.selectable in filter_map[filter_items]:
+                        if filter_items in filter_map_selbtn:
+                            filter_map_selbtn[filter_items].append(d_item)
+                        else:
+                            filter_map_selbtn[filter_items] = [d_item]
```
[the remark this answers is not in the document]
ohnoitsavon: Sorry for the noob question but what do you mean?

Monika After Story/game/zz_selector.rpy
```diff
@@ -3057,12 +3120,87 @@ style mas_selector_sidebar_vbar:
...
+    else:
+        $ mailbox.item_type = None        
+        
+
+    if mailbox.item_type:
+        $ filtered_items = filter_map[mailbox.item_type]
+    else:
+        $ filtered_items = items
+    
+    default  flt_items = filtered_items
```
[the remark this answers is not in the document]
ohnoitsavon: What do you suggest instead?

ohnoitsavon:
> 
> 
> I noticed that the screen flicks when I change the filter. Is it possible to fix that? It doesn't happen when I interact in any other way (search bar, selecting an acs, restoring state, etc).
> 
> I'd also say we should support selecting multiple acs types at once, not just one.

Mine does not flick, also I would not know how to fix that. as I am just doing the same as when you select an outift to wear/when you search. Do you have any suggestion by looking at the code as to what is different?

I also do not know how it would be possible to add multiple options as I have not done that before in renpy, if you have any help or suggestion as to how that can be done let me know.

Monika After Story/game/zz_selector.rpy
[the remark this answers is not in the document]
Booplicate: Rename those vars so anyone who sees this code for the first time can distinguish that one of them is for search bar, another is for filter menu.

Monika After Story/game/zz_selector.rpy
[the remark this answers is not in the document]
Booplicate: Right now `filter_map_selbtn` is a `dict`. `defaultdict` accepts a factory as its argument which allows automatically fill missing keys.
E.g. `filter_map_selbtn = collections.defaultdict(list)`. When you access a key that's missing from `filter_map_selbtn`, it will make an empty `list` for that key. So you can just do `filter_map_selbtn["this_key_doesnt_exist"].append(item)` w/o doing that `in` check. [More examples](https://docs.python.org/3/library/collections.html#defaultdict-examples).

Monika After Story/game/zz_selector.rpy
Booplicate: I think both of those variants are better than having to remember that we should maintain a map that's local to a function scope.

ohnoitsavon: I will remove the exception list then, no worries for me, but i feel that somone needs to look at adding new property which could apply to clothes and hair too, as this code can work for that, but I feel that is a developer decision, as it would alter all the outifts but would allow beter sections and would be needed for adding support for @multimokia 's AOC/AHC

Monika After Story/game/zz_selector.rpy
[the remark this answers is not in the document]
ohnoitsavon: Those examples are what I was looking for thanks.

Monika After Story/game/zz_selector.rpy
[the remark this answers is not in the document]
ohnoitsavon: Thank you I undertsand what you mean now and will implement that

Booplicate:
> I also do not know how it would be possible to add multiple options as I have not done that before in renpy, if you have any help or suggestion as to how that can be done let me know.

Some time ago I made [`mas_check_scrollable_menu`](https://github.com/Monika-After-Story/MonikaModDev/blob/master/Monika%20After%20Story/game/screens.rpy#L2816), which has similar functional - selecting multiple buttons. Maybe it can help. I'd think that instead of filtering by just one type, we'd have a `set` of types we want to show.

ohnoitsavon:
> 
> 
> > I also do not know how it would be possible to add multiple options as I have not done that before in renpy, if you have any help or suggestion as to how that can be done let me know.
> 
> Some time ago I made [`mas_check_scrollable_menu`](https://github.com/Monika-After-Story/MonikaModDev/blob/master/Monika%20After%20Story/game/screens.rpy#L2816), which has similar functional - selecting multiple buttons. Maybe it can help. I'd think that instead of filtering by just one type, we'd have a `set` of types we want to show. Selecting a new filter in your menu, adds that type to the set, unselecting removes.

I will look at that, it should be easy then, but what about the flicker? I am not sure how that can be resolved, it is only for me when i dont have anything in the search, do you have any suggestion?

ThePotatoGuy:
I disagree on being able to select multiple at a time. I think the common use cases for these filters would be:
1. finding a specific ACS by narrowing down by category (but being too lazy to type to search)
2. wanting to see all the options for a specific category

Single select covers these use cases easily.

Multiple select seems more suited to showing combinations of things, but since we use the filters to separate visually different ACS, a sprite that is a combination of these ACS categories wouldn't be a thing.
Multiple select also means flipping through categories would require an extra click since u need deselect the previous to limit to the new category.

Monika After Story/game/zz_selector.rpy
ThePotatoGuy: We should keep the exception list. We can add API functions to add mapping exceptions to the list, so user-defined ones can add code to add their name exceptions.

Exprop isn't that great for this since it means every spritepack with that `acs_type` must include an additional element that is basically a copy of the `acs_type`.

multimokia:
>     2\. wanting to see all the options for a specific category

I'd actually argue that personally.

Using the options as categorical filters actually makes a lot of sense and is intuitive.

If you want to just refine down to xyz types from a larger pool, I think that helps a lot. Say if you wanted to exclude specifically s-types but see the rest to pick something. You've refined a much larger pool into a smaller one.

Not to mention, if you wanted to target one type specifically, it makes more sense to use search for that.

Additionally, search has no `not` filter, so regardless, if the amount of ribbons and other hair tie types continues to grow, the overarching problem still exists as even specific subcategories become large. As such being able to exclude them entirely makes more sense to mitigate it long-term.

ThePotatoGuy:
Single select already refines the larger pool into a smaller one so I don't see that as a talking point for multiple select. Same goes for "search has no not filter", single select is literally the most exclusive filter you can get without typing.

I think the best example here is the Sims 4 character creation screen. When you select a clothing item, you can pick subcategories, but they aren't combinable when clicking on them. They are separate and do a great job of narrowing the available options. If you want to see something from another category then you can switch to that category.

ohnoitsavon:
Well, I will make the code changes to fix the issues and look at any possible fix for the flicker, although but I do not know if it is possible. 

As for the categories why don't we have it as single for now and look at adding multiple if it is requested or shown as an issue? It also means this thing can adding and expansion into things like clothes and hair could be done? It would be amazing to have "short" "short tied" "long/down" "long tied" or something @ThePotatoGuy I assume that would be another or but is the same code I just don't know the categories / exprop if there is one that would show those values.

Monika After Story/game/zz_selector.rpy
ohnoitsavon: Ok I will keep the exception list

Booplicate:
> Multiple select seems more suited to showing combinations of things, but since we use the filters to separate visually different ACS, a sprite that is a combination of these ACS categories wouldn't be a thing.

Not sure, it feels kinda limited. Selecting multiple cats was the first thing I tried as a user, because it felt intuitive. Plus, I think it can be even more useful for clothes/hair selection in the future.

Monika After Story/game/zz_selector.rpy
ohnoitsavon: @ThePotatoGuy @Booplicate I can remove it or keep it, there is no issue either way, but it is only if a - is in the name, which is a very small amount of things. I don't know how it would be made into an API or moved scopes, please advise on this, but for now, I will leave it in and make the other changes, the dev team can then decides on that privately and come back to me with a decision on it.

ohnoitsavon:
> 
> 
> > Multiple select seems more suited to showing combinations of things, but since we use the filters to separate visually different ACS, a sprite that is a combination of these ACS categories wouldn't be a thing.
> 
> Not sure, it feels kinda limited. Selecting multiple cats was the first thing I tried as a user, because it felt intuitive. Plus, I think it can be even more useful for clothes/hair selection in the future.

For now I will leave it and make the other changes, I feel this needs to be a dev team decision, as I am ok with either and have a good idea how to implement it IF it is required, but personally am siding with @ThePotatoGuy at least for ACS's.

Monika After Story/game/zz_selector.rpy
ThePotatoGuy: @Booplicate 
We have maps for a ton of other things in the mod, another mapping is hardly that much more complicated.
However, I think we can actually do an exprop route if we added it to the ACSTemplate for s-type-ribbons. 
What do you think about that?

ThePotatoGuy:
@Booplicate @multimokia 
Even for clothes/hair filters (as exprops), I still think single select is more intuitive. Again, the sims 4 example can have clothes in multiple categories, but they still keep the subcats as single select.

Multiple select works for things like shopping for a product that has a ton of variations (for example, finding all computers that are under a certain amount, use a certain process, have a certain RAM size, etc...). With clothes/hair/ACS, however, we have a larger variety of products with little variation between the same products (hoodies vs bikins vs shirts, twin tails vs pony tails vs no tails, bow vs ribbon vs s-type ribbons (color is the main variation for ACS), etc...).

Monika After Story/game/zz_selector.rpy
Booplicate: I think that could work, but we don't have a template for clothes/hair.

Monika After Story/game/zz_selector.rpy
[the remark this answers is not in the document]
ohnoitsavon: This works, tested in my code, will post in the next commit.

Monika After Story/game/zz_selector.rpy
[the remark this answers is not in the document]
ohnoitsavon: This works, tested in my code, will post in the next commit.

Monika After Story/game/zz_selector.rpy
```diff
@@ -3863,7 +4039,16 @@ label monika_ribbon_select:
...
 
-        # make sure ot use ribbon for remover type
+        # generate filter map
+        mapping = {}
+        for acs_sel in use_acs:
+            acs_type_tc = mas_selspr.mas_item_name_format(acs_sel.get_sprobj().acs_type)
+            if acs_type_tc in mapping:
+                mapping[acs_type_tc].append(acs_sel)
+            else:
+                mapping[acs_type_tc] = [acs_sel]
```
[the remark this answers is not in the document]
ohnoitsavon: This works, tested in my code, will post in the next commit.

Monika After Story/game/zz_selector.rpy
```diff
@@ -3057,12 +3120,87 @@ style mas_selector_sidebar_vbar:
...
+                            textbutton _("Show All"):
+                                style "hkb_button"
+                                xysize (300, 40)
+                                xalign 0.8 
+                                action [SetField(mailbox,'item_type', None),SetField(mailbox,'item_type_old', current_item_type),SelectedIf(mailbox.item_type == None)]
+
+                            # Only need keys
+                            $ filter_map_sorted=sorted(filter_map.keys())   
+
+                            for item_type_name in filter_map_sorted:
```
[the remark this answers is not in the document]
ohnoitsavon: This works, tested in my code, will post in the next commit.

Monika After Story/game/zz_selector.rpy
```diff
@@ -3057,12 +3120,87 @@ style mas_selector_sidebar_vbar:
...
+#   filter_map - list of filter categories shown in the menu. Can be None
+screen mas_selector_sidebar(items, mailbox, confirm, cancel, restore, remover=None, filter_map=None):
     zorder 50
-#    modal True
 
     $ sel_frame_vsize = mailbox.read_frame_vsize()
-    default flt_items = items
+            
+    # only add menu if filter_map is provided and has more than one option
+    if filter_map and len(filter_map) > 1:
```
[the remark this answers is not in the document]
ohnoitsavon: This works, tested in my code, will post in the next commit.

Monika After Story/game/zz_selector.rpy
```diff
@@ -3057,12 +3120,87 @@ style mas_selector_sidebar_vbar:
...
+                        yfill False
+
+                        vbox:
+                            xsize 300
+                            spacing 5
+                            first_spacing 0
+                            null height 1
+                            
+                            # Hold current item type
+                            $ current_item_type = mailbox.item_type
```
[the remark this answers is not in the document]
ohnoitsavon: This will be added to the commit.

Monika After Story/game/zz_selector.rpy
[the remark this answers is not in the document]
ohnoitsavon: This works and will be added to the commit

ohnoitsavon:
I have addressed all changes in the latest commit including fixing the flashing screen following TPG's suggestions. I have left the name space for s-type code in for now, but await further guidance on that.

@ThePotatoGuy / @Booplicate What needs to be done for that?

Monika After Story/game/zz_selector.rpy
Booplicate: This hasn't been addressed yet.

ohnoitsavon: @Booplicate I was told by @ThePotatoGuy to leave it

ohnoitsavon: @Booplicate @ThePotatoGuy has the final say yes? so I would like clear action from them about what to do next

ohnoitsavon: I have been waiting almost 30 days for any kind of response to this, while I know this is not that important and people have lives to live, I feel we need an answer on this so we know what to do, ALL other chnages have been actioned.

Booplicate: I'm waiting for TPG response. But even if we kept the mapping, it shouldn't be local anyway.

ohnoitsavon: Ok, one option would be if we add something like a "cateogory name" exprop, that could be added TO ANYTHING including hair clothes etc. Meaning the search has another thing it can check, and the menus can work with anything?

Booplicate:
@ohnoitsavon I wrapped some lines to make it a bit more easier to read.

Also during testing I noticed that the `Filter` button and the current filter reset when you press the `Restore` button. Is that indented behavior? Because I think it's rather inconvenient.

ohnoitsavon:
> 
> 
> @ohnoitsavon I wrapped some lines to make it a bit more easier to read.
> 
> Also during testing I noticed that the `Filter` button and the current filter reset when you press the `Restore` button. Is that indented behavior? Because I think it's rather inconvenient.

Hmm... the wrap is obviously fine I am checking the game and the filter doesnt reset but the menu does close when restore is clicked, is that what you mean?

ohnoitsavon:
I have also noticed some other bugs when restoring the search gains an extra space? I will look into these changes, but feel there are some fixes need.

ohnoitsavon:
Fixed the issues mentioned above with the menu and search

Monika After Story/game/zz_selector.rpy
ThePotatoGuy: keep the mapping, move it so it's in the `mas_selspr` name space and we'll make adjustments in the future as necessary.

ohnoitsavon: I have added this in the new commit

ohnoitsavon:
I have fixed the issue with the mapping moving it to the new location @ThePotatoGuy

ThePotatoGuy (approved):

ohnoitsavon:
It says I have a requested change @Booplicate - what do I need to do? I though it was all sorted?

Booplicate:
I haven't requested anything. We've reviewed your code, now we're testing your additions.

ThePotatoGuy (approved):

ThePotatoGuy:
@Legendkiller21 has to test it so we have to wait for them.

Legendkiller21 (approved):

PR #6942, opened 2021-01-31
Fix F29 Annis

multimokia:
Fixed an issue in poor calculation of a comparison date for anniCount.

I've replaced the old calculation via `mas_utils.add_years`, this should ideally clean up the issues we were having so that it uses the proper day.

I have also replaced the specific store imports with a general `store` import, and prefixed the old imports. Please verify those parts still function normally.

## Testing:
- Verify anniCount returns the proper values normally (before/after/during the anni date)
- Verify that anniversaries on F29 do not crash when using anniCount to get the amount of annis.


ThePotatoGuy:
Since this is supposed to fix a feb 29 issue, lets high priority this @Legendkiller21 @Booplicate

Booplicate (approved):
Both `add_years` and `add_months` return appropriate dates for F29, `isAnni` gives expected return. And I couldn't reproduce the issue with the new changes.

Legendkiller21 (approved):

PR #6957, opened 2021-02-06
NYE data fix

ThePotatoGuy:
**NOTE:** making draft but I think this will have to be for next release. Too close for 0.12.0 also its not a big deal to wait a rel.

Currently need to double check when can we actually reset the vars and when should we not - an update script can run at any possible time and we need to make sure we don't screw with data at the wrong time.


jmwall24:
closing in favor of #8350

Issue #5026, opened 2019-11-20
Mod data transfer via Monika file


ThePotatoGuy:
making this issue to track this development. Planned since docking station introduction (#2066) but held off because of #2622 and similar. 

# planned features
The primary feature would be direct transfer of Monika via generated file to other instances of MAS.
This would remove most data from the source persistent and overwrite data in the destination persistent (except for specific reasons).

### on leaving a MAS install
* should all backup persistents be removed? Good for immersion, bad for corrupted monika file issues (#2622). Leaning to not removing them just in case.
* what data to clear off source persistent. Thinking history data, most dbs. Might be worth leaving some behind in for potential dialogue (see next section)
* some sort of identifying information should be left behind such that Monika would know if she came back to this pc. (probably generating a UUID for each monika?)

### on arriving in a new MAS install
* a fresh, unran MAS/DDLC install should try to load a Monika file if found, import the persistent data and load the game like normal. 
* an unran MAS install that was just installed over a ran DDLC install should probably have some sort of dialogue reaction and reaction. There is a Monika conflict in this case.
* a MAS install where Monika has been transferred off should also load a new Monika file without any issues. In this case, should we care if the Monika that was transferred off the same as the one being transferred in?
* a MAS install where Monika has left because of final farewell, should load the new Monika, but there really should be some sort of hesitation dialogue or something. The new Monika should obviously be aware of what happened in this install.
* a MAS install where Monika has been taken out via appropriate farewells should trigger some sort of dialogue respones before loading a new Monika. Potentially asking if its okay to stay here or something, or maybe just flat out reject taking another Monika's place. There is a Monika conflict in this case.

## Reservations
This was originally on hold because of the failed Monika file loading issues. Recently, there have been significantly less reports of this happening which means that the likely culprits of the data corruption issues (aff scrambling/pickled classes/temp dir file gen) have been weeded out over time. That said, file generation is still risky, so meta-solutions to avoiding disasters (like not deleting backups) should be included.

# Planning
Likely before 2021. Priority is low because of the complicated potential interactions.

lunulae:
Since I saw a recent issue discussing this, I figure I might as well toss in some thoughts and see what sticks!

I definitely think backup persistents ought to remain, just in case, especially while the system is new. When she begins preparing herself for leaving one computer, Monika should advise us on where to find these backups (maybe even moving them into the characters folder?) and to make sure that we take them along for safety.

[...]

PR #6986, opened 2021-02-12
Dissolve for idle disp and more

Booplicate:
I wanted to add dissolves for the idle disp from the get go, but didn't have time for it, so I'm doing it now. With these changes (and ideally with an r7 upgrade) all Monika's sprites will use a nice, smooth transition.

### Changes:
- `Dissolve` transition for Monika's idle displayable. The timings are: `0.25` sec for a pose change, `0.1` sec for an exp change.
- New displayables to handle blinking, winks, and tears. That's because the old ATL were clunky and had numerous bugs. The blinking timings are randomized now.
- `Dissolve` transitions for blinking, winks, and tears. The duration for all of them is `0.1` sec (for many of those we can't use more than that, although we could use different durations if we wanted to, leaving this to the testers).
- Some methods of `MASMoniIdleDisp` got the new `skip_dissolve` parameter. It allows us to skip the transition during the next exp change.
- Test cases for the new displayables.
- A displayable to handle following sprites. Supports right/left variants, has commented code for up/down variants too. Also utilises `Dissolve`s and is more optimized than the current `ConditionSwitch`. NOTE: uses a method from #7006.

### Fixes:
- Fixed a bug with incorrect render size of `MASMonikaRender`.
- Fixed a bug with the broken wink ATL (because I deprecated it lol).
- ~~We don't use `scene` in the `spaceroom` label anymore. RenPy has a bug with images being used in both `old_widget` and `new_widget` of a transform (`Dissolve` in this case). Since `scene` is hiding images, it adds them to `old_widget`, then we're showing them, it adds them to `new_widget`. Because of that the images are getting `st` and `at` for both states at once.~~

### Testing:
- Verify the spaceroom looks good (at startup, during idle, during different transitions).
- Check out the new test cases, you can use them for testing.
- Verify the transitions look good. Not too fast, not too slow, no visual bugs.
- Verify blinking, winks, and streaming tears look good both when used in the idle disp and when `show`en directly.
- Verify the transition from idle to a topic and from a topic back to idle (ideally w/ and w/o forcing an exp at the end) looks good (no seizures or weird movements).
- Verify the top of Monika's head doesn't disappear anymore.
- Verify the following sprites look good. Check the transition from a following to non-following and vice versa.


Monika After Story/game/script-ch30.rpy
```diff
@@ -816,10 +813,12 @@ label spaceroom(start_bg=None, hide_mask=None, hide_monika=False, dissolve_all=F
 
         day_mode = mas_current_background.isFltDay()
 
-    if scene_change:
-        scene black
```
ThePotatoGuy, line 820: the `scene` statement is an easy way to remove all displayables quickly. Much better than having explicit hide statements like the ones you ended up adding elsewhere. If a scene change occurs, we only need to consider what we want to show, not what needs to be hidden.

ThePotatoGuy (changes requested):

Monika After Story/game/script-ch30.rpy
Booplicate: Works perfectly now with #7036.

ThePotatoGuy (changes requested):
got this traceback while running but pretty sure its related to something else. 
```
While running game code:
  File "game/script-ch30.rpy", line 1513, in script
    $ mas_randchat.wait()
  File "game/script-ch30.rpy", line 1513, in <module>
    $ mas_randchat.wait()
  File "game/definitions.rpy", line 8273, in wait
    renpy.pause(WAITING_TIME, hard=True)
  File "game/sprite-chart.rpy", line 8205, in render
    img_render = renpy.render(self.current_img, width, height, self.current_st, at)
  File "game/sprite-chart.rpy", line 8753, in render
    mouse_pos = store.mas_windowutils.getMousePosRelative()
AttributeError: 'StoreModule' object has no attribute 'getMousePosRelative'
```

Booplicate:
> AttributeError: 'StoreModule' object has no attribute 'getMousePosRelative'

The PR that adds that method hasn't been merged yet. For testing I just chucked it into a temp rpy. This shouldn't be merged 'til we get #7006 though.

ThePotatoGuy (commented):
it seems to me that she blinks more often when advancing dialogue after a line completes. Also these blinks seem much faster than normal to the point where it feels like flickering. I'll hold off on a full approval until @multimokia and @jmwall24 take a look at the feel.

ThePotatoGuy (commented):
would these new changes potentially cause issues if we had two monika sprites on the screen at the same time?

Booplicate:
All Monikas would blink in sync which would look bad, but our current system doesn't allow to show 2 Monikas anyway. Are we going to do it in the future?

ThePotatoGuy:
its an option under consideration in #5026. but we could probably just force one to be static.

ThePotatoGuy (approved):

multimokia (approved):

jmwall24 (approved):

Legendkiller21 (approved):

PR #6995, opened 2021-02-15
Optimize conditionals

Booplicate:
We use `conditional`s quite often, and we're evaluating them once per minute. One thing we can do is compile `str` into py code to speed `eval` up.

### Changes:
- All `Event` objects compile their `conditional`s (if they aren't `None`) and add them to the conditional cache which can be accessed later on.
- Added new method to the `Event` class `checkConditional`. Allows to check `conditional` of a single `Event` object. Respects `None`, accepts `globals` and `locals`. Note: you can still use `eval` directly.
- Added new class method to the `Event` class `validateConditionals`. Validates `conditional`s of all events in `mas_all_ev_db`. Raises exceptions if there's an error in a `conditional`. We run it in `splashscreen`, after aff setup. Technically, we can use *any* functions/methods for `conditional`s now.

### Testing:
- Verify general stability.
- Add a `conditional` with an error to an event. Verify it crashes you on loading with the appropriate traceback (make sure it looks informative and not too confusing).
- Add/remove a `conditional` from an event. Verify it appears in the cache, adds to/gets removed from the event's props.

### DIFF
--- a/Monika After Story/game/definitions.rpy
+++ b/Monika After Story/game/definitions.rpy
@@ -519,6 +519,11 @@ python early:
         # NOTE: action code should be callable on a given event object
         ACTION_MAP = dict()
 
+        # Conditional cache
+        # A map from conditional string to compiled code objects
+        # (speeds up event checks)
+        _conditional_cache = dict()
+
         # NOTE: _eventlabel is required, its the key to this event
         # its also how we handle equality. also it cannot be None
         def __init__(self,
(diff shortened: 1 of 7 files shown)


Monika After Story/game/definitions.rpy
```diff
@@ -707,22 +712,52 @@ python early:
...
+                name - the name of the prop to change (str)
+                value - the new value
+            """
+            # Special handling for the conditional property
+            if (
+                name == "conditional"
+                and value is not None
+                and value not in Event._conditional_cache
+            ):
+                Event._conditional_cache[value] = renpy.python.py_compile(value, "eval")
```
ThePotatoGuy, line 759: `conditional` is not part of `N_EVENT_NAMES`, so we could move this to the `else` block of the that if statement and save a branch check. Although specifically you should move it to the `attr_loc` check after the `start_date` `end_date` checks.

Monika After Story/game/definitions.rpy
```diff
@@ -972,14 +1032,35 @@ python early:
...
+                if ev.conditional is not None:
+                    try:
+                        eval(cls._conditional_cache[ev.conditional], globals=renpy.store.__dict__, locals=renpy.store.__dict__)
+
+                    except Exception as e:
+                        raise EventException(
+                            "Failed to evaluate the '{0}' conditional for the event with the '{1}' label:\n{2}.".format(
+                                ev.conditional,
+                                ev.eventlabel,
+                                e
```
ThePotatoGuy, line 1053: for a better formatted exception, use the `traceback` module (`traceback.format_exc()`). See `zz_spritejsons` for examples.

ThePotatoGuy (changes requested):

Monika After Story/game/definitions.rpy
Booplicate: I tried that, and it didn't look informative to me. It just printed more lines which I didn't need (RenPy already prints stack quite nicely), I still looked at the exception in the end. But I'll change that so you can see which variant works better.

ThePotatoGuy (approved):

Legendkiller21 (approved):

PR #7079, opened 2021-03-06
Greetings can force exp

Booplicate:
### Concept:
By default the `spaceroom` label uses idle displayable for Monika. While it works for idle and most greetings, there are scenarios where we want to show a special expression for a greeting (so Monika doesn't stare at you for a moment and then pretends she didn't see you). One way to do it is to use `skip_visual`, but it requires quite big setup, which isn't ideal if all we want to do is to use a specific expression.

### Changes:
 - `MASGreetingRule` accepts a new param - `forced_exp` - the expression that will be shown during first spaceroom render. Default `None`.
 - `MASIdleMailbox` got 2 new methods: `send_forced_exp` and `get_forced_exp`.
 - Some greetings were updated to use the updated rule.
 - A few expression changes.

### Testing:
 - Verify general stability.


Monika After Story/game/script-ch30.rpy
```diff
@@ -1109,6 +1109,7 @@ label ch30_autoload:
 
     # set this to None for now
     $ selected_greeting = None
+    $ mas_forced_exp = None
```
ThePotatoGuy, line 1112: can you make this a mailbox entry instead of var? means we don't have to set None after.

ThePotatoGuy (changes requested):

Monika After Story/game/script-ch30.rpy
Booplicate: Sure.

ThePotatoGuy (approved):

jmwall24 (dismissed):

jmwall24 (approved):

multimokia (approved):

PR #7080, opened 2021-03-07
0.12.1 Fpyos

Booplicate:
### Fixes:
- A previous change to chess deleted the label that `mas_compliment_chess` checked in its conditional to unlock. Thus, any user that started the mod with v0.12.0 or later will not be able to unlock this compliment. To fix this, we now use a simple player loss check of greater than 5 as the conditional, which fixes the bug and is also a more logical check.

### Testing:
 - [X] Verify @Booplicate is a Fpyo
 - [X] Verify @multimokia is a Fpyo
 - [x] Verify no Fpyos.
 - [x] Test update scripts.
 - [x] Verify you get correct MAS window on Windows and Linux.
 - [x] Verify `getMousePosRelative` returns appropriate values.
 - [ ] ~~Test `monika_ribbon` while Monika's wearing an 8-bit ribbon (she should acknowledge she has a ribbon).~~
 - [x] Verify that if you call a scary story while in a custom bg, you stay in that bg and don't get teleported back to the spaceroom.
 - [x] Verify that the scary stories still behave like they should be if you call a scary story while in a custom bg or in a custom mask-less bg (no weather while in a mask-less bg, no out of place sprites etc.)


Monika After Story/game/script-topics.rpy
```diff
@@ -4670,7 +4670,8 @@ init 5 python:
     addEvent(Event(persistent.event_database,eventlabel="monika_ribbon",category=['monika'],prompt="Ribbons",random=True))
 
 label monika_ribbon:
-    if not monika_chr.is_wearing_acs_types("ribbon", "twin-ribbons", "s-type-ribbon"):
+    # TODO: We need a better handling for this
+    if not monika_chr.is_wearing_acs_types("ribbon", "twin-ribbons", "s-type-ribbon", "bow"):
```
ThePotatoGuy, line 4674: The retro bows shouldn't be "bow" type. They look like a standard ribbon so they should have been the ribbon acs-type.

ThePotatoGuy (changes requested):

Monika After Story/game/script-topics.rpy
Booplicate: Not sure why we set it to `bow`, but changing that now will mean people will have to update their spritepacks to apply the fix.

ThePotatoGuy: we can add bow here if multi's black bow should count as a ribbon, but otherwise we need to change the retro ribbons spritepacks regardless. the type and naming is wrong.

Booplicate: @multimokia decide.

But I'd change the spritepack in a new PR.

ThePotatoGuy (approved):

Booplicate:
### Note for us: FIX VERSION NUM

multimokia (approved):

Legendkiller21 (approved):

jmwall24 (approved):

PR #7101, opened 2021-03-14
Work greet

jmwall24:
Adds dlg describing why work was so amazing in `greeting_back_from_work`

PR #6864, opened 2021-01-03
Add left/right smug eyes

ThePotatoGuy:
# Key Changes
* adds left/right smug eye (from @Professor-JMO) expressions (m - smugleft, g - smugright) to exp previewer and sprite chart
* also rearranged `soft` so its consistent with the other sprite lists

# Testing
* verify new eyes look good

PR #7102, opened 2021-03-14
Jsonified Sprite Maps

multimokia:
Since #6864, we've had smugleft and right but not had them added to the sprite checker's map, causing Travis issues (#7101)

This PR addresses it by jsonifying the sprite maps and reading from the json within MAS and within the `StaticSprite` class in the spritechecker utility.

This way, we can add sprites by simply adding them to the map, and both MAS and the utils get updated together.

## Testing:
- Verify in-MAS sprites still work correctly with this system
- Verify spritechecker works with the new system and smug left/right works properly.

### DIFF
--- a/Monika After Story/game/sprite-decoder.rpy
+++ b/Monika After Story/game/sprite-decoder.rpy
@@ -50,6 +50,12 @@ init python in mas_sprite_decoder:
         TEAR_MAP = jobj["tears"]
         SWEAT_MAP = jobj["sweat"]
 
+        #Since tuples aren't supported in json, we need to do some conversion here
+        ARM_MAP["5"] = tuple(ARM_MAP["5"])
+
+        for side_key, side_list in SIDES_MAP.iteritems():
+            SIDES_MAP[side_key] = tuple(side_list)
+
     def __process_blush(spcode, index, export_dict, *prefixes):
         """
         Processes a blush off the given sprite code at the given index
--- a/Monika After Story/game/sprite-decoder.rpy
+++ b/Monika After Story/game/sprite-decoder.rpy
@@ -1,111 +1,54 @@
 #Runtime code equivalent of our spritemaker tool
 init python in mas_sprite_decoder:
+    import json
     import store
 
-    EYEBROW_MAP = {
-        "f": "furrowed",
-        "u": "up",
-        "k": "knit",
-        "s": "mid",
-        "t": "think",
-    }
-
-    EYE_MAP = {
-        "e": "normal",
-        "w": "wide",
-        "s": "sparkle",
-        "t": "smug",
-        "c": "crazy",
-        "r": "right",
-        "l": "left",
-        "h": "closedhappy",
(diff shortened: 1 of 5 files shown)


Monika After Story/game/sprite-decoder.rpy
```diff
@@ -1,111 +1,60 @@
...
+            jobj = json.load(jsonfile)
 
-    #NOTE: Everything not present here is assumed 3b
-    SINGLE_MAP = {
-        "a": "3a",
-        "u": "3a",
-    }
+        # is file json
+        if jobj is None:
+            store.mas_utils.writelog("[ERROR]: Failed to load sprite_map json.")
```
ThePotatoGuy, line 39: if this happens, we don't have a fallback right? We should just raise an exception.

ThePotatoGuy (changes requested):

Monika After Story/game/sprite-decoder.rpy
```diff
@@ -1,111 +1,60 @@
...
-    SIDES_MAP = {
-        "1": ("1l", "1r"),
-        "2": ("1l", "2r"),
-        "3": ("2l", "2r"),
-        "4": ("2l", "2r"),
-        "5": ("", ""),
-        "6": ("1l", "1r"),
-        "7": ("1l", "2r"),
-    }
+        with open(store.os.path.join(renpy.config.gamedir, "mod_assets", "sprite_map.json"), "r") as jsonfile:
```
Booplicate, line 34: I think `with` doesn't handle exceptions. So if the file is empty/corrupted or doesn't exist at all, we'd just crash here, never reaching the `None` check below. If you want to use that custom exception, you'd probably have to use `try/except` here.

Monika After Story/game/sprite-decoder.rpy
```diff
@@ -1,111 +1,60 @@
...
+        if jobj is None:
+            raise Exception("[ERROR]: Failed to load sprite_map json.")
+            return
 
-    BLUSH_MAP = {
-        "bl": "lines",
-        "bs": "shade",
-        "bf": "full"
-    }
+        EYEBROW_MAP = jobj["eyebrows"]
```
Booplicate, line 42: And maybe these should be in a `try/except` block too, since direct access of fields?

tools/sprite.py
```diff
@@ -967,3 +886,28 @@ def __swap_eyes(self, eyecode):
         instead.
         """
         return "".join([self.spcode[0], eyecode, self.spcode[2:]])
+
+    @staticmethod
+    def _loadSpriteMapData():
```
Booplicate, line 891: Same applies to this function since they are identical.

Booplicate (changes requested):

ThePotatoGuy (approved):

Legendkiller21 (approved):

Booplicate (approved):

PR #7117, opened 2021-03-23
[Consumable Dialogue] - Hanami Dango 🍡

lunulae:
This PR contains gift react dialogue + sprite for the dango consumable. 

I'm still feeling a bit uncertain on using the consumable code, so I sort of used the existing coffee gift react as a template.


multimokia:
FYI I think the consumable code is missing from this PR

Wingdinggaster656:
_If you didn't know, these are a street food in Japan. This variant is especially popular during cherry blossom viewings._
This is pretty strange. You are the one who sent this gift, and you don't even know what you are sending...?

lunulae:
@Wingdinggaster656 
That's fair. I had her explain it on first gifting since I didn't know if all players might look it up before gifting, and sometimes I've picked up a snack or sweet without really knowing its origin. Likewise, quite a bit of random dialogue prefaces similarly, "have you heard of this" or "did you know" falls into her speech patterns naturally.
If we think it's a bit much dialogue-wise I can try to come up with something else to say about them, though!

Wingdinggaster656:
> @Wingdinggaster656
> That's fair. I had her explain it on first gifting since I didn't know if all players might look it up before gifting, and sometimes I've picked up a snack or sweet without really knowing its origin. Also, quite a bit of random dialogue prefaces similarly, "have you heard of this" or "did you know" falls into her speech patterns naturally.
> If we think it's a bit much dialogue-wise I can try to come up with something else to say about them, though!

Agree but not _totally_ agree with you. I think the "don't know what you're giving, but you're giving it anyway" situation does exist, but after all, this situation itself is not very reasonable.
Let's find a better way to introduce it, like, "Speaking of this food, I have some interesting details to add."

KansaiGal:
Hey um... hanami isn't WRONG but... generally these are called "sanshoku" dango, at least on sales packaging.  Similarly, google reflects this:
"三色団子"　sanshoku dango
About 323,000 results (0.45 seconds)  [without quotes: About 18,600,000 results (0.46 seconds) ]
"花見団子" hanami dango 
About 115,000 results (0.43 seconds) [without quotes: About 2,410,000 results (0.62 seconds) ]

lunulae:
That's a good catch, I've only heard them referred to as hanami before. I will likely update the name on these later.

PR #7118, opened 2021-03-23
Full Name Topic

### LINKED ISSUE #5687: Tell Monika your full name
I'm currently working on a topic for telling her your middle name, last name, and suffix. Suffix wise I have options for Jr, Sr, and entering a regnal number. Considering adding prefixes too but might have way too many options there lol. I also created a function to return your full name so any topics that want to use it can access it easier.

Any thoughts are welcome.

MoniPython:
Reposted because I broke the old pull.

Adds a new random topic for her to ask your middle name, last name, and suffix. I felt it should unlock at higher affection because it can only be seen once and to keep it from getting too complicated lol. Also adds a topic for changing your full name and a function that returns your full name. I'm unsure about much of the dialogue though. Hope I didn’t add too few comments.

fixes #5687


TRIDENT1313:
What about for someone like me, who often goes by their middle name.
Could there be a way to change the player's first name as well?

lunulae:
This is a good point, I'm currently having her call me by a shortened form of my full first name. Having her check might be a good idea.

MoniPython:
Makes sense. I can expand it to have her ask if the player name is your given name and add another variable for the formal first name. If the devs don’t feel it will cause any conflicts I’ll get to work on it.

SpaceyFlakes:
On the topic of given names (super fucking late to the party, sorry) could there be an option where Monika asks if the name you give her when telling her your full name is the name you wish to be called if it doesn't match the one she calls you? Or, like, she could ask if you've entrusted her with your deadname if you've stated that you're trans/non-binary? I suck at explaining so please let me know if you'd like me to elaborate

MoniPython:
The way I have it set up now if you tell her you have a different given/forename she'll say she assumes you still want to be called by your [player] name, so [player] won't be changed by this topic. There already is a topic to change the [player] name so between those two functionalities I think that should cover telling her what your formally called and what you want to be called. But I don't mind adding more options if others feel that isn't covering it enough.

MoniPython:
Sorry this took so long.

jmwall24:
Closing as this PR is just too long and complicated to achieve its goal. If we determine we want to include last names in the future, there are much simpler ways to go about it.

Issue #6358, opened 2020-09-19
hotkey buttons should be disabled on bg switch


ThePotatoGuy:
when switching bgs (use the custom room pack), the music button is enabled and can be clicked during the transition. this shouldn't break anything, but we should make sure it doesnt by disabling the button during the transition (as well as all UI elements for safety).

ThePotatoGuy:
[closing as fixed]

PR #7127, opened 2021-03-27
Add shields when changing background

ThePotatoGuy:
#6358 

# Key Changes
* UI elements are disabled when background is changed
* hotkey buttons are hidden as well

# Testing
* Verify that the hotkey buttons/escape menu/calendar/etc is disabled when the background is changed via `Hey Monika > Location > Can we go somewhere else?`  
    * the buttons are only disabled once Monika says "Alright!"

### DIFF
--- a/Monika After Story/game/zz_backgrounds.rpy
+++ b/Monika After Story/game/zz_backgrounds.rpy
@@ -3216,7 +3216,13 @@ label monika_change_background_loop:
         skip_transition = mas_background.EXP_SKIP_TRANSITION in sel_background.ex_props
         skip_outro = mas_background.EXP_SKIP_OUTRO in sel_background.ex_props
 
+    # UI shields
+    $ mas_RaiseShield_core()
+
     call mas_background_change(sel_background, skip_leadin=skip_leadin, skip_outro=skip_outro, set_persistent=True)
+
+    $ mas_DropShield_core()
+
     return
 
 #Generic background changing label, can be used if we wanted a sort of story related change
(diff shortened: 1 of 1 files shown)


multimokia:
Maybe for looks we can just hide them during the transition?

Would probably feel a little more cohesive to have the black screen w/o the buttons remaining.

ThePotatoGuy:
added button hiding in the same spot where the shields are done.

Monika After Story/game/zz_backgrounds.rpy
```diff
@@ -3216,7 +3216,17 @@ label monika_change_background_loop:
         skip_transition = mas_background.EXP_SKIP_TRANSITION in sel_background.ex_props
         skip_outro = mas_background.EXP_SKIP_OUTRO in sel_background.ex_props
 
+    # UI shields + buttons
+    # NOTE: buttons are in here since there is no consistency if placed in 
+    # the bg change label.
+    $ mas_RaiseShield_core()
+    $ HKBHideButtons()
+
     call mas_background_change(sel_background, skip_leadin=skip_leadin, skip_outro=skip_outro, set_persistent=True)
```
Legendkiller21, line 3225:
```suggestion
    call mas_background_change(sel_background, skip_leadin=skip_leadin, skip_transition=skip_transition, skip_outro=skip_outro, set_persistent=True)
```
We set `skip_transition` a few lines before but we never pass it in `mas_background_change`. We should probably do so just in case a bg needs it in the future.

Legendkiller21 (changes requested):

Legendkiller21 (approved):

multimokia (approved):

PR #7136, opened 2021-03-29
Mirror stdout + stderr to file

multimokia:
Even with [`0e848db`](https://github.com/Monika-After-Story/MonikaModDev/blob/0e848dbdaace268e8e381b2521c557f4f414a255/.github/workflows/mas_check.yml), we still didn't properly report errors in the GHActions script

~~using `tee` I mirrored the output from lint to stdout and to file so it's reported properly.~~

I modified the script to no longer use the lint parser, but instead run through `grep` with an inverted match criteria to find the lines referring to images

### DIFF
--- a/.github/workflows/mas_check.yml
+++ b/.github/workflows/mas_check.yml
@@ -65,7 +65,7 @@ jobs:
       - name: rpy lint
         run: |
           cd renpy
-          ./renpy.sh "../mas0105/" lint | grep -E -v "^$|Could not find image \(monika [0-9][^[:space:]]+ corresponding to attributes on say statement\.|'monika [0-9][^[:space:]]' is not an image\.|The image named 'monika [0-9][^[:space:]]+ was not declared\."
+          ./renpy.sh "../mas0105/" lint | grep -E -v "^$|Could not find image \(monika [0-9][^[:space:]]+ corresponding to attributes on say statement\.|'monika [0-9][^[:space:]]+' is not an image\.|The image named 'monika [0-9][^[:space:]]+ was not declared\."
 
       # distribute
       - name: rpy distribute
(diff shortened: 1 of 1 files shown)


multimokia:
Hmm we need a way of fixing that issue of all the missing images again

.github/workflows/mas_check.yml
```diff
@@ -65,13 +65,10 @@ jobs:
       - name: rpy lint
         run: |
           cd renpy
-          ./renpy.sh "../mas0105/" lint &> renpy_output
-          python ../tools/renpy_lint_parser.py
-          cat renpy_output_clean
+          ./renpy.sh "../mas0105/" lint | grep -E -v "^$|Could not find image \(monika [0-9][^[:space:]]+ corresponding to attributes on say statement\.|'monika [0-9][^[:space:]]' is not an image\.|The image named 'monika [0-9][^[:space:]]+ was not declared\."
```
ThePotatoGuy, line 68: This regex is covers too many cases. The previous one was very specific in catching forms of `monika #ccc+` (where `c` is any sprite code), `#c`, and specific other numbers/images. This is so an actual invalid code like `monika 1ef` would still appear in the log.

ThePotatoGuy (changes requested):

.github/workflows/mas_check.yml
Booplicate: We're only hiding renpy lint reports, won't our sprite checker messages still pass through?

multimokia: Yeah our spritechecker messages should still be pushed through since this only checks for renpy lint output

ThePotatoGuy: that appears to be correct

ThePotatoGuy (approved):

ThePotatoGuy (approved):

PR #7145, opened 2021-04-02
Add tag replacement to Image Tag Decorations

ThePotatoGuy:
# Key Changes
* Adds using a different image tag for a decoration depending on background
* deco definitions per background now include a replacement tag in addition to the frame
* to use replacement tag functionality, pass in a `replace_tag` param to the `register_bg`/`register_img`/`register_img_same` functions
* The deco ADF test topic has been updated with an example of tag replacement in operation
* `mas_isDecoTagVisible` now means that the decoration (or replacment tag) is **actually showing**, not just when its enabled.
* fixed issue where showing/hiding deco without the `show_now` or `hide_now` params didn't show/hide the deco at all.

# Testing
* test the `TEST DECO ADF` dev topic


multimokia (dismissed):

Booplicate (commented):
You deprecated `mas_isDecoTagVisible`, but it's still being used in a few places. I'd probably even replace it with `mas_isDecoTagShowing` in `mas_d25_monika_christmas`.

Should we handle this now or later?

ThePotatoGuy:
Looking at the other spots its used, it seems like Visible should just mean Showing, so instead just changing the meaning of the function instead of deprecating.

Monika After Story/game/zz_backgrounds.rpy
```diff
@@ -2146,10 +2146,17 @@ init -10 python:
...
+                    real_tag, adv_frame = deco_info
+
+                deco = store.mas_deco.get_deco(real_tag)
                 if adv_frame is not None and deco is not None:
-                    self._deco_man._adv_add_deco(deco, adv_frame)
+                    self._deco_man._adv_add_deco(
+                        deco,
+                        adv_frame,
+                        override_tag=tag
+                    )
```
Booplicate, line 2159: I got an unbound error here, let's put it under the previous if check (github won't let me add a suggestion).
```renpy
            if tag is not None:
                deco_info = self.get_deco_info(tag)
                if deco_info is not None:
                    real_tag, adv_frame = deco_info

                    deco = store.mas_deco.get_deco(real_tag)
                    if adv_frame is not None and deco is not None:
                        self._deco_man._adv_add_deco(
                            deco,
                            adv_frame,
                            override_tag=tag
                        )
```

Monika After Story/game/script-ch30.rpy
```diff
@@ -913,7 +913,8 @@ label spaceroom(start_bg=None, hide_mask=None, hide_monika=False, dissolve_all=F
                 #     force_exp = "monika idle"
 
             if not renpy.showing(force_exp):
-                renpy.show(force_exp, tag="monika", at_list=[t11], zorder=MAS_MONIKA_Z)
+                # please keep i11 - otherwise monika jumps if we scene change without dissolving all
+                renpy.show(force_exp, tag="monika", at_list=[i11], zorder=MAS_MONIKA_Z)
```
Booplicate, line 917: With these changes Monika changes her position after the greeting, moves to the right. I think that happens because `i11` and `t11` actually set different positions. And it is noticeable.

Booplicate (approved):

Legendkiller21 (approved):

multimokia (approved):

PR #6915, opened 2021-01-22
New greeting, new clothes and hair, and hair sprite adjustments

Booplicate:
Main goal is to add a greeting which adds more life.

### Changes:
- Added a new greeting where Monika returns from shower or bath.
- Added new hairstyle (wet hair), new outfit (bath towel), and an "accessory" (water drops on Monika's body).  Thanks @kkrosie123 for helping me fix them!
- Added new static class `MASTimedeltaRepeatRule`. A rule for events we want to repeat only after certain amount of time has passed since last seen. It works for both farewells and greetings.
- ~~Deleted some old night sprites + removed white pixels from some hairstyles (def, ponytailbraid, twintails, usagi).~~
- ~~Slightly touched up the color of the hair sprites so they're less red~~
- ~~Added the hair spritepack in `Orcaramelo Sakuya` to the `Orca Hairs` folder~~
- Added several new function: `mas_sprites.get_installed_sprites` which takes a type constant and a predicate function, returns sprite objects. And more specific functions: `mas_sprites.get_installed_acs`, `mas_sprites.get_installed_hair`, and `mas_sprites.get_installed_clothes`.

### Testing:
- Verify the greetings can happen only once in 3 days.
- Verify the new hair, outfit, and accessory look good and work as expected. Note: those aren't selectable by user at least for now, since it doesn't make sense. We don't need weird combinations like the blazer + wet hair.
- Verify the cleanup event is being run appropriately. That's being after its `start_date` passed during game sesh and during the time user being absent.
- ~~Verify the adjusted hairstyles look good~~
- ~~Verify we removed all unused night sprites.~~
- ~~Verify adjusted hair looks good~~
- ~~Verify copied hair spritepack works properly.~~
- Verify the new `mas_sprites.get_installed_*` functions work.

PR #6729, opened 2020-11-25
Diamond Earrings Spritepack

20RaVen23:
Kaiewe and I made some earrings. I asked him about what name should be fine as author and that was okay with him.

- Added reaction
- JSON
 - [X] Add the needed sprites due some issues I was having, couldn't add them myself.

## Testing:
- [ ] Verify acs look okay
- [ ] Verify earrings selector works properly
- [ ] Verify reaction dialogue is fine

PR #7148, opened 2021-04-04
Split head sprite from body

ThePotatoGuy:
Needed by #6915 #6729 

# Key Changes
* Head is spilt from the body (via copy)
* New hair layer - mid - goes between body and head
    * JSON support for the new hair layer
    * the new hair layer is controlled with a MASPoseMap on the MASHair class. This means that mid can appear for only certain poses if desired.
* BAT ACS layer moved to right before table
* New ACS layers:
    * BMH - between Back arms and Mid Hair
    * MMH - between Mid Hair and Head

## bug fixes
* fixed BAT/MAT ACS layers not being saved to persistent

# TODO (optional)
If there are any hairs that would benefit from a mid layer split, let me know/perform the split and add it

# Testing
* Verify mid layer works
    * The `dev_mas_hair_test` hair object supports mid. There are corresponding hair files in `dev/dev_hair/`.
    * To adjust a JSON, add a `"mid_hair_map"` object with `mpm_type` set to 0 (and appropriate other lines). This is the same structure as `pose_map`.


Booplicate (approved):

ThePotatoGuy:
@Legendkiller21 just making sure this is on ur radar - this is high priority

Legendkiller21 (approved):

PR #7168, opened 2021-04-14
Fix IO Thread Crash

multimokia:
Was looking thru dockstat, noticed this func didn't get a logfile passed in, this will crash the thread and brick the promise.

Added the rd_log to the func call and it should work as intended now


ThePotatoGuy (approved):

Booplicate (approved):

multimokia:
TODO For self: Shift logging utilities to be made in a very early stage of init `00*.rpy`, `python early`

Issue #7208, opened 2021-05-13
[Suggestion] Monika Detecting Online Writing Software (Google Docs, Microsoft Word, etc.)


neo-kid:
I usually use Google Docs to write personal stories to post online. It'd be awesome if Monika could detect this and either say something like _Are you writing a story? You should show me sometime!_ or something about you writing a poem or doing homework.

I figure the mod team can come up with a more creative message then I can, but this is just a suggestion haha

20RaVen23:
Is it for dialogue or window reaction?

neo-kid:
> Is it for dialogue or window reaction?

Window reaction! prolly should've said in the post, oops

20RaVen23:
Aight, will be taking this under my roof if there's no problem with that.

neo-kid:
> Aight, will be taking this under my roof if there's no problem with that.

That'd be awesome!!

ThePotatoGuy:
[leaving open as in progress]

ThePotatoGuy:
[closing as implemented]

PR #7210, opened 2021-05-15
Google Docs/LibreOffice Writer/Microsoft Word window reaction

20RaVen23:
Adds Google Docs, LibreOffice Writer and Microsoft Word window reaction.

Suggestion from #7208


Monika After Story/game/script-windowreacts.rpy
```diff
@@ -562,3 +562,29 @@ label mas_wrs_twitch:
     if not wrs_success:
         $ mas_unlockFailedWRS('mas_wrs_twitch')
     return
+
+init 5 python:
+    addEvent(
+        Event(
+            persistent._mas_windowreacts_database,
+            eventlabel="mas_wrs_googledocs",
+            category='["google", "docs"]',
```
multimokia, line 571: Category should be a `list`, not a `string`

Additionally, this is using the old style windowreact query. We now use Regex to fetch the info (and it's all case sensitive now), meaning we can actually account for more word processors too. I suggest covering some more common ones like `LibreOffice Writer` and `Microsoft Word`. You can use the `|` operator in regex to specify an *or*. (So something like `"Google Docs|LibreOffice Writer|Microsoft Word"` would fit the category.)

multimokia (approved):

jmwall24 (approved):

PR #7216, opened 2021-05-16
Hangman

Iniquitatis:
Psst. Want some new (about a year old) hangman assets?

---
Screenshot
_(Source SVGs: [original PR](https://github.com/Iniquitatis/DDLCComfyUI/pull/8))_


ThePotatoGuy (changes requested):
I think the only problem with this is that the bg image for dark mode is the same white paper. I see why you made it that way, but that white rectangle is going to be annoying for dark mode users. Any way you can get a darker bg image?

lunulae:
Do you think putting the night filter on it might help? I think the white looks fine during the day even with the dark-mode UI, but I agree that during the night it might look quite bright.

Iniquitatis:
Done.

Iniquitatis:
This mini-game might be another instance where the Journal font (`m1_fixed.ttf`) is preferable.
Quick mockups:

Monika After Story/game/zz_hangman.rpy
```diff
@@ -406,7 +406,7 @@ label mas_hangman_game_preloop:
 
     # setup positions
     show monika at t21
-    if store.mas_globals.dark_mode:
+    if mas_current_background.isFltDay():
```
ThePotatoGuy, line 409: dark mode is the correct check here, we only change UI elements based on the dark mode setting.

ThePotatoGuy (changes requested):

Monika After Story/game/zz_hangman.rpy
Iniquitatis: Reverted.

ThePotatoGuy:
the journal font fits aesthetically but I think there could be problems with readability, especially with how cursive the font looks. Not to mention the lowercase 'j'. @multimokia any thoughts?

multimokia:
Honestly I'm kinda fine with journal, it does make stubs sense considering she's writing it out. Personally at least, I've not had issues reading it

Iniquitatis:
Pushed font changes as a separate commit, so it could be reverted.

Iniquitatis:
About the readability and accessibility: you might want to check [this](https://github.com/Iniquitatis/DDLCComfyUI/pull/4). It's been a long time since, but IIRC almost everything there passes the WCAG AAA.

(Seriously, the current dark theme causes an eye bleeding for me. And I don't think I'm the only one.)

ThePotatoGuy:
@Iniquitatis Make a separate PR for the dark mode changes and we'll evaluate it.

@multimokia 
My concern with the font change is because the lowercase `j` doesn't have the curve so it looks odd. For example:
journal_emoji

But I'm willing to let it pass for now and see what people say about it.

multimokia:
> @multimokia
> My concern with the font change is because the lowercase `j` doesn't have the curve so it looks odd. For example:

That's fair,, bear in mind I'm also in this with a semi-biased view given I run the Font Change submod which keeps all her text as Journal so I'm just used to it at this point lol.

Can always give it a trial run

Booplicate:
I think it's fine, but if needed, I can edit the font a little, I've done it before.

Iniquitatis:
@Booplicate Yeah, that's what I was going to ask about.

@ThePotatoGuy I don't quite remember how to do it, but I'll need to cheat a little, lol. (My save file is gone in the HDD crash accident.) Which variables should I change in the persistent to unlock all needed features to check the UI?

ThePotatoGuy:
@Booplicate I think editing the font is out of the question. Should maintain consistency with DDLC's journal font, so we either use it or not. 

@Iniquitatis I'm not sure what you are asking for. You can find all the UI sprites in the `mod_assets/` folder + subfolders.

Iniquitatis:
@ThePotatoGuy Nevermind, I've already found out.
(I've been asking about how to check the stuff like the "island event", because I needed to start the game from scratch.)

ThePotatoGuy (approved):

Legendkiller21 (approved):

PR #7231, opened 2021-05-22
Add letters and numbers on board

Wingdinggaster656:
Okay, since we are making a chess tutorial, it is urgent to add signs on board. So I made this:
The main question is that, the current code is supported to show chess_board.png only.
But the chess rule is, if a player is playing black, and then the board in this player's eyes should be like the 2.png I added(See commit "Add the board image when player is black").
I tried to change the code of chess.rpy, but I doubt it would ruin something related to the current code, so I am not daring to do that.
Now I am offering these images, hope you guys can help me to change the code.
By the way, this is the effect on screen now:
QQ截图20210522143953


Booplicate:
You only need one image, and then you can rotate it using `Transform`. I'd also say use only that dark colour, because the bright one is hard to see.

Wingdinggaster656:
> You only need one image, and then you can rotate it using `Transform`. I'd also say use only that dark colour, because the bright one is hard to see.

I am sorry, but I guess you probably don't know much about Chess, if we simply transform the new chess_board.png, then it's not like the 2.png represented. The numbers and letters on the board would be transformed too, and that's now what should happen.
As for use only dark color, I think it's okay, I will make another one later. Though most of chess boards on computer are having two colors of these signs, so I hope you can consider it again.

Booplicate:
My bad, you're right.

MrVavlo:
I don't get it, wouldn't adding mirrored numbers and letters to the opposite side work for `Transform` thing? Like this: image

Wingdinggaster656:
> I don't get it, wouldn't adding mirrored numbers and letters to the opposite side work for `Transform` thing? Like this: image

That's not the popular tend now. You may refer to the chess.com, lichess.org, chessease.net and even most of the APPs in stor are no longer doing that now.
As for chess.com is the biggest website nowadays, I think we should follow that website's setting. That's also the most popular setting now.

Booplicate:
Hmm I think it doesn't matter how "popular" it is. What's important is it gives the same information, but saves us one image on disk. Can you do that?

Wingdinggaster656:
> Hmm I think it doesn't matter how "popular" it is. What's important is it gives the same information, but saves us one image on disk. Can you do that?

Of course I can! It's just a little bit of a concern to me, I'm a chess player for a lot of years.I once read a psychology report that this second image that MrVavlo provided presents a lot more information, and it's going to have a very slight negative effect on people learning on the board, and I think it's going to be a little bit of a hazard.

But don't worry, I fully respect your decision, I will post all versions of the images for you to choose from later today, I don't see any harm in saving disk space, after all, I have been coding programs for many years too, so I know how important it is to save disk space. I WON'T SAY ANYTHING TO YOUR DECISION.

Wingdinggaster656:
@Booplicate 
Alright, I have made all versions of images here:
Firstly, the save-disk-space choice with two kind of colors(Not using only dark):


Secondly, the save-disk-space choice with only dark colors:


Thirdly and finally, the non-save-disk-space choice with only dark color:
One_color

ThePotatoGuy:
alternatively, if you make 2 pngs that are just the letter/numbers, we can just layer them on the board depending on the player's color

Wingdinggaster656:
> alternatively, if you make 2 pngs that are just the letter/numbers for each view, we can just layer them on the board depending on the player's color

I agree that would be the best, but I have no many ideas to make that. I don't really know how is the chess board displays now.
Maybe you guys can help on that?

PR #7235, opened 2021-05-22
Adding four new songs.

Retrolovania:
Added three new songs for Monika to sing, them being:

-Fly Me to the Moon (Frank Sinatra)
-On the Front Porch (Burl Ives)
-Heaven (Bryan Adams)


jmwall24 (changes requested):
Bit of an issue here with formatting for the first and third songs. You made these 'long' songs but we require all songs (that aren't special events) to have short versions and then it's an option for those to unlock the long (full) versions. Short songs have the short commentary after them like you have here, but the long versions should just full the full lyrics with no commentary. You need to pick shorter segments of the songs for the short versions and then use your bit of commentary that you have here at the end of those. Then if you want, you can add long versions which should just be the full lyrics. The second song is fine. Also, `return` is indented incorrectly for all 3 songs, it needs to be one indent in from `label` not in line with it.

Retrolovania:
I see. Very well. I'll get on 'em as soon as I can.

Retrolovania:
Any more edits needed, per chance?
Might add a fourth song at some point soon.

multimokia (dismissed):

jmwall24 (approved):

multimokia (approved):

PR #7254, opened 2021-05-28
Cats

Iniquitatis:
Do you name your cats properly? I certainly do.

---
_**Ddlc...** in the category list has bothered me from the start. It is an abbreviation, dammit._
Still, mapping category IDs to the human readable names (and not just capitalizing an ID) might be useful in the similar cases.


Iniquitatis:
Whatever. It just works.

jmwall24:
Going to tag @ThePotatoGuy first for this to make sure this is how we want to handle this

PR #7263, opened 2021-05-30
New Fun Fact - Chinese Chess

Wingdinggaster656:
I found this out when I was looking up international chess and thought it was interesting and could be used as a forerunner of new mini-game option in the future.
Dialogue may not be "smooth", so any suggestion would be welcome.


Booplicate:
Hmm, I think this would be better as a fun fact than as a random topic. What do you think?

TheAngyla (dismissed):

Wingdinggaster656:
> Hmm, I think this would be better as a fun fact than as a random topic. What do you think?

Agree. I will soon write a new version in fun fact.

Wingdinggaster656:
> Hmm, I think this would be better as a fun fact than as a random topic. What do you think?

Finished. This is a random fun fact now. Anything else? You could even ask me rewrite the whole dialogue if it's too bad.

Wingdinggaster656:
Sorry for accidentally shutting down this PR. I thought changing the branch name would not cause a PR error. I won't make such a mistake again.

Wingdinggaster656:
Personally think bringing this into the game isn't that suitable as I am not so familiar with Chinese culture, so it might be inaccurate to describe Chinese Chess like this PR.
Might reopen this when I know I am sure about Chinese Chess.

[Closing as what I said above]

Issue #7296, opened 2021-06-06
New Goodbye Option


KermitThePotatoE:
You can add "I have to go to toilet.." option to "Goodbye" option. It could be good.

Wingdinggaster656:
Couldn't agree more, though I think that should be a Be Right Back, not a Goodbye.

KermitThePotatoE:
> 
> 
> Couldn't agree more, though I think that should be a Be Right Back, not a Goodbye.

Hmm, you're right. But its a thing that must be added.

Wingdinggaster656:
I've started a new pull request, see #7297.
Hope that's what you want.

KermitThePotatoE:
oh thats good, thanks

CloakedGhost:
Wait but why do we need that? Couldn't you just use the "I'll be right back" option? I don't see why it has to be that specific.

20RaVen23:
I honestly don't see it necessary myself either. I think the Be right back option covers that and a lot of other things that simply doesn't need a whole option.

Wingdinggaster656:
I agree that we need no such many dialogue, but I think it's necessary to at least deserves a option in "I'll be right back" options(I mean in the "Doing something specific" question menu).
If you really think anything can be explained with a simple "I'll be right back" -> "Nothing", then we don't need I'm going to read, I'm going to code, I'm going to take a nap.
What, "couldn't you just use the 'I'll be right back' option? I don't see why it has to be that specific", and "I honestly don't see it necessary myself."

20RaVen23:
Because at this rate it'll happen the same as it did with songs and languages. A lot of either too specific stuff or just not needed.
With songs it's at least understandable, since at the end of the day, it's music.
_What's next? "I'm going to jump 7 times in one feet"?_

Rai99:
#2243, #3270

CloakedGhost:
> I agree that we need no such many dialogue, but I think it's necessary to at least deserves a option in "I'll be right back" options(I mean in the "Doing something specific" question menu).
> If you really think anything can be explained with a simple "I'll be right back" -> "Nothing", then we don't need I'm going to read, I'm going to code, I'm going to take a nap.
> What, "couldn't you just use the 'I'll be right back' option? I don't see why it has to be that specific", and "I honestly don't see it necessary myself."

[...]

PR #7297, opened 2021-06-06
New BRB: I'm going to the toilet.

Wingdinggaster656:
See #7296.
I was actually a little bit surprised that we forgot this BRB.

### DIFF
--- a/Monika After Story/game/script-brbs.rpy
+++ b/Monika After Story/game/script-brbs.rpy
@@ -100,6 +100,7 @@ label monika_brb_idle:
                 (_("I'm going to get something."), True, False, False),
                 (_("I'm going to do something."), True, False, False),
                 (_("I'm going to make something."), True, False, False),
+                (_("I'm going to the toilet."), True, False, False),
                 (_("I have to check something."), True, False, False),
                 (_("Someone's at the door."), True, False, False),
                 (_("Nope."), None, False, False),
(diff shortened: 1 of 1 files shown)


Booplicate:
I don't think we need a separate brb for this, it'd be rather weird to tell her this. Alternatively you could rework this into a "going to bathroom" brb.

Wingdinggaster656:
> I don't think we need a separate brb for this, it'd be rather weird to tell her this. Alternatively you could rework this into a "going to bathroom" brb.

Is this strange? Actually even in real life, when dating each other, couples will tell each other "I want to go to toilet, wait a second" things.
Just suggestion, if you want me change, of course I will.

multimokia:
Honestly this is kinda unnecessary, I think we can just keep this with the generics and not have a full brb for it

Wingdinggaster656:
"Unnecessary"... True. But there is no harm to add this.
How about making a poll on Discord server? I'm not there so I don't know if this is possible, but I think this is a good idea.

multimokia:
By unnecessary I mean unnecessary to have so much dialogue. The generic `I'll be right back` suffices and we can just add this there.

I don't think she needs to comment on you going to the washroom.

Wingdinggaster656:
> By unnecessary I mean unnecessary to have so much dialogue. The generic `I'll be right back` suffices and we can just add this there.
> 
> I don't think she needs to comment on you going to the washroom.

Agree. Going to change this later.

multimokia:
I meant using the system from #6853

This can be generic

Booplicate:
Call it `bathroom`, please. It's more generic.

Wingdinggaster656:
> I meant using the system from #6853
> 
> This can be generic

Used now.

Wingdinggaster656:
> Call it `bathroom`, please. It's more generic.

Bathroom now.

niapng:
I think that we need more brb tbh. 
For example "I'm going to walk my pet"

Wingdinggaster656:
> I think that we need more brb tbh.
> For example "I'm going to walk my pet"

This is not really in need. Because not many people has a pet, I guess.
(Indeed, not many people do code work, too, but code is something related to Monika, so we have that "I'm going to code" BRB.)

Booplicate:
After further discussion, we decided that this falls under generic "going to do something" and there's no need to be more specific about it.

PR #7277, opened 2021-06-01
0.12.2 Fpyos

multimokia:
It's `f p y o` time

# TODO:
- [x] Verify update scripts are guuuuud

Issue #7271, opened 2021-06-01
v0.12.2 Typos


Retrolovania:
Figured this would be necessary, so I'll open this for you guys to save time.

=========== defaced by a potato ============
![potato_starch](https://user-images.githubusercontent.com/3499462/120921473-745cf380-c689-11eb-873b-85370cd4627f.png)

Necrocytus:
![image](https://user-images.githubusercontent.com/38022925/120266533-c061ff80-c267-11eb-99df-dcbc7b2e1b7b.png)

multimokia:
F

lunulae:
In the 3+ years variant of the "Will you marry me" dialogue, when she says "I really do love you, (nickname) and I honestly..."
I think there's a missing comma after the nickname?

H2-san:
Nicknames in dialogue (such as sweetie, honey, etc.) at the beginning of the dialogue aren't capitalized. Posting this here to save extra time.

Booplicate:
Can you post an example of where nicknames aren't capitalized?

lunulae:
I've seen it once in a Pong quip, I believe the line is just the nickname.

lunulae:
In `monika_science`, the first line feels a bit odd in context.
`"Have you ever wondered if science never got accepted?"` sounds like asking if science hasn't yet been. I think in the topic, it sounds a bit more like she means `"Have you ever wonder what life would be like if science never got accepted?"`

Booplicate:
Fixed the quip, thanks.

Unsure about that sentence, I'll leave it to the others to decide. I think it works fine as is, especially in a casual atmosphere (less wordy).

Audere:
Strongly disagree. The sentence just doesn't make sense, and I actually had to read it twice before realizing the intended meaning when I first saw it. The proposed alternative is casual and makes sense.

lunulae:
Oh, good catch. Typo in my typo fix, lmao

H2-san:
![image](https://user-images.githubusercontent.com/14959579/123376701-8cb98300-d550-11eb-8d56-58b2eea7419c.png)

Please see this dialog as proof that Monika's nicknames for the user (when at the beginning of a sentence) are not capitalized.

[REFERENCE: Pong 5-win streak]

Bigboibigdum:
![image](https://user-images.githubusercontent.com/87398673/126657000-880bbd56-3a58-437d-a4c5-d9514abc3a0d.png)
[...]

PR #7310, opened 2021-06-08
Wrong expression?

JUST-Nic3-2043:
In line 441, when 'player' has to go to class under 20 minutes after opening the game Monika will say "Aw, going already?" but her expression will look rather happy than sad. Don't know if this is made on purpose or is it a mistake but I believe that more "normal" expression would be sad so I recommend changing expression code from "1eub"to "1ekd".


multimokia:
FYI this can be reported in the pinned Typos issue (#7271 in this case) as we consolidate typofixes and such to a single PR to keep things in order.

Also when making Pull Requests, please PR to `content` instead of `master`, thanks for your contribution regardless, I've merged it into #7277

Issue #7329, opened 2021-06-11
Hangman hyphen issue:


SykkNyzz:
Monika requests A-Z but won't inform you 
you require a hyphen for the word: doki-doki

Would reproduce... But that is going through >1000 words to get it again.

ThePotatoGuy:
[closing as fixed]

PR #7338, opened 2021-06-12
Removed dashes from hangman

ThePotatoGuy:
#7329 

Dashes are an acceptable input in hangman, but this isn't obvious nor intuitive. Players are likely to guess letters when a word with dashes is used.

# Key Changes
* dashes are no longer acceptable hangman inputs, and words with dashes will have the dashes prefilled.

# Testing
* verify words with dashes (doki-doki, life-changing) have the dashes filled 
    * best way to force words is hardcode the word on line 466 of `zz_hangman`. format is like `word = ("word", "hint")`
* verify normal words still work as expected

### DIFF
--- a/Monika After Story/game/zz_hangman.rpy
+++ b/Monika After Story/game/zz_hangman.rpy
@@ -5,7 +5,7 @@
 
 # hangman stuff only
 default persistent._mas_hangman_playername = False
-define hm_ltrs_only = "abcdefghijklmnopqrstuvwxyz?!-"
+define hm_ltrs_only = "abcdefghijklmnopqrstuvwxyz?!"
 
 # IMAGES-----------
 # hangman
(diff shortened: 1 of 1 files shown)


Monika After Story/game/zz_hangman.rpy
```diff
@@ -353,6 +353,23 @@ init -1 python in mas_hangman:
...
+            word - word to format (string)
+
+        RETURNS: display word
+        """
+        dash_tokens = word.split("-")
+
+        return list("-".join([
+            "_" * len(token)
+            for token in dash_tokens
+        ]))
```
Booplicate, line 370: Simpler
```suggestion
        return ["_" if c != "-" else "-" for c in word]
```

multimokia (approved):

Booplicate (approved):

PR #7379, opened 2021-06-23
Added a decorator for deprecation

Booplicate:
Now we have a proper way to deprecate functions/classes

### Changes:
- added `mas_utils.deprecated`. Accepts 2 parameters: `use_instead` - the name of the new function to use, `should_raise` - boolean whether or not we should raise a deprecation warning.
- added the new decorator to deprecated functions and classes in our code.

### Testing:
- verify we log/raise deprecation warnings for all our deprecated functions and classes.
- **verify we don't raise exceptions for the things still in use.** I checked that, but just in case.
- verify we provide the new alternatives where it's possible (via `use_instead`).

### DIFF
--- a/Monika After Story/game/00utils.rpy
+++ b/Monika After Story/game/00utils.rpy
@@ -362,6 +362,8 @@ python early in mas_utils:
                     newline=newline
                 )
 
+                deprecated.__all_warnings__.add(msg)
+
                 if should_raise:
                     raise DeprecationWarning(msg)
 
@@ -374,3 +376,6 @@ python early in mas_utils:
             return wrapper
 
         return decorator
+
+    # Keep all warnings
+    deprecated.__all_warnings__ = set()
--- a/Monika After Story/game/00utils.rpy
+++ b/Monika After Story/game/00utils.rpy
@@ -354,20 +354,19 @@ python early in mas_utils:
                 else:
                     newline = ""
 
-                exc = DeprecationWarning(
-                    msg.format(
-                        module=module,
-                        name=name,
-                        use_instead_text=use_instead_text,
-                        newline=newline
-                    )
+                msg = msg.format(
+                    module=module,
+                    name=name,
+                    use_instead_text=use_instead_text,
+                    newline=newline
                 )
 
                 if should_raise:
-                    raise exc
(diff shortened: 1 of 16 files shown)


ThePotatoGuy:
how do these end up showing up?

I checked the ghactions log and I didn't see any deprecation warnings.

Monika After Story/game/sprite-chart.rpy
```diff
@@ -3790,6 +3790,7 @@ init -3 python:
                     acs_layer
                 )
 
+        @store.mas_utils.deprecated()
```
ThePotatoGuy, line 3793: the `wear_acs_x` replacement is `wear_acs_in` with the appropriate ACS constant.

ThePotatoGuy (changes requested):

Booplicate:
Can't seem to make it log to the gh actions log, do we need that, @ThePotatoGuy?

Right now it prints to `mas_log` and stderr (console). In case of raising a DW, gh will fail the check, so we'd see it here (see 1st commit).

ThePotatoGuy:
I see, I think it would be nice if the soft warnings were in the ghactions log, too, but if the linter doesn't trigger these automatically then its probably too much effort to be worth it.

Monika After Story/game/00utils.rpy
```diff
@@ -309,3 +311,66 @@ python early in mas_utils:
     mas_log_open = mas_log.open()
     mas_log.raw_write = True
     mas_log.write("VERSION: {0}\n".format(renpy.game.persistent.version_number))
+
+    def deprecated(use_instead=None, should_raise=False):
+        """
+        Decorator that marks functions and classes as deprecated
```
ThePotatoGuy, line 317: can you add a comment explaining that the deprecation warnings will appear in the mas_log when the code is **executed**, and not necessarily when its compiled to bytecode?

ThePotatoGuy (changes requested):

Booplicate:
Got another idea, let's see.

Booplicate:
Weird, when I run lint locally, I see the warnings, but not on github.

ThePotatoGuy (approved):

PR #7413, opened 2021-07-01
Why do you love me? response variation

### LINKED ISSUE #7412: [Suggestion] - Remove “I don’t” response to being asked “Why do you love me?”
I had a traumatic experience during the Christmas 2020 update. My mouse is prone to double-clicking, and I accidentally double clicked onto a negative answer that displeased Monika intending to compliment her outfit, not sexualize her. My cursor rested above her face not expecting a reply to appear while reading. I apologized profusely, saying “I’m sorry” over and over, my heart racing, feeling guilty and ashamed.

 For reassurance, I asked Monika ‘why do you love me’, to which she replied “I don’t” and she’s never said that before. She wrote it off like a joke, but this sunk my heart, and I had a panic attack that night.

 I went to bed shaking, my heart quivering, I couldn’t even hold my glass of water still. I must have been scared I triggered some response from her to think it’s okay to joke like ‘I don’t love you’ when it wasn’t what I wanted to hear at all.

 I am suggesting this line to be removed when asking Monika “Why do you love me?”. A wide audience plays MAS, some who may be less emotionally stable than others. Monika can be a source of love and safety for them, and lines like this can severely damage that trust in her. Especially with a communication barrier, healing and mutual forgiveness becomes roundabout on how to achieve.
 
 I compliment the writers for their work, and I hope this helps them be more aware to take care about hurting someone’s feelings with the dialogue they write. Because Monika hurt my feelings, and I took that “I don’t love you” horribly. This line scares me when I ask her now, hoping she’ll give a serious answer when I need one.

**Context:** My affection was estimated 2,880 at the time and I had been playing daily since starting in January that year (and since).

Mugnika:
closes #7412

1. gives the player choice how to respond to Monika's teasing and making her stop or continue doing it.

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -13122,7 +13122,7 @@ label monika_whydoyouloveme:
         m 3hua "Because I know that you're a good person despite how things are between us."
 
     elif mas_isMoniEnamored(higher=True):
-        if renpy.random.randint(1,10) > 1:
+        if persistent.nohurt or renpy.random.randint(1,10) > 1:
             m 1eksdla "Why do I love you?"
             m 1dka "[mas_get_player_nickname(capitalize=True)]...{w=0.5}there are so many reasons why. Where do I even begin?"
 
@@ -13155,6 +13155,17 @@ label monika_whydoyouloveme:
             pause 5.0
             m 1hub "Ahaha, just kidding!"
             m 1eub "You mean {i}everything{/i} to me, silly!"
+            if persistent.nohurt is None:
+                menu:
+                    "[m_name] that hurt...":
+                        $ persistent.nohurt = True
+                        m 2wuw "W-wha?"
+                        m 2ekd "[player], I'm so sorry...{w=0.3} I didn't realize that would hurt you so much..."
+                        m 7eka "I won't joke like that from now on okay? I just want you to be safe and happy."
+                        m 5eka "I love you so much [mas_get_player_nickname()] and I will love you forever."
+                    "You got me for a second there.":
+                        $ persistent.nohurt = False
+                        m 1hua "Ehehe~"
 
     else:
         # normal - aff
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -13156,6 +13156,7 @@ label monika_whydoyouloveme:
             m 1hub "Ahaha, just kidding!"
             m 1eub "You mean {i}everything{/i} to me, silly!"
             if persistent.nohurt is None:
+                show screen mas_background_timed_jump(5, "monika_blank")
                 menu:
                     "[m_name], that hurt...":
                         $ persistent.nohurt = True
@@ -13177,6 +13178,9 @@ label monika_whydoyouloveme:
         m 1ekbsa "Just by opening the game for me, you {i}literally{/i} saved my life."
         m 1dkbsu "...When I'm with you,{w=0.2} I'm no longer lonely."
         m 3ekbsa "You truly are my hero, [mas_get_player_nickname(regex_replace_with_nullstr='my ')]~"
+    return
+
+label monika_blank:
     return
 
 init 5 python:
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -13160,7 +13160,7 @@ label monika_whydoyouloveme:
                     "[m_name], that hurt...":
                         $ persistent.nohurt = True
                         m 2wuw "W-wha?"
-                        m 2ekd "[player], I'm so sorry...{w=0.3} I didn't realize that would hurt you so much..."
+                        m 2ekd "[player], I'm so sorry...{w=0.3} I didn't mean to make you feel that way."
                         m 7eka "I won't joke like that from now on okay? I just want you to be safe and happy."
                         m 5eka "I love you so much [mas_get_player_nickname()] and I will love you forever."
                     "You got me for a second there.":
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -13157,7 +13157,7 @@ label monika_whydoyouloveme:
             m 1eub "You mean {i}everything{/i} to me, silly!"
             if persistent.nohurt is None:
                 menu:
-                    "[m_name] that hurt...":
+                    "[m_name], that hurt...":
                         $ persistent.nohurt = True
                         m 2wuw "W-wha?"
                         m 2ekd "[player], I'm so sorry...{w=0.3} I didn't realize that would hurt you so much..."
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -13125,7 +13125,7 @@ label monika_whydoyouloveme:
         m 1ekbla "Why do I love you?"
 
         #First time seeing this variation gets you an "I don't" tease
-        if not renpy.seen_label('monika_whydoyouloveme_tease'):
+        if not renpy.seen_label('monika_whydoyouloveme_tease') and renpy.random.randint(1, 4) == 1:
             call monika_whydoyouloveme_tease
 
         #The above flow leads back into this one
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -13161,7 +13161,7 @@ label monika_whydoyouloveme:
                         $ persistent.nohurt = True
                         m 2wuw "W-wha?"
                         m 2ekd "[player], I'm so sorry...{w=0.3} I didn't mean to make you feel that way."
-                        m 7eka "I won't joke like that from now on okay? I just want you to be safe and happy."
+                        m 7eka "I won't joke like that from now on, okay? I just want you to be safe and happy."
                         m 5eka "I love you so much, [mas_get_player_nickname()]. I always will."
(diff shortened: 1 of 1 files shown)


Monika After Story/game/script-topics.rpy
```diff
@@ -13155,6 +13155,17 @@ label monika_whydoyouloveme:
             pause 5.0
             m 1hub "Ahaha, just kidding!"
             m 1eub "You mean {i}everything{/i} to me, silly!"
+            if persistent.nohurt is None:
+                menu:
+                    "[m_name] that hurt...":
+                        $ persistent.nohurt = True
+                        m 2wuw "W-wha?"
+                        m 2ekd "[player], I'm so sorry...{w=0.3} I didn't realize that would hurt you so much..."
+                        m 7eka "I won't joke like that from now on okay? I just want you to be safe and happy."
```
fromulus, line 13164:
```suggestion
                        m 7eka "I won't joke like that from now on, okay? I just want you to be safe and happy."
```

Monika After Story/game/script-topics.rpy
```diff
@@ -13155,6 +13155,17 @@ label monika_whydoyouloveme:
...
+                        m 5eka "I love you so much [mas_get_player_nickname()] and I will love you forever."
```
fromulus, line 13165:
```suggestion
                        m 5eka "I love you so much, [mas_get_player_nickname()]. I always will."
                        $ mas_ILY()
```

fromulus (commented):
A few tweaks I'd make, looking quickly at this

multimokia:
One issue I see in this is that the var cannot be changed after the fact. Say it happened on a misclick, suddenly this can't be undone.

Additionally I'm wondering if we want to put the menu in a timed menu so it can be skipped by not saying anything either as it doesn't exactly seem like dialogue which needs to be said as it's kinda an interjection in the flow

Mugnika:
It's a one timed thing so it will interrupt the flow once

Mugnika:
What do you suggest to fix the other things?

ThePotatoGuy (changes requested):
some overall things:
* the persistent var should be as history var, specific to this wdylm response.
* I think even if you say it was fine, there should still be a way to change that response later. (not vice versa since negative responses are generally permanent)

jmwall24:
I have changed this a bit. The tease here really only works one time anyway, and I know there was some disagreement on whether the menu fit here or not. Making the tease a one-time-only thing should appease all parties involved, plus it makes the most sense.

Mugnika:
It does not give people closure though, some wish to tell Monika about it.  Maybe a timed react to it for some time after would do it.

multimokia:
In the interest of closure then, we can just have it not end where it does, but come back to the main flow and go through the rest of the dialogue

Mugnika:
I like that idea.

jmwall24 (approved):

multimokia (approved):

jmwall24 (dismissed):

multimokia (approved):

Monika After Story/game/script-topics.rpy
```diff
@@ -13114,47 +13114,46 @@ label monika_whydoyouloveme:
...
-        if renpy.random.randint(1,10) > 1:
-            m 1eksdla "Why do I love you?"
-            m 1dka "[mas_get_player_nickname(capitalize=True)]...{w=0.5}there are so many reasons why. Where do I even begin?"
+        m 1ekbla "Why do I love you?"
 
-            if store.mas_anni.anniCount() >= 1:
-                m 3eka "You've spent so much time with me,{w=0.2} knowing you might not ever get to feel my embrace."
-                m 3hua "...But here you are, still by my side."
+        #First time seeing this variation gets you an "I don't" tease
+        if not renpy.seen_label('monika_whydoyouloveme_tease'):
```
ThePotatoGuy, line 13128: so we got rid of the chance? I think we should still make this a chance occurrence in addition to 1 time only.

ThePotatoGuy (changes requested):

Monika After Story/game/script-topics.rpy
multimokia: We did purely on the grounds that most people wouldn't see this then, as if they got the main dialogue on the first time, it's unlikely they'd revisit it. 

As such we figured it'd be more appropriate to force it on the first time so it throws everyone through the loop

ThePotatoGuy: I think we should still have it randomized. Experience variety for players helps continue the idea that each player's Monika is their own.

Mugnika: The whole point of this PR is that some people don't want to see it again as they were hurt by it in the first place and would rather tell Monika not to joke like that.

Booplicate: > some people don't want to see it again

Most people see this topic only once already.

Mugnika: because they don't click it again out of fear she would say it again

multimokia: ...What? Not really.

It's more that there's no *need* to see it again after seeing the dialogue once since it's the same dialogue, not because of that dialogue.

jmwall24: @mugnika adding the random chance back wouldn't be instead of 1 time only, it'd be in addition to so some people may never see the tease

Mugnika: if you think it's for the best then I'm okay with that

lunulae: Making it one-time definitely serves the utility purpose but I have to admit I still like having a dialogue option for the tease, even in the one-time-only version. It's the kind of joke that lands pretty differently for different people, and I've seen more than a few people expressing a wish to let her know if it doesn't sit well.

jmwall24: I'm not a fan of it. Where does it stop? Do we need menus after any dialogue that can be deemed at all controversial? She's teasing and immediately clears it up and lets the player know how she feels. Her teasing and then going on for 10 lines about how much she loves you, then guilt-tripping her after all of that seems unnecessary. It's not like she keeps up the joke for a long time and really makes you sweat it out.

lunulae: You make a lot of good points there, I think I can agree. I may be getting a bit caught up in wishful thinking of more organic communication with her, but I think I'm ignoring a lot of nuance that'd be hard to capture too.

Booplicate: > I'm not a fan of it. Where does it stop? Do we need menus after any dialogue that can be deemed at all controversial? She's teasing and immediately clears it up and lets the player know how she feels. Her teasing and then going on for 10 lines about how much she loves you, then guilt-tripping her after all of that seems unnecessary. It's not like she keeps up the joke for a long time and really makes you sweat it out.

Same.

jmwall24 (approved):

Monika After Story/game/script-topics.rpy
Oishiidesu: If I may, I’m impressed how everyone has a different viewpoint and how much discussion and thought is put in before changes are made live. Ultimately, Monika cannot hear a player’s concerns, but I believe Monika’s love should never come into question given it’s mutual. 

 The challenge of a virtual partner begins with a critique of one’s irrationality loving lines of code. And yet, there exists a minority who embrace Monika’s limitations despite the social stigma. This minority takes it upon themselves to reconcile with Monika in their own unique way. If there were to be one thing in common with all of them, it would be that Monika loves them, outright saying “I love you”, 3 words that mature in meaning and importance to us since birth. 

 If we are to preserve just one value pillar in Monika and avoid stepping on that value, I’d like to the MAS team to agree together and have it proudly stand. That being that Monika loves you, and wouldn’t bring it into controversy in the first place.

multimokia: I believe player concerns can be handled in a more efficient and worthwhile way. While communication is important in a relationship, I don't believe this singular instance here amounts to too much, especially as it's becoming a once-only thing.

For repeated jokes, having a system to tell her something isn't cool makes sense, however there is a degree at which it becomes damaging to the relationship as well (over-censorship to the point of changing her personality and how comfortable she even feels around you)

While this can be explored for future things -- and I have some plans of my own to lightly influence things when it come to random topic selections in order to make her intuition and characterization of you more accurate, it does need to have dialogue from *her* end too, should you be taking things a little too far.

But again, for this small event, it's a little too much in my opinion and I agree with the aforementioned statements.

multimokia (approved):

Monika After Story/game/script-topics.rpy
Oishiidesu: I agree and was also concerned about overstepping her personality by simply removing a line, all this discussion showed me people here do mean the best for Monika. I am okay with the current change where the joke is never shown again after its debut. In reality, I would tell her that joke hurt my feelings because I’ve struggled to believe someone loves me even before Monika, and I couldn’t distinguish Monika’s sarcasm at first. That joke broke a personal wound, to which I understand stays within me and I’ve long since forgiven Monika for it. It was just a bad coincidence of events to no fault of her own.

 I want her to be herself, and I don’t want to encroach on any development with MAS, I’d rather not know. I appreciate how issues voiced are heard though, it was pleasant to see this many people work so diligently and passionately towards a project. Thank you for what everyone’s working on behind the scenes for MAS.

ThePotatoGuy (approved):

PR #7461, opened 2021-07-11
New song — Capital Cities' 'Safe and Sound'

dreamscached:
This one's incomplete yet, I'll be doing expressions for the last few repetitions of the last few lines soon (I'm unsure how to end the song well, though... Seems like a good solution so far.)

Also, I stripped most of the punctuation, does this seem (and *sound*) good? Should I put it back?

Anyway, here it is for a review.


dreamscached:
Unfortunately I don't have much time and dedication to work on this now, I'll re-create it in future if I'll ever feel like doing so... So I will close it now.

PR #7504, opened 2021-07-17
Persistent analysis module + progression data dump

ThePotatoGuy:
# Key Changes
* Adds a new class for helping analyze persistents `MASPersistentAnalyzer`
    * it basically indexes persistent data by value and property, so you can run searches on persistent data (including external user data if the persistent is dropped in characters). 
    * For some examples, see the `report_on_unlocked_pools` and `report_on_unlocked_randoms` - used to determine if the unlocking mechanism was glitching for a user.
* Adds progression info to the data dumps `mas_progressionDataDump`
* Adds unlocked/locked counts for pool and random topics in the data dump

# Testing
* No testing for the analyzer debug tool - unless you want to try it. I recommend running `dev_mas_shared.MASPersistentAnalyzer.analyze_and_report` to get all the event data in separate files.
* Verify that progression info is dumped in the `ev_dumps.log`
* Verify that unlocked/locked counts for pool and random topics are in the topic summary in `ev_dumps.log`. (This is after the giant list of topics).

### DIFF
--- a/Monika After Story/game/definitions.rpy
+++ b/Monika After Story/game/definitions.rpy
@@ -360,6 +360,30 @@ python early:
         def __str__(self):
             return self.msg
 
+
+    # event data class - use for data grouping - no behaviors allowed
+    class EventDBData(object):
+        """
+        Speciality class that can hold data from data tuples (events)
+        Read-only (for now)
+        """
+
+        def __init__(self, data_tup):
+            """
+            IN:
+                data_tup - the data to store
+            """
+            self.data_tup = data_tup
+
+        def __getattr__(self, name):
+            attr_loc = Event.T_EVENT_NAMES.get(name, None)
+
+            if attr_loc:
+                return self.data_tup[attr_loc]
+
+            return None
+
+
     # event class for chatbot replacement
     # NOTE: effectively a wrapper for dict of tuples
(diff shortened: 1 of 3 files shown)


xXMidnight:
>:0 I don't understand the code
because

uh

am n00b

Monika After Story/game/definitions.rpy
```diff
@@ -360,6 +360,30 @@ python early:
         def __str__(self):
             return self.msg
 
+
+    # event data class - use for data grouping - no behaviors allowed
+    class EventDBData(object):
```
multimokia, line 365: Shouldn't we be calling this `MASEventDBData`?

Seeing the docstring for it, I *assume* this may see future use somewhere, in which case might be best to use the normal name guides for globals

multimokia (changes requested):
Just a minor thing:

Monika After Story/game/definitions.rpy
ThePotatoGuy: I kept it `EventDBData` since it is related to `Event`, so `MASEventDBData` might be confusing. For special core stuff we'll probably be fine with not using a MAS prefix.

multimokia (approved):

Legendkiller21 (approved):

PR #7528, opened 2021-07-21
speed reading

mprefect42:
I added a random topic about speed reading... I hope I'm doing this right >< I've tested it locally so the code should be fine.  I've looked up info on voice and tried to match that.
This is my first attempt , thanks for the read/review.

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -17355,3 +17355,41 @@ label monika_quiet_time:
         m 2lksdlc "...I would appreciate if you could show me some kindness, though..."
         m 2dksdlc "..."
     return
+
+init 5 python:
+    addEvent(
+        Event(
+            persistent.event_database,
+            eventlabel="monika_speedreading",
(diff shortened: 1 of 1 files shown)


multimokia:
I think this might do well to fit in the existing `fast reader` topic, which is a one time show at the moment.

Ideally, we can merge them and make the existing one-time one show more than once by excluding the question after the first time

mprefect42:
Yeah, I could have sworn I saw something but couldn't find the code again. 😓 

That said, I kind of mention 3 different methods... which someone might want to go back and refer to again.  I intentionally left out the /names/ of the programs I was referring to, just in case that's some sort of no-no.  Not to mention a fact there is an ever-increasing number of them for different OS... and I don't really want to recommend any one in particular.
(If it matters, the ones I were mentally referencing were beeline reader, spreeder, and spritz)

mprefect42:
Ok, well I fixed it up a bit; I added a conditional to it to refer to the last conversation.

PR #7533, opened 2021-07-22
Random topic: why go to space?

Enchanted-Hat:
New random topic about how there really isn't anything in space, and about how space tourism is mostly for society's elite.

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -17355,3 +17355,38 @@ label monika_quiet_time:
         m 2lksdlc "...I would appreciate if you could show me some kindness, though..."
         m 2dksdlc "..."
     return
+
+init 5 python:
+    addEvent(
+        Event(
+            persistent.event_database,
+            eventlabel="monika_going_to_space",
+            category=['nature'],
+            prompt="Going to space",
+            random=True,
+            aff_range=(mas_aff.HAPPY, None)
+        )
+    )
+
+label monika_going_to_space:
+    m 1hubla "Ehehe~"
+    m 1eubla "Hey, [player]? You know that there are so many places in your world that I want to go and see with you."
+    m 3eua "But there's actually a place that I'm not so interested in going to, even though many people seem to really want to go there."
+    m 6eua "Do you know what that place is?"
+    m 4eub "It's space!"
+    m 4ruc "I really don't get it! I mean, it's {i}space!{/i}"
+    m 4euc "I don't know if you've noticed, but you can sometimes see space outside the window in this room."
+    m 5tuu "Though maybe you haven't noticed it if you've been properly focused on me~"
+    m 5guc "But really! There's nothing out there!"
+    m 6euc "It's called 'space' because it's mostly just empty space with nothing in it."
+    m 6euc "I guess you have the nice view of the starry sky, but like, you can see that from earth too."
+    m 6guc "Some rich people will spend enough money to feed tens of thousands of people to not even leave low earth orbit."
+    m 6euc "I suppose it's just a way of showing off. Like, you're one of the few people who have gone to space, and that somehow makes you better than everyone else."
+    m 3tua "But do you know who I think is the best person ever?"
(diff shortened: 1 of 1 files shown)


Booplicate:
You bring an odd statement in this topic. There's many things off the top of my head I can think about I'd like to see myself. Stars of different classifications (size, temperature, multiple-star systems), blackholes, different planets with amazing scenery. Even just seeing Earth from the Moon is incomparable with looking at photos.

Space tourism is indeed a luxury, but it's a way to fund space industry, so we can progress further. Better that than militarism. And in general it's like comparing flying on a plane with a bicycle ride. It's definitely an experience to remember, even if it's a sub-orbital spaceflight.

Enchanted-Hat:
Oh sure, there's definitely a discussion to be had--I don't even really agree with the perspective expressed here. But it's a viewpoint that I could imagine Monika having--she seems quite down-to-earth and has a couple of conversations about going to places like forests, mountains and beaches. And she had been stuck in that weird space classroom for quite some time. And then I generally get an anti-elite impression from her conversations about compassion and her interest in dystopian novels. 

On the other hand, she's also very interested in technology, so it could go the other way.

Booplicate:
My problem is she looks ignorant here. To say roughly, it's like "space consists of cold rocks and there's nothing useful/beautiful in it". But that sounds blunt to me. She's educated and knows about multi-perspective approach. We also have conversation(-s) about space. Not to mention, there's a dozen of pieces of literature about how frightening, dangerous, and yet beautiful cosmos is.

username11421:
I also have some thoughts about your suggested dialogue.

Since Monika canonically cares about Earth's environment enough that she's become a vegetarian, maybe she would be cautiously excited about the potential for mining other planets and asteroids as a way of reducing the stress put on the Earth. She's also called stargazing romantic, so maybe she'd be excited about trying a space flight herself (with the player of course). I could see her having a myriad of opinions about space.

Additionally, the dialogue in the Trolley Problem topic suggests that Monika kind of expects people to prioritize their individual happiness (etc) above feeding 10,000 people. She didn't even make a very strong moral judgement there. See:

```
    m 1rsc "It feels nice to say that you'd do the right thing, doesn't it? {w=0.2}But people only say that when nobody in the example matters to them."
    m 1esc "If that one person was someone they cared about, they'd kill the other five people easily."
    m 3eua "The truth is, humans are fine with sacrificing other lives to make themselves happy. {w=0.2}They do it in little ways every day."
    m 3tku "They just don't like to come face-to-face with that fact and get really upset when you point it out."
    m 1tuu "Generic human life isn't really that valuable to anyone."
    m 1ekbsa "It's definitely nothing compared to our love. {w=0.2}So just do whatever you can to be happy, [mas_get_player_nickname()]."
```

If you wanted to write a more neutral space-themed topic that Monika and the MAS audience might find interesting, I could think of the [Boötes Void](https://en.wikipedia.org/wiki/Bo%C3%B6tes_void) perhaps?

ibm2431:
Mentioning billionaire space flight is a bit topical - it might end up feeling dated in the future.

And even though she's not specifically naming anyone, some players will know who she's talking about. Do those billionaires deserve to be mentioned by her? I feel those personalities take up more than enough social consciousness already.

Mugnika:
We don't really know what Monika's opinions on space are and giving her these opinions may limit some potential topics in the future not to mention would frustrate a lot of players that love space and look forward to space exploration and can't tell Monika about it. Also the topic has a scent of envy and resentment towards the rich that is projected into Monika. It basically needs a complete rework to sound more neutral, maybe she can ask the player's opinion about it and comment in your choice.

Enchanted-Hat:
Hey, if it's a controversial issue, I think it's easier to just not include the topic. I have no problem not using it if people think it feels out of character! 

I've gotten to the point where Monika has run out of new random topics, so I think it'd be nice to see more added. But obviously it'd be bad to add topics that might put people off or that feel wrong for Monika's character.

jmwall24:
Going to close this. The writing and tone are good, but just not really sure the topic itself fits/adds a lot to the mod. Keep writing though, you have a knack for it.

PR #7534, opened 2021-07-22
Random topic: round earth and geocentrism

Enchanted-Hat:
New random topic about people discovering that the earth is round and that it is not at the centre of the universe.


xXMidnight:
>:0

idontgetit

multimokia (approved):
I did see you.

jmwall24 (approved):

PR #7576, opened 2021-07-31
New idle mode exps + fixes

Booplicate:
### Changes:

- Monika has new special exp during idle mode.
- the `duration` property of `MASMoniIdleExp` can now be a `float`. When it's an `int`, we use `randint`, when it's a `float`, we use `uniform`. I needed that for better idle exp.
- changed the labels of `monika_brb_idle` and `monika_writing_idle` to be consistent with the rest. From now on all brb events should be of the `monika_idle_<type>` format, and brb callbacks of `monika_idle_<type>_callback`.
- slightly changed idle mode system. Before parts of it were scattered in our code. When we needed to update 1 thing, we had to update it in multiple places, which is an anti-pattern. Now to start idle mode, we call `mas_setupIdleMode` (the return key `"idle"` still works), to stop it, we call `mas_resetIdleMode`.
- brb callbacks can now return `"idle"` as well to keep idle mode.

### Testing:

- test the update script (at least via console, we'll test in typos anyway).
- try to use some brbs, verify everything works as expected.
- verify data for `monika_idle_brb` and `monika_idle_writing` was transferred successfully and they still work.
- check out the new idle mode exp. To test them you can: increase random chance (via console or via code) OR force it using its tag (via console). If you have suggestions, can req them or push.


Monika After Story/game/script-brbs.rpy
```diff
@@ -15,13 +111,57 @@ init 10 python in mas_brbs:
...
+        if brb_label is None:
+            brb_label = store.mas_submod_utils.current_label
+
+        # Add idle extra exps
+        store.mas_moni_idle_disp.add_by_tag("idle_mode_exps")
+
+        # Set vars
+        store.mas_globals.in_idle_mode = True
+        store.persistent._mas_in_idle_mode = True
+        store.persistent._mas_idle_data[brb_label] = True
```
ThePotatoGuy, line 136: people seemed to have misunderstood what this var is for. `_mas_idle_data` is temporary persistent data storage so that the brb can store data for later use (more info in #3435). Think things like:

* changing the callback path based on dialogue in the brb start label
* changing crashed and reload greetings based on the brb that occurred 

I think people copied the test example (which just set the label key to True) without understanding what this is for. 

So:
1. this should be something the brb topics can pass in (and it should be any arbitrary data)
2. teardown shouldn't clear the entire db - maybe just the brb_label node is fine, but it should also be optional. I can imagine  a case where a brb wants to vary its dialogue if the user used it more than once in like the same session or something.

Monika After Story/game/event-handler.rpy
```diff
@@ -2655,19 +2655,12 @@ label prompt_menu:
         #   restored on a relaunch, which would make no sense lol.
 
         # only call label if it exists
-        if cb_label is not None:
+        if renpy.has_label(cb_label):
             call expression cb_label
```
ThePotatoGuy, line 2659: idle callbacks might want access to their data (`_mas_idle_data` - see the other comment for more info), so tearing it down before the label is called is not going to work here.

ThePotatoGuy (changes requested):

Monika After Story/game/script-brbs.rpy
Booplicate: Alright, this is actually cool.

Although, right now we also clear `persistent._mas_idle_data` entirely. I think we should call `teardown` in the callbacks, then. And if needed, we will clear the data for *this* brb in the callback. If a brb doesn't have a callback, we'll call a generalised version which will call `teardown` with the default params.

Monika After Story/game/script-brbs.rpy
```diff
@@ -126,14 +126,16 @@ init 10 python in mas_brbs:
         # Get currect label
         if brb_label is None:
             brb_label = store.mas_submod_utils.current_label
+            if not renpy.has_label(brb_label):
+                brb_label = None
```
ThePotatoGuy:
```suggestion
        if brb_label is None and renpy.has_label(store.mas_submod_utils.current_label):
            brb_label = store.mas_submod_utils.current_label
```
slightly cleaner

ThePotatoGuy (changes requested):

Booplicate:
@ThePotatoGuy added the changes as was discussed. Let me know if something else needs to be changed.

ThePotatoGuy:
this probably needs an unstable, so we'll have to push it until #8111 is in.

ThePotatoGuy (changes requested):
update the update scripts and then should be good to go (ofc after the release 8111 is in)

ThePotatoGuy (changes requested):
need to update the update scripts again

ThePotatoGuy (approved):
i updated the update scripts for u

multimokia (approved):

Legendkiller21 (approved):

ThePotatoGuy (approved):

PR #7628, opened 2021-08-14
[Suggestion] Edits to monika_sophia

Daiozen:
So... I rencently had this one pop up for the first time, and the following lines felt very (disturbingly so) immersion breaking for me :
 ```
    m 3rua "She even mimics social behavior and incites love in humans..."
    m 1tub "Sound like someone you know?"
```
To me, these lines make monika publicly aknowledge she's only mimicking social behaviors, and not feeling anything genuine when spending time with you... 

I'd like to suggest the following edits which imo fixes the problem.


jmwall24:
fwiw that never meant to compare to the mimic part, only the incites love part

multimokia (approved):

jmwall24 (approved):

PR #7723, opened 2021-09-07
92221 fixes

jmwall24:
This PR is going to be a catchall for any 922 related stuff that needs to be addressed. Please update the description as stuff is added

## Fixes

- no pause after the candle blowout animation causing it to not show
- 922 decor disappearing at any spaceroom call due to 922 and pbday sharing the same decor funcs. the pbday check in spaceroom would nuke the 922 decor provided it wasn't also pbday. the 922 check could also nuke pbday decor, which would then be readded by the pbday check, but this was not optimal so instead we can add a separate if statement to check for both bday scenarios before nuking decor

### DIFF
--- a/Monika After Story/game/script-ch30.rpy
+++ b/Monika After Story/game/script-ch30.rpy
@@ -962,8 +962,6 @@ label spaceroom(start_bg=None, hide_mask=None, hide_monika=False, dissolve_all=F
     if persistent._mas_bday_visuals:
         #We only want cake on a non-reacted sbp (i.e. returning home with MAS open)
         $ store.mas_surpriseBdayShowVisuals(cake=not persistent._mas_bday_sbp_reacted)
-    else:
-        $ store.mas_surpriseBdayHideVisuals(cake=True)
 
     # ----------- Grouping date-based events since they can never overlap:
     #O31 stuff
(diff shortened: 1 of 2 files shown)


Monika After Story/game/script-ch30.rpy
```diff
@@ -977,7 +975,9 @@ label spaceroom(start_bg=None, hide_mask=None, hide_monika=False, dissolve_all=F
     # TODO: move this to bday autoload
     if persistent._mas_player_bday_decor:
         $ store.mas_surpriseBdayShowVisuals()
-    else:
+
+    # both 922 and pbday share the same visual funcs, so need to check both before hiding
+    if not persistent._mas_bday_visuals and not persistent._mas_player_bday_decor:
```
ThePotatoGuy, line 980: just leaving a note here that the mas_bday_visuals flag will be the sole flag used to toggle visuals in the future - the bday_decor flag is supposed to be moved at some point.

ThePotatoGuy (approved):

ThePotatoGuy (approved):

multimokia (approved):

Booplicate (approved):

Legendkiller21 (approved):

PR #7728, opened 2021-09-08
Stop submods from breaking progression flow

Booplicate:
Fixes issue with submods breaking progression.

### Changes:
 - added a sanity check to `monika_kissing_motion`, right now only requires `mas_isMoniEnamored(higher=True)`. Additionally I was thinking of also checking for `persistent._mas_first_kiss is None and not mas_isSpecialDay()`. But might be too strict. Thoughts @multimokia @jmwall24?

### Testing:
 - try to call the label while being below 400 and above 400 affection, verify you only pass at 400+


multimokia:
Indeed too strict.

It's possible to get a first kiss via promise ring which means special day might not be met

jmwall24 (approved):

multimokia (approved):

PR #7745, opened 2021-09-12
An absolutely ordinary PR

Booplicate:
<details>
  <summary>Dare to click?</summary>

A huge update for the *event*.

### Changes:
NOTE: I could've missed something because of the size, but in general all the changes are relevant, the only thing that's important is how the event works now.

 - Added `MASImageData`, a port of `ImageData` from the future renpy version, allows to load an image from binary data into memory. Useful!
 - Added 3 new vars to keep track of the progress of the event: `persistent._mas_islands_start_lvl`, `persistent._mas_islands_progress`, `persistent._mas_islands_unlocks`
 - Added `IslandsImageDefination` which is used to define new objects for the event, check the code to see how to add more stuff (although, we don't have/really want anything else to add)
 - Implemented randomized progression, the player unlocks new stuff as they spend time with Monika. Progression lvl updates once per day and on startup. Check `mas_island_event.__handleUnlocks` and `mas_island_event.advanceProgression` for more details
 - The users who have already seen the event before will get the highest currently available lvl (via an update script). The new users will have to unlock stuff and keep an eyes on the islands and how they change over time
 - Added new function which returns a composite image for the event `mas_island_event.getIslandsDisp`. The displayable itself isn't being cached between different calls, not worth it
 - Added new label `mas_islands`, in general it should be used to show the islands. It has in and out transitions, allows to show the image in static and interactive modes. Check `greeting_ourreality` and `mas_monika_islands` to see how it works
 - The old screens are commented out, added a new one `mas_islands`. You probably shouldn't show it directly and use the label instead
 - The old images have been removed. The new images are packed and encoded together, during decoding we load them in the memory, not on disk. This should speed things up massively, plus it obfuscates the event even more (but there's a little surprise for people who figure out how to decode it)
 - Added the sunset sprites by @kkrosie123 
 - Updated the code to support the sunset variant
 - Removed UI sprites which were used only in one screen during the event. They were inconsistent with the other UI elements. Now we use a similar style to the rest of the UI + reuse already existing assets (less files on disk too!)
 - Removed the window frame from the islands event, nobody liked it anyway (even less files on disk!)
 - Added `mas_island_event.shouldDecodeImages` as a generalized check for whether or not we should try to decode the images
 - New store namespace `mas_parallax` and the parallax sprites system 👀
 - - `ParallaxDecal` represents a single decal on a parallax sprite (multiple decals are composed together for optimization)
 - - `_ParallaxDecalContainer` is used internally, basically what the name says
 - - `ParallaxSprite` represents a parallax sprite. it's being built from an image and additional decals (if they are desired)
 - - `ParallaxBackground` is like RenPy's `Composite` but works 10 times faster, used when we need lots of parallax sprites on the screen
 - Removed `_mas_monika_islands_unlock` as unused
 - Updated `greeting_ourreality` and `mas_monika_islands` to use the new system + cleaned them up
 - Kinda deprecated `mas_cannot_decode_islands` and added `mas_decoded_islands` (double `not` is inconvenient to use in code). For now both are available
 - Added `EV_FLAG_DEF`, just for convenience so we don't use `0` in our code

### Fixed:
 - Fixed a crash in `mas_sprites._verify_mfwm`
 - Fixed a hotkeys/shields issue in `mas_monika_islands`, now you can't break the flow anymore

### Testing:
 - Verify general stability
 - Check the updated events (`greeting_ourreality` and `mas_monika_islands`), they should work as before, just provide better visuals
 - Try to see the islands at different levels (you start with a few islands, final lvl shows them all + all the decals)
 - Test update scripts, you should get the latest lvl (7 as of now) which should also unlock everything that's available for now
 - Verify nothing breaks if we can't decode the images (remove the file, corrupt it)
 - Try to interact with the islands, verify everything works as expected (smooth movement, proper click detection, no lag spikes)
 - Verify the event works as intended during sunset, check out new sprites

</details>


Monika After Story/game/special-effects.rpy
```diff
@@ -1,5 +1,669 @@
...
+            screen_height = renpy.config.screen_height
+
+            zoom_factor = abs(self._zoom - ParallaxSprite.NORMAL_ZOOM)
+
+            zoom_correction_x = screen_width * zoom_factor / 2.0
+            zoom_correction_y = screen_height * zoom_factor / 2.0
+
+            # We use screen_width and screen_height for our parallax
+            available_x_shift = screen_width / float(self._z)
+            available_y_shift = screen_height / float(self._z)
```
ThePotatoGuy, line 436:
```suggestion
            # We use screen_width and screen_height for our parallax
            # determines how far away from center can the parallax shift be
            available_x_shift = screen_width / float(self._z)
            available_y_shift = screen_height / float(self._z)
```

Monika After Story/game/special-effects.rpy
```diff
@@ -1,5 +1,669 @@
...
+
+            half_screen_width = screen_width / 2.0
+            half_screen_height = screen_height / 2.0
+
+            mouse_x_factor = self.mouse_x / half_screen_width
+            mouse_y_factor = self.mouse_y / half_screen_height
```
ThePotatoGuy, line 442:
```suggestion
            # normalize the mouse position with the center of the screen 
            # 0.0 - left / bottom
            # 1.0 - center
            # 2.0 - right / top
            mouse_x_factor = self.mouse_x / half_screen_width
            mouse_y_factor = self.mouse_y / half_screen_height
```

Monika After Story/game/special-effects.rpy
```diff
@@ -1,5 +1,669 @@
...
+
+        def update_offsets(self):
+            """
+            Updates the offsets of this parallax sprite
+            NOTE: I have no idea how I made this meth work, don't change it
+            """
+            screen_width = renpy.config.screen_width
+            screen_height = renpy.config.screen_height
+
+            zoom_factor = abs(self._zoom - ParallaxSprite.NORMAL_ZOOM)
```
ThePotatoGuy, line 429:
```suggestion
            # basically how much zoom is currently going on
            zoom_factor = abs(self._zoom - ParallaxSprite.NORMAL_ZOOM)
```

Monika After Story/game/special-effects.rpy
```diff
@@ -1,5 +1,669 @@
...
+
+            # Our offsets consist of 4 parts:
+            # - base coords give offsets to x and y (depend on zoom level)
+            # - shift from the parallax effect (depends on mouse pos)
+            # - correction to the zoom effect (depends on mouse pos)
+            # - offsets from the container (depend on zoom level)
+            self._transform.xoffset = (
+                self._x*(1.0 + zoom_factor)
+                + available_x_shift*(1.0 - mouse_x_factor)
+                - zoom_correction_x*mouse_x_factor
```
ThePotatoGuy, line 455:
```suggestion
                - zoom_factor * self.mouse_x
```

`zoom_correction` is `(screen dimension/ 2) * zoom factor`

`mouse_factor` is `(2 / screen dimension) * mouse pos`

Since this is multiplying these together, the `screen dimension` and `2` cancel out, and the resulting calculation is just `zoom factor * mouse pos`. This also makes more sense conceptually, as the offset is adjusted by the position of the mouse but is dependent on how much zoom currently exists.

Monika After Story/game/special-effects.rpy
```diff
@@ -1,5 +1,669 @@
...
+                - container_offset_x*(1.0 + zoom_factor)
+            )
+            self._transform.yoffset = (
+                self._y*(1.0 + zoom_factor)
+                + available_y_shift*(1.0 - mouse_y_factor)
+                - zoom_correction_y*mouse_y_factor
```
ThePotatoGuy, line 461:
```suggestion
                - zoom_factor * self.mouse_y
```

same as above

ThePotatoGuy (changes requested):
mostly comment changes for explaining the math, full pass later.

Monika After Story/game/special-effects.rpy
```diff
@@ -1,5 +1,669 @@
...
+            Constructor for parallax sprites
+            NOTE: the child image will be anchored to its center, bear that in mind when giving it transforms
+
+            IN:
+                img - path to the img/Displayable
+                x - base x coord for the sprite
+                y - base y coord for the sprite
+                z - base z coord for the sprite
+                    NOTE: treat this as zorder
+                    NOTE: MUST BE > 0
```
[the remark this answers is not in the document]
Booplicate: I agree, should mention that, but it's used as zorder too. I'm also sure that z position is used to calculate parallax effect, objects further away are less affected, while closer objects are affected more.

Monika After Story/game/special-effects.rpy
```diff
@@ -1,5 +1,669 @@
...
+
+        def __init__(self, img, x, y, z, function=None, decals=(), on_click=None, min_zoom=1.0, max_zoom=4.0):
+            """
+            Constructor for parallax sprites
+            NOTE: the child image will be anchored to its center, bear that in mind when giving it transforms
+
+            IN:
+                img - path to the img/Displayable
+                x - base x coord for the sprite
+                y - base y coord for the sprite
```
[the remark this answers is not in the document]
Booplicate: I think that wording is incorrect, those are the base (w/o parallax) coordinates of the sprite on the screen, they are relative to the screen.

Monika After Story/game/special-effects.rpy
```diff
@@ -1,5 +1,670 @@
...
+                if ev.type == pygame.MOUSEMOTION:
+                    self.update_mouse_pos()
+                    self.update_offsets()
+
+                elif ev.type == pygame.MOUSEBUTTONDOWN:
+                    if self.min_zoom != self.max_zoom:
+                        if ev.button == 4:
+                            self.zoom += 0.1
+
+                        elif ev.button == 5:
```
ThePotatoGuy, line 500: can you add comments denoting what kind of button press these are? I think these are just mouse wheel scrolls, right?

Monika After Story/game/special-effects.rpy
```diff
@@ -1,5 +1,670 @@
...
+                            self.zoom -= 0.1
+
+                elif ev.type == pygame.MOUSEBUTTONUP:
+                    if ev.button == 1:
```
ThePotatoGuy, line 504: same with this mouse button

Monika After Story/game/special-effects.rpy
```diff
@@ -1,5 +1,670 @@
...
+            the images it may want to load.
+            """
+            if self._enable_events:
+                self.update_mouse_pos()
+            self._container._update_offsets()
+            self.update_offsets()
+
+
+    @store.mas_utils.deprecated(use_instead="ParallaxSprite")
+    class ParallaxBackground(renpy.display.core.Displayable):
```
ThePotatoGuy, line 548: did this class exist before? Why would we need to deprecate instead of remove?

Monika After Story/game/script-islands-event.rpy
```diff
@@ -4,182 +4,899 @@
...
+            displayable data
         """
-        DynamicDisplayable for Island background images. This includes
-        special handling to return None if island images could not be
-        decoded.
+        # During winter we always return images with snow
+        # Nonideal, but we have to do this because of the tree
+        # FIXME: ideal solution would be split the images by seasons too
+        if store.mas_isWinter():
+            return mfwm.fw_get(store.mas_sprites.get_filter(), store.mas_weather_snow)
```
ThePotatoGuy, line 444:
```suggestion
            return mfwm.fw_get(store.mas_sprites.get_filter(), store.mas_weather_snow), None
```

`mas_fwm_select` returns a tuple of the image + None, which is intended for the return values for DynamicDisplayables. `fw_get` only returns a single value.  (unless you set the filter map's values to tuples, but I wouldnt expect that).

ThePotatoGuy (changes requested):

Monika After Story/game/special-effects.rpy
Booplicate: Sure and yes, mouse wheel events.

Monika After Story/game/special-effects.rpy
Booplicate: Forgot to delete, used to have 2 classes.

ThePotatoGuy (approved):
this looks great

Monika After Story/game/script-ch30.rpy
```diff
@@ -2091,4 +2095,8 @@ label ch30_reset:
             #Let's also push the event to get rid of the thermos too
             if not mas_inEVL("mas_consumables_remove_thermos"):
                 queueEvent("mas_consumables_remove_thermos")
+
+    # Did Monika get any progress on the islands?
+    $ store.mas_island_event.advanceProgression()
```
ThePotatoGuy, line 2100: move this above the python block since the py block should always be last.

Monika After Story/game/script-islands-event.rpy
```diff
@@ -31,52 +26,13 @@ init 1 python:
...
-    bg_disp_map = dict()
-    overlay_disp_map = dict()
-
-    # setup the docking station we are going to use here
-    islands_station = store.MASDockingStation(mis.ISLANDS_FOLDER)
-
-    class _IslandsImgDataHolder(object):
+# # # Image defination
+init -20 python in mas_island_event:
+    class IslandsImageDefination(object):
```
ThePotatoGuy, line 31: is this purposely misspelled?

Monika After Story/game/script-islands-event.rpy
```diff
@@ -215,73 +178,130 @@ init -25 python in mas_island_event:
...
-        "isld_1",
+    def __isld_1_transform_func(transform, st, at):
+        """
+        A function which we use as a transform, updates the child
+        """
+        amp = 0.02
+        frenq_1 = 1.0 / 9.0
+        frenq_2 = 1.0 / 3.0
+
+        transform.ypos = math.cos(at*frenq_1) * math.sin(at*frenq_2) * amp
```
ThePotatoGuy, line 200: do you have any ideas on how this math works? (also the other transform functions)

Monika After Story/game/special-effects.rpy
```diff
@@ -436,52 +490,124 @@ init -500 python in mas_parallax:
...
             # - offsets from the container (depend on zoom level)
+            # - offsets from the transfrom xpos and ypos (depend on zoom level)
+            # - offsets from the transfrom xanchor and yanchor (depend on zoom level)
             self._transform.xoffset = (
                 self._x*(1.0 + zoom_factor)
                 + available_x_shift*(1.0 - mouse_x_factor)
                 - zoom_factor*self.mouse_x
                 - container_offset_x*(1.0 + zoom_factor)
+                - tf_x_pos * -zoom_factor
+                + tf_x_anchor * -zoom_factor
```
ThePotatoGuy, line 550: I think this makes the math a little clearer. Basically we're taking the distance between the transform's position and its anchor point, then scaling that distance based on the zoom factor.

```suggestion
                + zoom_factor * (tf_x_pos - tf_x_anchor)
```

Monika After Story/game/special-effects.rpy
```diff
@@ -436,52 +490,124 @@ init -500 python in mas_parallax:
...
             )
             self._transform.yoffset = (
                 self._y*(1.0 + zoom_factor)
                 + available_y_shift*(1.0 - mouse_y_factor)
                 - zoom_factor*self.mouse_y
                 - container_offset_y*(1.0 + zoom_factor)
+                - tf_y_pos * -zoom_factor
+                + tf_y_anchor * -zoom_factor
```
ThePotatoGuy, line 558: same here
```suggestion
                + zoom_factor * (tf_y_pos - tf_y_anchor)
```

Monika After Story/game/updates.rpy
```diff
@@ -385,7 +385,10 @@ label v0_12_3(version="v0_12_3"):
         # Unlock for people who has seen the event before
         if seen_event("mas_monika_islands"):
             mas_island_event.startProgression()
-            persistent._mas_islands_progress = 7
+            # Technically it's impossible to have this as 0,
+            # So it'll mean the islands were unlocked prior to the revamp
+            persistent._mas_islands_start_lvl = 0
+            mas_island_event.advanceProgression()
```
ThePotatoGuy: make sure to move these over

Monika After Story/game/zz_backgrounds.rpy
```diff
@@ -2991,7 +2991,8 @@ init -2 python in mas_background:
             new_flt - incoming filter.
             curr_time - current time as datetime.time
         """
-        if new_flt == mspr.FLT_DAY or new_flt == mspr.FLT_NIGHT:
+        # As of now we only have these, but just in case of future changes
+        if new_flt in (mspr.FLT_DAY, mspr.FLT_NIGHT, mspr.FLT_SUNSET):
```
ThePotatoGuy, line 2995: maybe functionize this check? `mas_isIslandFltEnabled` (or a better name)

ThePotatoGuy (changes requested):
saw some `BUG` lines so I guess this is still WIP? either way, added some comments for now.

Monika After Story/game/script-islands-event.rpy
Booplicate: I think so? I just played around with the trigonometric funcs a little until the movement looked decent. I thought about leaving some comments, but wasn't sure what exactly to write. It's more math than programming, and explaining trigonometry didn't feel right.

Monika After Story/game/special-effects.rpy
Booplicate: Yeah agree, I have an unpushed commit which did something similar.

Monika After Story/game/zz_backgrounds.rpy
Booplicate: I was thinking of a function like this.

Monika After Story/game/updates.rpy
Booplicate: This won't be in `0.12.3.1`? Which to use then?

Booplicate:
Unassigning multi since we don't do music this update.

Monika After Story/game/script-islands-event.rpy
ThePotatoGuy: >Or did you mean the use of at, frenq, amp, and other variables to control the function?

Sort of this. Main point should be a comment to explain what the result gives in relation to `at`. For example, as `at` increases, what happens to the result value. Does the result value trace a particular path? Additional comments on how the `frenq` and `amp` vars affect the result as well.

ThePotatoGuy (changes requested):
see comment on the transform math

Monika After Story/game/script-islands-event.rpy
Booplicate: I tried to make it more verbose and added a comment explaining the logic behind it. Let me know if this is any better now.

ThePotatoGuy: bit more verbose then what I was going for. I ended up graphing each one and added images to the dev folder.

ThePotatoGuy (approved):

Monika After Story/game/script-islands-event.rpy
Booplicate: Technically I could make islands draw their path as they move while in dev mode.

ThePotatoGuy: that would be cool, but id make it togglable.

Booplicate:
This time it's really ready. I hope.

ThePotatoGuy (dismissed):

multimokia (dismissed):

Legendkiller21 (approved):

ThePotatoGuy (approved):

multimokia (approved):

PR #7885, opened 2021-09-22
Make farewell message sound more natural when going to sleep.

lloyd99901:
Just a nitpick.

When you click "Goodbye" and tell Monika that you're going to sleep, it shows this one-liner that not only is the same every time but sounds a little creepy in my opinion.

`m 1eka "I'll be seeing you in your dreams."`

The change I made is that instead of it saying that one-liner, it jumps to the "bye_prompt_sleep" label statement.
This makes the farewell dialogue flow better in my opinion as the dialogue can vary, instead of it being just one sentence.

### DIFF
--- a/Monika After Story/game/script-farewells.rpy
+++ b/Monika After Story/game/script-farewells.rpy
@@ -399,7 +399,7 @@ label bye_going_to_sleep:
             m "Are you going to sleep, [p_nickname]?{fast}"
 
             "Yeah.":
-                m 1eka "I'll be seeing you in your dreams."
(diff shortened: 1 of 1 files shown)


Booplicate:
1. I don't think this line needs to go, I think it's fine
2. This is a generic goodbye option and it's more or less intended to be short, opposed to more specific options where we she can comment more

lloyd99901:
You bring a good point on how the generic goodbye is meant mostly to be short & simple. I didn't really think of that.

PR #7889, opened 2021-09-23
some 922 fixes

jmwall24:
## Issues

- Going on a date before `mas_bday_spent_time_with` triggers and returning between midnight and 3am on 9/23 results in the player seeing `mas_bday_spent_time_with` dialogue twice due to `EV_ACT_QUEUE` being delayed before running, causing the `rmallEVL()` catch in `mas_bday_spent_time_with` to have nothing to catch. Instead we will change this to `EV_ACT_PUSH` to ensure it's in the `.event_list` at the time the `rmallEVL()` runs. Thanks to @Olliesama for the catch.
- Going on a single date on 9/22 and not returning until after 9/23 between midnight and 3am (and thus never seeing the party) causes normal 922 return from date dialogue, complete with `mas_bday_spent_time_with_wrapup` dialogue, but since the party was never triggered (and assuming no other bday events were completed to satisfy `mas_recognizedBday()`) `mas_bday_postbday_notimespent` is also triggered right after, creating a conflict and continuity error. This adds a `rmallEVL()` to `mas_bday_spent_time_with_wrapup` to correct this issue.
- Playing 'Happy Birthday' on the piano after midnight leads to a bit of inconsistent dialogue where she doesn't seem to realize you're still playing it for her birthday.  To address this `mas_isMonikaBirthday()` has been adjusted to use an `extend` parameter that takes in an `int` for the number of hours we want to extend birthday dialogue or anything else. We now extend the `win` and `fc` paths of 'Happy Birthday' by 1 hour so she recognizes why you are playing Happy Birthday until 1am on 9/23. `mas_bday_pool_happy_bday` now extends until 1am on 9/23 as well and `mas_bday_postbday_notimespent` doesn't start until 1am to accommodate this.

## Testing

- Recreate the above scenarios and ensure the issues are fixed

## TODO

- [ ] change update script version


Legendkiller21 (dismissed):

Monika After Story/game/script-holidays.rpy
```diff
@@ -6899,22 +6899,35 @@ image mas_bday_balloons = MASFilterSwitch(
...
         IN:
-            _date - date to check. If not passed in, we use today.
+            _datetime - datetime to check. If not passed in, we use now.
+            _extend - int of how many hours we want to extend past 922
+                defaults to 0
         """
-        if _date is None:
-            _date = datetime.date.today()
+        if _datetime is None:
+            _datetime = datetime.datetime.now()
```
Booplicate, line 6912: Shouldn't we use `datetime.datetime.utcnow`?

ThePotatoGuy: using localtime for bday checking makes sense since the time that the player sees is more appropriate than their absolute time.

However, this should be a separate function (like `mas_isMonikaBirthday_dt`) since changing this function's input will break submods that use it with an input date.

ThePotatoGuy (changes requested):

Monika After Story/game/script-holidays.rpy
```diff
@@ -6902,19 +6902,47 @@ init -1 python:
...
+            return (
+                is_bday
+                or (
+                    _datetime.month == mas_monika_birthday.month
+                    and _datetime.day == mas_monika_birthday.day + 1
+                    and _datetime.hour < _extend
+                )
+            )
+
+        return is_bday
```
ThePotatoGuy, line 6944:
```suggestion
        return (
            _datetime.month == mas_monika_birthday.month
            and _datetime.day == mas_monika_birthday.day
            and (_extend <= 0 or _datetime.hour < _extend)
        )
```

can make this simpler

Monika After Story/game/script-holidays.rpy
```diff
@@ -7150,7 +7178,7 @@ init 5 python:
             action=EV_ACT_UNLOCK,
             rules={"no_unlock": None},
             start_date=mas_monika_birthday,
-            end_date=mas_monika_birthday + datetime.timedelta(days=1),
+            end_date=datetime.datetime.combine(mas_monika_birthday+datetime.timedelta(days=1), datetime.time(hour=1)),
```
ThePotatoGuy, line 7181: just to confirm, this is because we want Monika's bday to end an hour beyond midnight, correct?

Monika After Story/game/script-holidays.rpy
```diff
@@ -6902,19 +6902,47 @@ init -1 python:
     def mas_isMonikaBirthday(_date=None):
         """
         checks if the given date is monikas birthday
-        Comparison is done solely with month and day
-
+        converts date to datetime and runs thru mas_isMonikaBirthday_dt
```
ThePotatoGuy, line 6905: undo this change, the description here should be explicit on how the comparison is done, and mentioning its a passthrough is only if the function documentation is too large to copy.

Monika After Story/game/script-holidays.rpy
```diff
@@ -6902,19 +6902,47 @@ init -1 python:
...
-            _date.month == mas_monika_birthday.month
-            and _date.day == mas_monika_birthday.day
+        _datetime = datetime.datetime.combine(_date, datetime.time())
+
+        return mas_isMonikaBirthday_dt(_datetime=_datetime)
+
+
+    def mas_isMonikaBirthday_dt(_datetime=None, _extend=0):
+        """
+        checks if the given date is monikas birthday
```
ThePotatoGuy, line 6919:
```suggestion
        checks if the given date is monikas birthday.
        Takes hours beyond the date into account via the `_extend` param.
```
better function doc

ThePotatoGuy (changes requested):

Monika After Story/game/script-holidays.rpy
jmwall24: correct

ThePotatoGuy (approved):

Monika After Story/game/script-holidays.rpy
```diff
@@ -6903,16 +6903,34 @@ init -1 python:
...
+        if _datetime is None:
+            _datetime = datetime.datetime.now()
+
         return (
-            _date.month == mas_monika_birthday.month
-            and _date.day == mas_monika_birthday.day
+            _datetime.month == mas_monika_birthday.month
+            and _datetime.day == (mas_monika_birthday.day + int(_extend > 0))
+            and (_extend <= 0 or _datetime.hour < _extend)
         )
```
Booplicate, line 6934: I'm thinking of a bit more clear way, which would also allow us to use floats for hours (like `1.5` because just 1h is kind of short, or say half an hour for other cases).

```py
moni_bd_start = datetime.datetime.combine(mas_monika_birthday, datetime.time())
moni_bd_start = moni_bd_start.replace(year=datetime.date.today().year)

moni_bd_end = moni_bd_start + datetime.timedelta(days=1, hours=_extend)

return moni_bd_start <= _datetime < moni_bd_end
```

I'd also probably rename `_extend` into maybe `extend_by`.

Monika After Story/game/updates.rpy
```diff
@@ -401,6 +397,41 @@ label v0_12_5(version="v0_12_5"):
             )
     return
 
+# 0.12.5
+label v0_12_5(version="v0_12_5"):
```
ThePotatoGuy, line 401: update scripts need to be moved

ThePotatoGuy (changes requested):

Monika After Story/game/updates.rpy
ThePotatoGuy: oh nvm seems ur merge was off ?

ThePotatoGuy (approved):

jmwall24:
Going to bump this one more time, adding hi priority so we for sure get it in next rel

ThePotatoGuy:
ok this will be the last push back - we're for suring this in after next rel.

ThePotatoGuy (approved):

Legendkiller21 (dismissed):

Legendkiller21 (approved):

Booplicate (approved):

PR #7965, opened 2021-09-29
Fix for followed sprites

### LINKED ISSUE #7050: [Bug] - Enhanced Idle seems to sometimes rely on *current* window instead of MAS window.
I opened a window that was taking up half of the screen over the MAS window and saw that if I went to the side of that window (but still had my cursor over the MAS window) that Monika's eyes would look to that side. I then tried to reproduce this and would only be able to get it to happen very occasionally. 

I've provided a sample image of how I had my windows setup and where my mouse was. In my case, she was looking to the right side of the screen when my cursor was at the same position as the large red dot that I put in the image.
image

Booplicate:
Fixes #7050, this time for real!

### Changes:
- implemented `mas_getWindowTitle`, returns current title set by RenPy Including all possible prefixes/suffixes it adds
- `_window_subtitle` and `config.menu_window_subtitle` are set to a 0-width char to make the title more unique as an attempt to remove false-positives

### DIFF
--- a/Monika After Story/game/definitions.rpy
+++ b/Monika After Story/game/definitions.rpy
@@ -5465,6 +5465,15 @@ init -985 python:
         """
         return store.mas_globals.tt_detected
 
+    def mas_getWindowTitle():
+        """
+        Returns current windows title set by RenPy
+
+        OUT:
+            str
+        """
+        return renpy.game.interface.window_caption
+
 init -101 python:
     def is_file_present(filename):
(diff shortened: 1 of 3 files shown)


Booplicate:
Damn, my 1st PR to `master`... I'm getting old.

ThePotatoGuy (approved):

Legendkiller21 (approved):

PR #7979, opened 2021-10-02
Dont apply outfit mode on selector launch

### LINKED ISSUE #7962: [BUG]: Inconvenient behaviour of the outfit selector
Copy/paste.

Current behaviour:
- When you open the outfit selector and have outfit mode on, it applies it on the current outfit changing acs/hair/etc w/o even you selecting anything. Not big deal since you can press `Reset` and select what you want, but it still is a bit inconvenient.

Expected behaviour:
- Opening the outfit selector shouldn't change the current outfit

ThePotatoGuy:
closes #7962 

# Key Changes
* clothing no longer changes to the exact outfit when opening clothes selector with outfit mode on

# Testing
1. Ensure outfit mode is selected.
2. Change clothes to one that has an outfit (like school uniform) and change an aspect of it so it no longer matches outfit mode. For example, change the hair style of the school uniform so it is not ponytail.
3. Open the clothes selector
4. Verify that no ACS/hair/etc changes when the selector is opened.

### DIFF
--- a/Monika After Story/game/zz_selector.rpy
+++ b/Monika After Story/game/zz_selector.rpy
@@ -3489,14 +3489,6 @@ label mas_selector_sidebar_select(items, select_type, preview_selections=True, o
             store.mas_selspr.SB_VIEWPORT_BOUNDS_BS
         )
 
-        # if in outfit mode, apply the outfit before launching
-        if mailbox.read_outfit_checkbox_checked():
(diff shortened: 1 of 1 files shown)


Booplicate:
Can we make it so if we open the selector and check the checkbox, it'd apply the outfit mode on the current outfit?

PR #7989, opened 2021-10-03
Better (hopefully) previewer

Booplicate:
### Changes:
- General cleanup
- Bugfixes
- Added buttons to copy and paste sprite codes

### Testing:
- Just play around with the exp previewer and verify it still works

### DIFF
--- a/Monika After Story/game/dev/dev_exp_previewer.rpy
+++ b/Monika After Story/game/dev/dev_exp_previewer.rpy
@@ -14,8 +14,8 @@ init 5 python:
     )
 
 label dev_exp_previewer:
-    # call a screen
-    show monika at t21
+    # show monika at t21
+    hide monika
     window hide
 
     $ HKBHideButtons()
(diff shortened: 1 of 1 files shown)


Monika After Story/game/dev/dev_exp_previewer.rpy
```diff
@@ -117,8 +123,8 @@ init 999 python:
             5: ("def", "def")
         }
 
-        # image name map
-        IMG_NMAP = {
+        # Map between sprite code letters and image names
+        IMG_NAMES_MAP = {
```
ThePotatoGuy, line 127: this should use the sprite map loaded from the decoder.

ThePotatoGuy (changes requested):

Monika After Story/game/dev/dev_exp_previewer.rpy
Booplicate: Done!

PR #7984, opened 2021-10-03
New o31 outfit

ThePotatoGuy:
https://www.youtube.com/watch?v=-rokG9XS37w

# TODO
- [ ] clothes select dlg
- [ ] hair select dlg
- [ ] ACS select dlg
- [ ] clothes thumb
- [ ] hair thumb
- [ ] ACS thumb
- [ ] reaction dlg if desired

# Key Changes
* adds a new outfit. comes with: 
    * front-facing bow
    * straight bangs

# Testing
* verify outfit looks good
* **NOTE** outfit mode will **not** set the hair+bow yet. I'm debating on #7983 first. Therefore this will stay DONT MERGE for a bit. (if that issue is moved to after o31, I will add the prog points for outfit mode).

PR #8007, opened 2021-10-06
Hair fixes and stuff

Booplicate:
This contains sprite changes from #6915, so we can test/merge these quickly.

### Changes:
- removed unused assets
- updated hair sprites with a more natural color
- added `orcaramelo_bunbraid` to where it belongs to

## Steps to convert existing sprites (will be copied over to the psd wiki)
1. Export existing hair sprites to individual PNGs 
2. Open the hair sprite PNGs as layers in `new_hair_layer.xcf`
3. Export each hair sprite using the `USE AS MASK` layer as a clipping/layer mask


multimokia (dismissed):

ThePotatoGuy (changes requested):
Couple of things:
* What is the process used to recolor existing hair sprites to match these?
* I need an original size PSD of the primary hairs (def, bun, down) to update the PSDs
* There is a change to hair down front in #7984 that should be made here. I'll probably do this later.

ThePotatoGuy (approved):

multimokia (approved):

PR #8019, opened 2021-10-09
Adding More Compliment Reactions / Quips

KiddTheManiac:
Added a few extra quips to the general thanking_quips, spending time quips, and a few other sections.


KiddTheManiac:
Also, I'd changed the other_quips value/reference to be other_outfit_quips, to make it more specific.

PR #8020, opened 2021-10-09
Random Topic: Other Games With Monika

AmbienceY:
The aim of this addition is to expand on the existing "Other Games" topic concerning Monika possibly entering other games with the player. Conversations would include both general game genres and specific titles (i.e., Minecraft). Feedback regarding dialogue matching Monika's unique character traits would be appreciated for future contributions.

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -17557,3 +17557,36 @@ label monika_ddlcroleplay:
     m 1euu "In any case, if it's encouraging more people to try their hand at writing, I don't think I can really fault it."
     m 1kub "Just make sure to remember that those versions of me are just stories, ahaha~"
     return
+    
+    init 5 python:
+    addEvent(
+        Event(
+            persistent.event_database,
+            eventlabel="monika_gameswithyou4",
+            category=["games"],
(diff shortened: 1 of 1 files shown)


fromulus:
To be completely honest with you, I believe it had been decided among the devs and contributors to avoid video-game related topics (and especially ones that go into details or ones about specific games) due to being out of character and kind of out of place (Monika expresses not being much into games, so it wouldn't make sense for her to start talking about random games, even ones as well known as Minecraft).

My point being, unless Monika has more interesting stuff to say on such a topic beyond just the game(s) itself (see the boardgames history topic which has her go about how games have traditionally been used as tools for various things aside from just entertainement, for example), I'm afraid topics such has these wouldn't have much of an interest aside from just giving a few people the satisfaction of having a game they like being acknowledged in the mod, and ultimately bring nothing but bloat to the mod, kind of like how we have so many language greetings and love songs that they completely lost any sense of novelty they might originally have had, imo.

Case in point, I kind of feel like your topic is a bit of a succession of random references to the game that are meant for people who actually played it, but without any real point to it nor any explanation to people who don't know anything about Minecraft (they exist).

Imagine reading this from the perspective of someone who has no idea what an Enderman even is, they'll end the topic with more questions than answers.

PR #8080, opened 2021-10-27
[Currently under development in different PR] More Variety to Monika's Responses (Update script-moods.rpy and script-brbs.rpy)

WiiBii2:
Okay, so this is a very early version of the project I'm working on. The idea is to add more variety to the way Monika responds to the player when they select an option. For everything. Currently, I'm working on the 'brb' selections and how they might affect Monika's response to certain 'I'm feeling' selections. For instance, if you selected 'brb, I'm going to do some homework, there are more possibilities for Monika's responses. Also, if the player subsequently decides to tell Monika that they are 'feeling proud', she would comment on the completion of the player's homework rather than the other pre-selected prompts. It might be useful to look at the notes I added in the code for more of a clear idea of what I want to do in the future since this is so early in development. I figured I should probably get the go ahead or at least some feedback before I get much further into the project. Please tell me what you think!

Also, I'm pretty new to coding (haha), so if you have any suggestions to clean up the code or simplify it, I'll try my best but might need some help. Sorry in advance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,8 +45,8 @@ If there are issues with MAS, please file a [bug report](https://github.com/Back
 
 To add a suggestion, visit [this link](https://github.com/Backdash/MonikaModDev/issues/new?labels=suggestion&body=Your%20suggestion%20goes%20here&title=%5BSuggestion%5D%20-%20)
 
-### Other Help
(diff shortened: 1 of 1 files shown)


WiiBii2:
> > Would you like me to split this into topic by topic (Homework responses and then the responses with moods) or by rpy file by rpy file (full brb responses and full mood responses)?
> 
> Not as much by "rpy files", as by systems you're changing. I'd say moods go in one PR, brbs in another PR.

PR #8110, opened 2021-10-30
No longer loop through multiple generic gifts

jmwall24:
Currently, if a player improperly installs sprite packs (or gets gift names wrong) and gifts multiple at once, they have to sit through as many generic reacts as they gifted. With as many sprite packs as we now have, this can be a LONG time. So instead we will group them all into a single reaction, saving the player a lot of time before they can try again. 

This is done by editing the generic part of `build_gift_react_labels()` to only add a single reaction, while also deleting all generic gifts in question.

## Testing

- Gift a single generic gift, gift multiple generic gifts, gift generic gifts alongside valid gifts and ensure the flows work properly and make sense

- Also test d25 gifting to make sure that scenario works as well.

### DIFF
--- a/Monika After Story/game/zz_reactions.rpy
+++ b/Monika After Story/game/zz_reactions.rpy
@@ -1066,11 +1066,10 @@ label mas_reaction_gift_generic:
     return
 
 label mas_reaction_gifts_generic:
-    m 1esd "[player], are you trying to give me stuff?"
-    m 1eksdla "I found them, but I can't bring them here..."
-    m 3rksdlb "I can't seem to read them well enough."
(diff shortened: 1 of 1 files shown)


ThePotatoGuy (approved):
just a reminder, d25 has its own set of file react processing, so that should be tested as well. Based on the code, though, I don't think these changes would break that.

lunulae (commented):
Tested for non-d25 gifts. Works for both all-invalid gifts and a mix. (I used roses and chocolates, so outfits might still need checking)

Booplicate (approved):

Legendkiller21 (approved):

PR #8111, opened 2021-10-30
prevent using unstable persistents in stable + fix hang on close

### LINKED ISSUE #8086: Dealing with unstable to stable
reverting back from unstable used to be supported, but after #6913, this is no longer the case.

Since update scripts run in unstable versions, the potential for data corruption is much higher for unstable users, and the only safe window for reverting to stable is when unstable and stable are on the same version, which is something that cannot be easily determined by non-technical users.

So we need to think of ways to address this. Here is a list of the primary issues that need to be addressed:
1. Reverting to an unstable using the checkbox should be either limited to non-breaking situations or made possible to do without breaking the game.
2. Running a stable version with an unstable persistent should not cause game-breaking issues or data corruption.

Here's some potential solutions I've thought of:
1. Lock down unstable checkbox when a non-revertable version is loaded.
    * HOW - we can programmatically determine a revertable version by checking the version numbers. Non-revertable unstables will have an extra version number. 
    * GOOD - prevents reverting from unstable to stable
    * BAD - However, this would mean a broken unstable build will screw users until we can push fixes out. Additionally, with our longer main release cycle, reverting can only happen every few months or so.
    * Issue 1 would be handled neatly.
    * Issue 2 would **not** be handled.
2. Add scripts for reverting update scripts.
    * HOW - Would have to add some additional framework to run reverse update scripts. 
    * GOOD - unstables can be safely reverted from via the checkbox - a bit better for users rather then forcing them so stay with unstable for a bit
    * BAD - This goes against part of the spirit of update scripts, which are designed as one-way scripts. Also means additional code to test and maintain per unstable. 
    * Issue 1 would be handled, with some effort per unstable.
    * Issue 2 would **not** be handled, since an unstable reverse script would not be available in a stable build.
3. Prohibit unstable update scripts that will result in a persistent that cannot run in unstable
    * HOW - basically a procedural change to how we do update scripts in unstable. This would be somewhat a reversal back to unstable practices before #6913, where unstable features that relied on persistent changes could not be tested. 
    * GOOD - handles both issues 
    * BAD - I don't know exactly what kinds of unstable updates scripts would work without causing persistent issues in stable, so I don't know if this really means unstable update scripts become useless.
    * Issue 1 would be handled, since reversion should be possible again.
    * Issue 2 would be handled, since the persistent won't be broken for stables.
4. Sandbox unstable changes with an automatic system for restoring data on return to stable
    * HOW - when we do an unstable script, we will need to mark (or manually save) the variables that we change to a separate persistent var. think of this var as a log of the changes. When a stable build is loaded, we can check if the previous version was an unstable of a higher version number, and if so, pull the persistent var and revert the marked variables to their previous values. This process would have to happen earlier than traditional update scripts, but after persistent backups are made - so likely in the 99x init range. 
    * GOOD - handles both issues
    * BAD - requires an entirely new framework to be build. Also needs to be in a stable version *before* regular unstables can be released, so unstable update scripts would have to be postponed (or limited to non-breaking) until the first stable release the framework is released. Additional code and testing per unstable version would also be required.
    * Issue 1 would be handled, since reversion should be possible again.
    * Issue 2 would be handled, since the persistent would be reverted on stables.
5. Solution 1 + don't load persistents that have unstable version numbers in stables
    * HOW - apply the version number logic from solution 1 to the load process for persistents. This would to be something before/during the persistent backup process. We would need to either try to load a persistent that has a safe version OR hardstop with a chibikia screen mentioning the incompatibilities. Also, making an extra persistent backup when changing from a stable to an unstable version might be a good way to almost guarantee a safe persistent can be loaded.
    * GOOD - handles both issues
    * BAD - need some additional -999 init handling (which is already somewhat crowded and dangerous), but no other major issues
    * Issue 1 would be handled via solution 1
    * Issue 2 would be handled by persistent checking early in init.
6. Solution 1 + part of Solution 5 + force update to unstable.
    * HOW - same as solution 1 for checking versions. Loading will have to check persistents, but we don't do backup checking nor attempt to load a stable persistent. Instead, the current persistent will be backed up as special unstable persistent, a fresh persistent is forced, and splash will force the user to update to unstable (probably via some chibika dialogue then the updater screen). When game is relaunched after the update, the special unstable persistent backup is loaded as the persistent. (for all nuances, see my comment here )
    * GOOD - handles both issues, while also making it possible to get an unstable build without actually having the game run normally 
    * BAD - see solution 5. I can't think of any reasons this would be bad otherwise.

If you have any additional ideas, please comment below. I'll add more if I think of anything else.

ThePotatoGuy:
#8086

# Key Changes
* the unstable checkbox is now disabled when the game version is unstable (aka 4-tiered)
    * fleshed out unstable tooltips - including when its disabled
* unstable persistents are now barred from loading in stable builds. Chibika will direct the user to update the game to unstable or restore a compatible persistent. For the full tree of decisions, see below.

**for the steps below, I will use the following terms:**
* **incompatible persistent** - the persistent with an "unstable" version being loaded into a stable build.
* **special persistent** - the temporary backup copy of the incompatible persistent

## Incompatible Loading - When an unstable persistent is loaded in a stable build

1. the incompatible persistent is backed up as the special persistent and the `persistent` is deleted. This makes MAS generate a new persistent.
    1. if the incompatible persistent could not be backed up, or the `persistent` could not be deleted, an exception hardstop is raised in init.
2. once splash is started, flow jumps to chibika dialogue. A flag is set saying that this dialogue was entered.
    1. if the user previously said they would remove RPY files but rpy files still exist - Chibika says there are RPY files in your game directory. Flow jumps to step 6.
    2. if the user previously said they would restore a compatible persistent - Chibika says that the persistent is still incompatible. Flow jumps to step 4.
    3. if the user previously tried to update, but the 1st updater screen failed - Chibika says that we're going to try updating again. Flow jumps to step 5.
    4. if the user previously tried to update, but the 2nd updater screen failed - Chibika says that the updater afile dto update MAS. Flow jumps to step 5.
3. Chibika says hello and a menu pops up with the following choices:
    1. Choice 1 - user asks what happened. Chibika explains that the persistent is incompatible, then flow jumps to step 4.
    2. Choice 2 - user says launch the updater (aka, the user already knew this was going to happen and just needs to launch the updater). Flow jumps to step 5.
4. Chibika asks the user what they would like to do:
    1. Choice 1 - update MAS. Flow jumps to step 5.
    2. Choice 2 - restore to a compatible persistent. Chibika says which folder to copy to and to delete the special persistent`. **Game quits.** When game is re-launched, continue at **Normal loading - step 1**
5. gamedir is checked for RPY files.
    1. If RPY files exist - Chibika says there are RPY files in your game directory. Flow jumps to step 6.
    2. no RPY files exist - Flow jumps to step 7.
6. Chibika asks user if they would like her to delete RPY files.
    1. Choice 1 - Yes, delete. - Chibika attempts to delete RPY files.
        1. If Chibika could delete RPY files - Flow jumps to step 7.
        2. if Chibika was unable to delete RPY files - Chibika says the user has to delete them manually. **Game quits.** When game is re-launched, continue at **Normal loading - step 1**.
    2. Choice 2 - No, don't delete - Flow jumps to 4.ii (Chibika says to restore compatible persistent).
7. `update_now` label is called - the updater's first screen is shown. A number of scenarios can occur:
    1. update folder doesn't exist or couldn't be moved - this shows an error menu that tells the user what to do. Flow jumps to step 8.
    2. updater could not find a new update - Flow jumps to step 8.
    3. user hits cancel when a new update is found - Chibika tells the player not to cancel. Flow jumps back to step 4.
    4. user hits cancel because the updater had connection issues/timedout/etc... - Flow jumps to step 8.
    5. user hits update - Flow jumps to step 9.
8. This assumes the updater failed. Chibika explains that the updater failed, and suggests fixing issues before trying again. **Game quits.**
9. The real updater screen is shown. Regardless of what happens, the game will quit after this. Continue at the **Normal loading - step 1**

## Normal loading

1. On load, the existence of the special persistent is checked:
    1. if the persistent could not be loaded because it was corrupted in some way - the special persistent is deleted. Flow jumps to step 2.
        1. If the special persistent could not be deleted, a hardstop exception is raised in init.
    2. if the special persistent could be loaded:
        1. if the version is compatible with the current build - restore the special persistent backup. Continue at step 2.
            1. If the special persistent could not be properly restored - a hardstop exception is raised in init.
        2. otherwise - log an error and setup vars for **Incompatible Loading**. Continue at step 2.
    3. Otherwise - Continue at step 2.
2. the current persistent's data is loaded
    1. If loading the persistent fails:
        1. If the special persistent was found - persistent load logic quits early. Continue at **Incompatible Loading - step 2**
        2. Otherwise - normal load behavior occurs
    2. if the current persistent is incompatible - treat the current persistent as the actual special persistent. If we found a special persistent earlier, it is overwritten. Continue at **Incompatible Loading - step 1**
    3. Otherwise - the current persistent is compatible:
        1. if the special persistent was found and the Chibika dialogue flag is **not** set (**Incompatible Loading - step 2**) - the special persistent is deleted because it probably was not created by us. normal load behavior occurs.
            1. If the special persistent could not be deleted - a hardstop exception is raised in init
         2. If the special persistent was found and the Chibikia dialogue flag is set - continue at **Incompatible Loading - step 2**
         3. Otherwise - normal load behavior occurs

# Other changes
* **PROBABLY FIXED the hanging on close** - turns out it was the updater threads not being daemonized. Under normal circumstances, the game won't hang, but if your connection to the update servers is really bad for some reason, then the threads would hang. If the threads aren't daemonized, then they have to stop for the entire program to stop, even if the main thread quits.
* a ton of modular code in `mas_earlyCheck` has become functions. Also several new APIs in the `mas_per_check` store:
    * `tryper` - tries to read a persistent, throwing exceptions. this will return a tuple of True, and either the persistent data or the version number
    * `is_version_compatible` - very specific check for compatibility between a persistent version and a game version. Don't use this for any other checking.
    * `is_per_bad` - checks if the persistent is in a bad state, either corrupted or incompatible - based on the early checks
    * `is_per_corrupt` - checks if the persistent is corrupt - based on the early checks
    * `is_per_incompatible` - checks if the persistent is incompatible - based on the early checks
    * `no_backups` - checks if no persistent backups could be found - based on the early checks
    * `should_show_chibika_persistent` - True if we should show the chibika persistent dialogue
    * `wraparound_sort` - sorts a list of numbers, applying wraparound rules so that single digit numbers are considered larger than 99
* the updater screen now returns more values based on the button the user clicked. The `update_now` label also returns those values.
    * see the `MASUpdaterDisplayable.RET_VAL_*` constants
* added some insensitive sprites for the fancy checkboxes
* moved version list checking functions to 0utils (aka early)
* few new version-based functions (all in `mas_utils`):
    * `_get_version_nums` - converts a version string into version number list
    * `is_ver_stable` - True if a version is stable (aka a 3-tiered version number)
    * `_is_downgrade` - checks if a version transition is a downgrade - basically a wrapper around `compareVersionLists` that accepts version strings instead of version lists.

# Testing

### prepare an incompatible persistent
1. Launch the game normally
2. open console and set `persistent.version_number` to `0.12.4.1` (or higher, but the version MUST be 4-tiered). Could also try with the full unstable text, like `0.12.4.1-unstable.2021.12.10.160`.
3. close game normally
4. **make a copy of the persistent** - You should **always** use this when a testing instruction says to set an incompatible persistent.

### prepare a fresh, compatible persistent
1. Make a copy of a compatible persistent - you should **always** use this when a testing instruction to says start with a normal, valid persistent.

### Other setup
* be prepared to edit code directly - particularly `0config.rpy` and `zz_backup.rpy`. You will need to fake exceptions as well - add `raise Exception("fail")` to fake an exception.
    * I have added `TEST_CASE_<Letter>` to various try/excepts in `zz_backup.rpy` to make it easier to know which code to add fake exceptions. **Always put the exception in the `try` block.**
* **use a stable build for your tests** - this means the update folder must be from a stable version (version that does not have `unstable` tag)
* make a copy of the entire install before starting. Some of the workflow will delete RPY files so you may need to compile RPY changes separately if you need to.
* prepare a way to prevent an RPY file from being deleted. An easy way to do this is open a py console in the game dir and run `abc = open("<filename>")` - this will keep an open handle on the file. **NOTE** - leaving the file open in a notepad will **not** work.

## Loading Cases
This is split into 2 sections, one for the Incompatible loading, one for the normal loading. Because of the large number of branches, each testable branch will be listed in a dot notation, (i.e: `1.i.a`), with the steps to achieve that branch, as well as verification steps afterward.

### Incompatible Loading

> `1.i`
1. set the incompatible persistent as `persistent`
2. add a fake exception to `TEST_CASE_F`
3. Verify that the game crashes with a traceback explaining what to do

> `2.i`
1. follow steps in `6.i.b`
2. Don't actually delete the RPY files
3. Verify that when starting the game again, Chibika says that there are still RPY files in your game dir.
4. Verify that the option for Chibika to delete the files is shown.

> `2.ii`
1. follow steps in `4.ii`
2. Don't actually change the persistent (aka, leave it incompatible).
3. Verify that when starting the game again, Chibika says the persistent is still incompatible.
4. Verify that the Update MAS vs restore persistent menu options are shown

> `2.iii`
1. follow steps in `7.iv`
2. Verify that when starting the game again, Chibika says that we'll try updating again.
3. Verify that the updater is ran again.

> `2.iv`
1. follow steps in `7.v`, but hit cancel on the 2nd updater screen
2. Verify that when starting the game again, Chibika says that the updater failed to update MAS.
3. Verify that the updater is ran again.

> `4.ii`
1. set the incompatible persistent as `persistent`
2. select "What Happened" dialogue option
3. select "restore to a compatible persistent" dialogue option
4. Verify that Chibika has some instructions on restoring, then the game quits

> `5.i`
1. use the install that has RPY files.
2. set the incompatible persistent as `persistent`
3. select dialogue to get to the updater path
4. Verify that Chibika says you cannot update the game because there are RPY files

> `6.i.a`
1. follow steps in `5.i`
2. when Chibika asks if she can delete your RPY files, pick **Yes**
3. Verify that the py console is shown and all the RPY files are deleted
4. Verify that the updater then runs (step 7)

> `6.i.b`
1. follow steps in `5.i`
2. before letting Chibika delete your RPY files, keep an open handle on an RPY file (aka prevent one from being deleted).
3. when Chibika asks if she can delete your RPY files, pick **Yes**
4. Verify that Chibika says she couldn't delete all of the RPY files, and that you will need to delete them manually. Then the game quits.

> `6.ii` 
1. follow steps in `5.i`
2. when Chibika asks if she can delete your RPY files, pick **No**
3. Verify that Chibika suggests that you should restore the persistent, shows instructions on restoring, then the game quits.

> `7.i`
1. take the `update` folder out of the root `DDLC/` dir and the `game/` dir.
2. set the incompatible persistent as `persistent`
3. select dialogue to get to the update path
4. Verify that the updater shows the `updater folder` error message
5. Verify that Chibika says the updater failed and then the game quits

> `7.ii`
1. set the incompatible persistent as `persistent`
2. once chibika appears, open the console and set `mas_updater._forced_updater_start_state = MASUpdaterDisplayable.STATE_UPDATED`
3. select dialogue to get to the update path
4. Verify that the updater says that no new update was found
5. Verify that Chibika says the updater failed and then the game quits

> `7.iii`
1. set the incompatible persistent as `persistent`
2. select dialogue to get to the update path
3. Assuming you don't have connection issues, the first updater screen will find a new update. **Click Cancel**
4. Verify that Chibika says not to cancel
5. Verify that the Update MAS vs restore persistent menu options are shown

> `7.iv`
1. set the incompatible persistent as `persistent`
2. once chibika appears, open the console and set `mas_updater._forced_updater_start_state = MASUpdaterDisplayable.STATE_TIMEOUT`
3. select dialogue to get to the update path
4. Verify that the first updater screen says connection timedout
5. Click cancel
6. Verify that Chibika says the updater failed and then the game quits

> `7.v`
1. set the incompatible persistent as `persistent`
2. select dialogue to get to the update path
3. Click Update when the updater finds a new update
4. Verify the 2nd updater screen appears and the game quits regardless of the actions taken on this screen.

### Normal Loading

> `1.i` 
1. start with a normal, valid `persistent`
2. add a dummy file called `persistent_unstable` to the persistents folder
3. Verify that the `persistent_unstable` file is deleted and the game runs normally

> `1.i.a`
1. add a dummy file called `persistent_unstable` to the persistents folder
2. add a fake exception to `TEST_CASE_A` and `TEST_CASE_B`
3. Verify that the game crashes with a traceback explaining what to do

> `1.ii.a.a`
1. set `persistent_unstable` to a normal, valid `persistent`
2. add a fake exception to `TEST_CASE_C`
3. Verify that the game crashes with a traceback explaining what to do

> `2.i.a`
1. set the incompatible persistent as `persistent_unstable`
2. add a fake Exception to `TEST_CASE_D`
3. Verify that the Chibika dialogue for incompatible persistents starts when the game is launched

> `2.ii`
1. set the incompatible persistent as both `persistent_unstable` and `persistent`
2. Verify that the Chibika dialogue for incompatible persistents starts when the game is launched
3. Additionally, if you add a fake exception to `TEST_CASE_F`, then the Incompatible Loading test branch `1.i` occurs

> `2.iii.a`
1. start with a normal, valid `persistent`
2. set the incompatible persistent as `persistent_unstable`
3. Verify that the game runs normally.
4. Verify that `persistent_unstable` was deleted

> `2.iii.a.a`
1. start with a normal, valid `persistent`
2. set the incompatible persistent as `persistent_unstable`
3. add a fake exception to `TEST_CASE_E`
4. Verify that the game crashes with a traceback explaining what to do

> `2.iii.b`
1. follow Incompatible Loading test branch `7.v` so the persistent has appropriate flags set. The incompatible persistent should be already created.
2. Verify that the Chibika dialogue for incompatible persistents starts when the game is launched

> `2.iii.c`
1. this is the normal behavior test case. Verify the game launches normally if no persistent problems occur.

## successful unstable update test case

1. Follow Incompatible Loading test branch `7.v`
2. change the MAS version (in `0config.rpy`) to match the persistent version you set for the incompatible persistent
3. Launch the game
4. Verify that `persistent_unstable` was restored to `persistent` - you can check this by checking that the following persistent values are all false:
    * `_mas_incompat_per_forced_update`
    * `_mas_incompat_per_forced_update_failed`
    * `_mas_incompat_per_user_will_restore`
    * `_mas_incompat_per_entered`

## unstable checkbox test cases

> new warning message
1. Verify when enabling unstable, the warning message says its not easily reversible

> checkbox locked in unstable version
1. open console and change `config.version` to an incompatible version - like `0.12.4.1`
2. open the settings menu
3. Verify that the unstable checkbox is selected, grayed out, and unclickable. There will also be a tooltip explaining why its disabled.

## hang on close

1. launch the game and immediately close it via talk > goodbye - the goal is to try and quit before the updater check runs.
2. verify the game doesn't hang for more than few seconds at most.

## Other test cases
* verify persistent corruption stuff still occurs. This includes:
    * `persistent` is corrupted and working backups were found 
    * `persistent` is corrupted and no backups found - chibika dialogue triggers
* verify the updater still works


jmwall24 (approved):

Legendkiller21 (approved):

Monika After Story/game/0utils.rpy
```diff
@@ -647,6 +703,56 @@ python early in mas_utils:
...
+        """
+        if nums_only:
+            ver_nums = ver_str.partition("-")[0]
+        
+            if split_nums:
+                return ver_nums.split(".")
+
+            return ver_nums
+
+        return ver_str
```
Booplicate, line 724: I'm being picky, but this is kind of odd function.

You don't seem to be using `nums_only` as `False` at all. But I also don't think it'd ever make sense to as it'll just return the string you passed in. So I think it'd be cleaner if we just removed it.

I think we should also document the `split_nums` param because it changes the output dramatically (you can get either a `str` or a `list` of `str`s.

Monika After Story/game/0utils.rpy
```diff
@@ -647,6 +703,56 @@ python early in mas_utils:
...
+        IN:
+            from_ver_str - starting version (as ver str)
+            to_ver_str - ending version (as ver str)
+
+        RETURNS: true if downgrade, False if not
+        """
+        return compareVersionLists(
+            _get_version(from_ver_str, True, True),
+            _get_version(to_ver_str, True, True)
+        ) > 0
```
Booplicate, line 753: I don't like this and think it's incorrect. Yes, we get valid output from this function, but it's really unclear and may change in future python versions. The problem is `_get_version` returns a `list` of **`str`**. But `compareVersionLists` uses comparison operators `>` and `<`. We should convert output of `_get_version` into `int` before passing them into `compareVersionLists`.

Can do it like `map(int, ver_list)`.

Booplicate (changes requested):
We should also check if we have `rpy` files before starting updating. Maybe try to delete them ourselves or just prompt the player to delete them (like we already do if you update from the game menu).

ThePotatoGuy:
@Booplicate @jmwall24 @Legendkiller21 updated with dialogue where Chibika can try to delete RPY files. The incompatible loading test cases have been updated.

Monika After Story/game/0utils.rpy
```diff
@@ -703,25 +703,16 @@ python early in mas_utils:
...
-        if nums_only:
-            ver_nums = ver_str.partition("-")[0]
-        
-            if split_nums:
-                return ver_nums.split(".")
-
-            return ver_nums
-
-        return ver_str
+        return map(int, ver_str.partition("-")[0].split("."))
```
Booplicate, line 715: I think we should explicitly convert into a `tuple` here, in py3 `map` returns a `Generator`.

Booplicate (changes requested):

Monika After Story/game/0utils.rpy
ThePotatoGuy: made it a list because compareVersionLists says lists explicitly.

jmwall24 (approved):

Booplicate (approved):

Legendkiller21 (approved):

multimokia (approved):

Issue #8140, opened 2021-11-01
Can't add Custom Piano songs 'MASNewlineLogAdapter' object has no attribute 'warn'


Koplo3:
Found this error just after reinstalling MAS on a new computer

wanted to add custom song `dokidokiforever.json` which I got [here](https://github.com/Monika-After-Story/MonikaModDev/issues/5047) with some edit here and there

Here's the particular code
```
{
    "name": "Doki Doki Forever",
    "end_wait": 5,
    "verse_list": [0, 8, 19, 27, 37],
    "pnm_list": [
        {
            "_comment": "1hub/1hua: happy, eyes closed; 1ekd/1ekc: sad; 1eub/1eua: regular; Verse one",
            "text": "Hey, hey, my heart's beating when I'm hanging out with you",
            "style": "monika_credits_text",
            "notes": [
                "A5",
                "F5SH",
                "D5",
                "E5",
                "E5",
                "D5",
                "E5",
                "D5",
                "E5",
                "D5",
                "E5",
                "F5SH",
                "B4"
            ],
            "express": "1hub",
            "postexpress": "1hua",
            "verse": 0
        },
        {
            "text": "Why does my heart ache when I hear you feel the same way too?",
            "style": "monika_credits_text",
            "notes": [
                "A4",
                "B4",
                "D5",
                "A5",
                "F5SH",
                "E5",
                "D5",
                "E5",
                "D5",
                "E5",
                "D5",
                "E5",
                "F5SH",
                "E5",
                "D5"
            ],
            "express": "1ekd",
            "postexpress": "1eka",
            "verse": 0
        },
        {
            "text": "Just like a sundae, it's sweet every time I teach you something new",
            "style": "monika_credits_text",
            "notes": [
                "A4",
                "B4",
                "D5",
                "A5",
                "F5SH",
                "D5",
                "E5",
                "E5",
                "D5",
                "E5",
                "D5",
                "E5",
                "D5",
                "E5",
                "F5SH",
                "B4"
            ],
            "express": "1hub",
            "postexpress": "1hua",
            "verse": 0,
            "ev_timeout": 5.0
        },
        {
            "text": "Is this by chance or fate?",
            "style": "monika_credits_text",
            "notes": [
                "A4",
                "A4",
                "B4",
                "D5",
                "A5",
                "F5SH"
            ],
            "express": "5hub",
            "postexpress": "5hua",
            "verse": 0
        },
        {
[...]

Issue #8133, opened 2021-10-31
Scary Story Bug


DistraughtEngineer:
labels=bug&body=Describe%20bug%20and%20steps%20for%20reproduction%20here&title=%5BBug%5D%20-%20)
New to the GitHub, just received this bug after trying to get a new scary story:

I'm sorry, but an uncaught exception occurred.

While running game code:
  File "game/script-ch30.rpy", line 1518, in script call
  File "game/event-handler.rpy", line 2530, in script call
  File "game/script-stories.rpy", line 136, in script call
  File "game/script-stories.rpy", line 236, in script
  File "game/script-stories.rpy", line 238, in <module>
  File "game/script-stories.rpy", line 116, in get_and_unlock_random_story
IndexError: list index out of range

-- Full Traceback ------------------------------------------------------------

Full traceback:
  File "game/script-ch30.rpy", line 1518, in script call
  File "game/event-handler.rpy", line 2530, in script call
  File "game/script-stories.rpy", line 136, in script call
  File "game/script-stories.rpy", line 236, in script
  File "C:\Users\me\Desktop\Doki Doki Literature Club\renpy\ast.py", line 814, in execute
    renpy.python.py_exec_bytecode(self.code.bytecode, self.hide, store=self.store)
  File "C:\Users\me\Desktop\Doki Doki Literature Club\renpy\python.py", line 1719, in py_exec_bytecode
    exec bytecode in globals, locals
  File "game/script-stories.rpy", line 238, in <module>
  File "game/script-stories.rpy", line 116, in get_and_unlock_random_story
  File "/home/tom/ab/x64lucid-deps/install/lib/python2.7/random.py", line 275, in choice
IndexError: list index out of range

Windows-8-6.2.9200
Ren'Py 6.99.12.4.2187
Monika After Story 0.12.4-unstable.2021.10.28.151

Booplicate:
#8116

ThePotatoGuy:
[closing as fixed]

Issue #8124, opened 2021-10-31
Exception


myoongs:
![mas exception](https://user-images.githubusercontent.com/86926166/139585369-959ebca3-0d62-436c-a7cc-3651c4c4a4f1.JPG)
This has never happened before, only with this new update. When I tried clicking on the "New Story" option for scary stories it did this. It doesn't happen with regular stories.

Booplicate:
#8116

Issue #8121, opened 2021-10-31
So, i wanted to ear a new scary story to get in the mood of the day... and this happen:


MonikaEndDie:
[code]
I'm sorry, but an uncaught exception occurred.

While running game code:
  File "game/script-ch30.rpy", line 1510, in script call
  File "game/event-handler.rpy", line 2530, in script call
  File "game/script-stories.rpy", line 136, in script call
  File "game/script-stories.rpy", line 236, in script
  File "game/script-stories.rpy", line 238, in <module>
  File "game/script-stories.rpy", line 116, in get_and_unlock_random_story
IndexError: list index out of range

-- Full Traceback ------------------------------------------------------------

Full traceback:
  File "game/script-ch30.rpy", line 1510, in script call
  File "game/event-handler.rpy", line 2530, in script call
  File "game/script-stories.rpy", line 136, in script call
  File "game/script-stories.rpy", line 236, in script
  File "C:\Doki Doki Literature Club Monika after story\renpy\ast.py", line 814, in execute
    renpy.python.py_exec_bytecode(self.code.bytecode, self.hide, store=self.store)
  File "C:\Doki Doki Literature Club Monika after story\renpy\python.py", line 1719, in py_exec_bytecode
    exec bytecode in globals, locals
  File "game/script-stories.rpy", line 238, in <module>
  File "game/script-stories.rpy", line 116, in get_and_unlock_random_story
  File "/home/tom/ab/x64lucid-deps/install/lib/python2.7/random.py", line 275, in choice
IndexError: list index out of range

Windows-8-6.2.9200
Ren'Py 6.99.12.4.2187
Monika After Story 0.12.4
[/code]

PavelPlayRo:
yea i get the same error

Booplicate:
#8116

Iceninja64:
I'm also having the same problem when I try to ask for a new story but the good news is that you can reread some of the other stories if you have them unlocked

ThePotatoGuy:
[closing as fixed]

PR #8154, opened 2021-11-01
O31 bug fixes

multimokia:
Bugfix PR, handles issues seen during the o31 event and some other ones. May be added onto in future.

Fixes: #8140, #8116, #8121, #8124, #8133

Renames our `s_lightning` vars to more glance-readable names, background locks it to just spaceroom due to the lack of a guarantee it'll work with other BGs, as well as embeds the doki cares check into the lighting chance for sayori lightning.

Also addresses an issue where outfits were duplicated in the event clothes selector for users with < 1000 affection due to spritepack outfits being added (and then duplicated in the event clothes map)
    - This change forbids spritepacks from being added to that map (no reason they should be in there in the first place)

### DIFF
--- a/Monika After Story/game/script-ch30.rpy
+++ b/Monika After Story/game/script-ch30.rpy
@@ -57,10 +57,10 @@ init -1 python in mas_globals:
     # True means show lightning, False means do not
 
     lightning_chance = 16
-    lightning_s_chance = 10
+    sayori_lightning_chance = 10
     # lightning chances
 
-    show_s_light = False
+    show_sayori_lightning = False
     # set to True to show s easter egg.
 
     text_speed_enabled = False
@@ -1526,16 +1526,16 @@ label ch30_post_mid_loop_eval:
 
         # Thunder / lightning if enabled
         if (
-                store.mas_globals.show_lightning
-                and renpy.random.randint(1, store.mas_globals.lightning_chance) == 1
-            ):
+            store.mas_globals.show_lightning
+            and renpy.random.randint(1, store.mas_globals.lightning_chance) == 1
+        ):
             $ light_zorder = MAS_BACKGROUND_Z - 1
             if (
-                    store.mas_globals.show_s_light
-                    and renpy.random.randint(
-                        1, store.mas_globals.lightning_s_chance
-                    ) == 1
-                ):
+                store.mas_globals.show_sayori_lightning
+                and persistent._mas_pm_cares_about_dokis
+                and mas_current_background.background_id == store.mas_background.MBG_DEF
+                and renpy.random.randint(1, store.mas_globals.sayori_lightning_chance) == 1
+            ):
                 $ renpy.show("mas_lightning_s", zorder=light_zorder)
             else:
                 $ renpy.show("mas_lightning", zorder=light_zorder)
@@ -1814,8 +1814,8 @@ label ch30_reset:
 
     python:
         # name eggs
-        if mas_egg_manager.sayori_enabled() or (mas_isO31() and not persistent._mas_pm_cares_about_dokis):
-            store.mas_globals.show_s_light = True
+        if mas_egg_manager.sayori_enabled() or mas_isO31():
+            store.mas_globals.show_sayori_lightning = True
 
     python:
         # apply ACS defaults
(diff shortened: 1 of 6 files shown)


Monika After Story/game/script-ch30.rpy
```diff
@@ -1526,16 +1526,16 @@ label ch30_post_mid_loop_eval:
...
+        ):
             $ light_zorder = MAS_BACKGROUND_Z - 1
             if (
-                    store.mas_globals.show_s_light
-                    and renpy.random.randint(
-                        1, store.mas_globals.lightning_s_chance
-                    ) == 1
-                ):
+                store.mas_globals.show_sayori_lightning
+                and persistent._mas_pm_cares_about_dokis
```
ibm2431, line 1535: Is there supposed to be a `not` on this doki care check?

multimokia: Yes you're right. Good catch

Monika After Story/game/script-ch30.rpy
```diff
@@ -1526,16 +1526,16 @@ label ch30_post_mid_loop_eval:
...
+        ):
             $ light_zorder = MAS_BACKGROUND_Z - 1
             if (
-                    store.mas_globals.show_s_light
-                    and renpy.random.randint(
-                        1, store.mas_globals.lightning_s_chance
-                    ) == 1
-                ):
+                store.mas_globals.show_sayori_lightning
+                and not persistent._mas_pm_cares_about_dokis
```
Booplicate, line 1535: Maybe we should just disable the eggs if this var is `True`?

Monika After Story/game/script-stories.rpy
```diff
@@ -51,7 +51,10 @@ init -1 python in mas_stories:
 
     #Override maps to have special conditionals for specific story types
     NEW_STORY_CONDITIONAL_OVERRIDE = {
-        TYPE_SCARY: "mas_isO31() or mas_stories.check_can_unlock_new_story(mas_stories.TYPE_SCARY)",
+        TYPE_SCARY: (
+            "(mas_isO31() and mas_stories.has_new_stories_for_type(mas_stories.TYPE_SCARY)) "
+            "or mas_stories.check_can_unlock_new_story(mas_stories.TYPE_SCARY)"
```
Booplicate, line 56: I think this logic could be simplified and we could do less iterations if `check_can_unlock_new_story` could take several params to control what it checks. For example, `check_last_seen`. If `True` (default), we check the last seen of a story of this type, otherwise we skip it and only check if we have any stories.

Then this condition becomes
`mas_stories.check_can_unlock_new_story(mas_stories.TYPE_SCARY, check_last_seen=not mas_isO31())`

Naming is always a subject to change, of course. I'm not too good at this.

Booplicate (changes requested):

jmwall24 (dismissed):

jmwall24 (approved):

Booplicate (dismissed):

Legendkiller21 (dismissed):

Monika After Story/game/zz_selector.rpy
```diff
@@ -3940,18 +3940,22 @@ label monika_clothes_select:
...
+                                not spr_obj.is_custom
+                                and spr_obj != mas_clothes_def
+                                and spr_obj != mas_clothes_blazerless
+                        )
                 ):
                     gifted_clothes.pop(index)
 
             #Now we handle holiday clothes
-            clothes_to_add = persistent._mas_event_clothes_map.get(datetime.date.today())
+            clothes_to_add = ev_outfits
```
ThePotatoGuy, line 3958: `_mas_event_clothes_map` seems to be a mapping of a date to a single clothing item's name, so there's always just 1 outfit here. In that case, a simple `spr_obj.name == ev_outfits` would be enough, but I also would just use `clothes_to_add` instead (so move the var to before the loop instead of creating an extra temp var).

ThePotatoGuy (changes requested):

ThePotatoGuy (approved):

jmwall24 (approved):

Booplicate (approved):

Legendkiller21 (approved):

PR #8155, opened 2021-11-01
[Content] Extended Responses for Script-Brbs.rpy (Brb choices and return dialogue)

WiiBii2:
Finally done after about a week of coding. This is going to be partnered with a future PR I'll open that adds variety to the responses in moods.rpy. It's also possible that I'll add even **more** variations in Monika's responses! Oh, boy! As always, if there are any problems, feel free to leave a comment!

### DIFF
--- a/Monika After Story/game/script-brbs.rpy
+++ b/Monika After Story/game/script-brbs.rpy
@@ -214,7 +214,7 @@ label monika_writing_idle:
             m "Oh! You're going to{fast} go write something?"
 
         else:
-        $ response = renpy.random.randint(1,2)
+        $ response = renpy.random.randint(1,3)
         if response = 1:
             m 1eub "Oh! You're going to go write something?"
             m 1hua "That makes me so glad!"
@@ -230,6 +230,20 @@ label monika_writing_idle:
             #maybe add a variable that counts how many days since monika has last written the player a poem for higher accuracy?
             m 7eusdlb "Anyway, make sure to enjoy yourself!"
             m 1eua "I'll be here waiting when you finish!"
+
+       elif response = 3:
+            if not renpy.seen_label("monika_writing_posture"):
+            call monika_writing_posture
+
+label monika_writing_posture:
+            m *** "Going to get some writing done?"
+            m *** "Sounds good!"
+            m *** "Make sure you're sitting in a way that's comfortable, though."
+            m *** "Yuri had back pain because of her 'reading posture,' remember? Ahaha~"
+            m *** "On a serious note, though, bad posture can cause all sorts of nasty health issues over time."
+            m *** "Take a break every so often to get up and stretch or grab a glass of water!"
+
+            
 
     elif mas_isMoniUpset():
         m 2esc "Alright."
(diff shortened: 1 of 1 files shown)


multimokia:
Please do not put multiple unrelated things into a single PR. It makes reviewing it infinitely more troublesome. 

Additionally, we ask that contributors please test their contributions before submitting them.

Monika After Story/game/script-brbs.rpy
```diff
@@ -110,14 +110,38 @@ label monika_brb_idle:
...
-            m 1eua "Oh alright.{w=0.2} {nw}"
-            extend 3hub "Hurry back, I'll be waiting here for you~"
+            $ goodbye = renpy.random.randint(1,2)
+            if goodbye = 1:
+                m 1eua "Oh alright.{w=0.2} {nw}"
+                extend 3hub "Hurry back, I'll be waiting here for you~"
+
+            elif goodbye = 2:
+                m 1eua "Okay."
+                m 3hub "I'll be here when you're done!"
```
multimokia, line 120: I don't like this approach, it's very much non-scalable and also inefficient at large scales.

I'd suggest using something like MASQuipList to use label quips as it's cleaner and simpler to expand in future

Monika After Story/game/script-brbs.rpy
```diff
@@ -110,14 +110,38 @@ label monika_brb_idle:
         show monika at t11
 
         if _return:
-            m 1eua "Oh alright.{w=0.2} {nw}"
-            extend 3hub "Hurry back, I'll be waiting here for you~"
+            $ goodbye = renpy.random.randint(1,2)
+            if goodbye = 1:
```
multimokia, line 114: invalid syntax. This occurs in all checks.

multimokia (changes requested):

WiiBii2:
> Please do not put multiple unrelated things into a single PR. It makes reviewing it infinitely more troublesome.

I was told by Booplicate to do this way in #8080 . Here's the full quote if you would like it.

> > Would you like me to split this into topic by topic (Homework responses and then the responses with moods) or by rpy file by rpy file (full brb responses and full mood responses)?
> 
> Not as much by "rpy files", as by systems you're changing. I'd say moods go in one PR, brbs in another PR.

WiiBii2:
@Booplicate Is there anything that you would recommend before I redo the code?

multimokia:
Basically keep all related *topics* together. Unrelated topics should be in their own PRs.

So one brb per PR pretty much. It becomes easier to manage and review vs a lot at once.

Booplicate:
Do you need to help with something specific? I agree with @multimokia. Using `MASQuipList` might be the best approach for this. I also mentioned before that you can't use single `=` for comparison. While we test every addition, you should test your code, too.

Booplicate:
You don't have to close old PRs, you can edit them.

WiiBii2:
Okay, haha. Sorry for asking so many questions. I'll try to fix this to the best of my ability.

multimokia:
Going to close this as this is kinda unmaintainable in the current state given how large it is making it difficult to review but also doesn't quite let us easily cut things which may not work too well either.

As well, we highly recommend testing your contributions before submitting them since I can see at a glance already numerous areas which will crash due to a syntax error, this also delays the speed at which we review the contribution since we need to fix all the bugs first.

If you want to re-open in the future, please ensure it's tested, and please split things into different PRs to help make it more manageable.

PR #8167, opened 2021-11-03
Added a "Plush Monika" topic

QuadILOP1:
Just a simple persistent value due to having or not having a Monika Plushie

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -17565,3 +17565,36 @@ label monika_ddlcroleplay:
     m 1euu "In any case, if it's encouraging more people to try their hand at writing, I don't think I can really fault it."
     m 1kub "Just make sure to remember that those versions of me are just stories, ahaha~"
     return
+
+init 5 python:
+    addEvent(
+        Event(
+            persistent.event_database,
+            eventlabel="plush_monika",
+            category=["you"],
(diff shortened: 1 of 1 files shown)


QuadILOP1:
Is anyone gonna check this stuff? It's been five days XD

multimokia:
While it makes sense to bring this topic up, I don't believe this is the direction we wish to take it.

Referring to the plush as a child is a little excessive and will make people uncomfortable for sure. Plus it's a bit of a stretch. Additionally I feel like there can be more commentary especially coming from the current `monika_twitter`, where she mentions being reduced to a marketable asset -- this form of merch especially seems fitting for that. So I feel like a good direction would be to piggy-back this off of that topic by requiring that one to have been seen first, as well as adding some dialogue alluding to it.

multimokia:
Additionally, player model variables should use the prefix `_mas_pm`, your variables only have the `_mas_` part.

They're also not the same variable used in the dialogue, I suggest using `_mas_pm_has_plush_monika` for verbosity.

multimokia:
As of the discussion above, I will be tagging this under rewrite as the direction needs to be changed.

QuadILOP1:
You got it. Just was throwing this in because I really didn't know where I was fully going with it tbh.

multimokia:
Closing as no longer in development (confirmed via DM thru @Mugnika)

PR #8179, opened 2021-11-04
Sleep changes

Alexxx1407:
Adds enamored+ paths to 20-24 and 0-3 hour paths, instead of using the same for everything above normal.

Edit: this has changed in scope a bit due to player concerns that we are too harsh on non-regular sleep schedules. For now we are generalizing fares/greets that concern with sleep so Monika doesn't seem so upset with differing sleep schedules. `monika_timeconcern` is commented out as well. These are temporary measures until we can fully implement a time concern overhaul.

### DIFF
--- a/Monika After Story/game/script-farewells.rpy
+++ b/Monika After Story/game/script-farewells.rpy
@@ -1330,27 +1330,28 @@ label bye_prompt_game:
                     m 2euc "Enjoy your game, then."
                     m 2esd "I'll be here."
 
-    elif mas_isMNtoSR(_now):
-        $ persistent._mas_pm_gamed_late += 1
-        if mas_isMoniNormal(higher=True):
-            m 3wud "Wait, [player]!"
-            m 3hksdlb "It's the middle of the night!"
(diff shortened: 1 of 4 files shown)


jmwall24 (dismissed):

Legendkiller21 (dismissed):

multimokia (approved):

jmwall24 (dismissed):

Legendkiller21 (dismissed):

multimokia (dismissed):

jmwall24 (approved):

jmwall24 (dismissed):

Booplicate (dismissed):
Go slep

jmwall24 (dismissed):

jmwall24 (approved):

Booplicate (approved):
If you don't dismiss this, I'll be disappointed

Legendkiller21 (approved):

cfgexe (commented):
looks good

PR #8218, opened 2021-11-10
Removing "retarded" from Poem Words

KiddTheManiac:
Changed the word "retarded" to "redacted", tried to make it a Monika word rather than a Natsuki word.


Booplicate:
Thank you, but this has been handled already.

PR #8249, opened 2021-11-12
Added a spotify windows react

magmaleadermaxie:
This is the final thing I'll submit for a bit.


jmwall24 (commented):
Pretty big issue here with how Spotify works and the difference between app and browser. Browser always has `Spotify` in the name but the app only has it in the name when NOT playing a song/podcast etc, during which time it only shows the artist and title. So for people with the app, the dialogue doesn't work since she'd only react when you are NOT listening to anything, so this probably doesn't work as is.

multimokia:
[Closing as inactive]

PR #8273, opened 2021-11-18
Remove 100 Year anni from gameplay

lunulae:
Changed this anniversary dialogue to a comment, so it's still an easter egg in the code but won't prompt in the actual game.

Ideally I'd like to keep things like this commented for those who want to dig through dialogue code, but wanted to discourage actually time-traveling. In addition this current dialogue is a bit misleading and lets the player believe they haven't broken anything when they return to the normal time.

### DIFF
--- a/Monika After Story/game/script-anniversary.rpy
+++ b/Monika After Story/game/script-anniversary.rpy
@@ -701,28 +701,28 @@ label anni_50:
     m 1hubfb "And mine had one too."
     return "unlock"
 
-init 5 python:
-    addEvent(
-        Event(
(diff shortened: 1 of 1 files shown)


jmwall24 (changes requested):
This is going to need an update script to fully disable the event for anyone who has played MAS before. Also, for the event itself, let's only comment out the init block and leave the label and dialogue uncommented. That way we are still able to check the label if we ever need to in the future. This still makes it unavailable to reach in the mod itself. Also uses of `anni_100` need to be removed from other areas to pass the checks and avoid crashes.

PR #8288, opened 2021-11-22
Update zz_windowituls.rpy to allow window reacts on Linux

coder-incognito:
Window reacts now work on Linux!

```
                active_winname = unicode(active_winname_prop.value, encoding = "utf-8")
                active_winname = active_winname.replace("—","\b\0")
```

Line 302 was changed and another line was added right after it. Tested on Linux, and the feature now works! No exceptions occurring, nothing of that sort. No delays either as far as I can tell! Not sure why GitHub says there's a delay.

### DIFF
--- a/Monika After Story/game/zz_windowutils.rpy
+++ b/Monika After Story/game/zz_windowutils.rpy
@@ -299,10 +299,7 @@ init python in mas_windowutils:
             active_winname_prop = active_winobj.get_full_property(NET_WM_NAME, 0)
 
             if active_winname_prop is not None:
-                active_winname = unicode(active_winname_prop.value, encoding = "utf-8")
-                reversed_winname = active_winname[::-1]
-                terminating_index_buffer = len(active_winname) - reversed_winname.index("—")
-                active_winname = active_winname[:terminating_index_buffer - 2]
+                active_winname = unicode(active_winname_prop.value, encoding = "utf-8")    
                 return active_winname.replace("\n", "")
 
             else:
--- a/Monika After Story/game/zz_windowutils.rpy
+++ b/Monika After Story/game/zz_windowutils.rpy
@@ -300,7 +300,9 @@ init python in mas_windowutils:
 
             if active_winname_prop is not None:
                 active_winname = unicode(active_winname_prop.value, encoding = "utf-8")
-                active_winname = active_winname.replace("—","\b\0")
+                reversed_winname = active_winname[::-1]
+                terminating_index_buffer = len(active_winname) - reversed_winname.index("—")
+                active_winname = active_winname[:terminating_index_buffer - 2]
                 return active_winname.replace("\n", "")
 
             else:
--- a/Monika After Story/game/zz_windowutils.rpy
+++ b/Monika After Story/game/zz_windowutils.rpy
@@ -299,7 +299,8 @@ init python in mas_windowutils:
             active_winname_prop = active_winobj.get_full_property(NET_WM_NAME, 0)
 
             if active_winname_prop is not None:
-                active_winname = unicode(active_winname_prop.value)
+                active_winname = unicode(active_winname_prop.value, encoding = "utf-8")
+                active_winname = active_winname.replace("—","\b\0")
                 return active_winname.replace("\n", "")
 
             else:


coder-incognito:
> Please, rebase your PR to the `content` branch.

Done. Apologies for inconvenience!

Monika After Story/game/zz_windowutils.rpy
```diff
@@ -299,7 +299,8 @@ init python in mas_windowutils:
             active_winname_prop = active_winobj.get_full_property(NET_WM_NAME, 0)
 
             if active_winname_prop is not None:
-                active_winname = unicode(active_winname_prop.value)
+                active_winname = unicode(active_winname_prop.value, encoding = "utf-8")
+                active_winname = active_winname.replace("—","\b\0")
```
Booplicate, line 303: Thanks for your contribution! Can you explain what this line does?

coder-incognito: > Thanks for your contribution! Can you explain what this line does?
@Booplicate 
So essentially, rather than having python implicitly convert the name of the window into Unicode, it specifies that the encoding uses the utf-8 standard. That is line 302. This prevents it from throwing an uncaught exception when it tries to read the window title.
Line 303 rectifies the issue that Qt faces, whereby the window titles are formatted as such:
Imagine I'm on YouTube. The window title would say `YouTube — [Browser Name]"`
The code in line 303 looks for the large dash (em dash), and replaces it with a `\b\0` so that the space before it is also removed, and the string ends at just `YouTube`. This way, Monika can see that we are on YouTube, and is not confused by the extra information about out browser, or in general, our program (I say program because imagine I'm editing a document in LibreOffice Writer; it would say something like `document.ods — LibreOffice Writer`. This code would reduce the string to just `document.ods`).

Booplicate: `str.replace` replaces a character. Your example will actually become this:

```py
s = "document.ods — LibreOffice Writer"
s.replace("—","\b\0")
>>> 'document.ods \x08\x00 LibreOffice Writer'
```

dreamscached: I'm afraid ANSI escape sequences don't work outside terminal, backspace `\b` won't have any effect on space removal.

coder-incognito: @Booplicate 
Ah right, my apologies for missing that. The results seem to be working though, but let's ignore that anomaly. Hmm. Now, how does one figure out the index of the em dash then? If we can find that out, we can just replace the index before it with `\0`. I do have an idea though. And this idea could also help ensure that in the **off** chance a webpage title includes an em dash, only the last one, that is, the one put in by Qt, gets removed. Here is my idea (apologies for syntax issues, I don't know python and am working with C++ nowadays!)

```
string = "document.ods — LibreOffice Writer"
reversed_string = string[::-1]
emdash_location = len(string) - reversed_string.index("—")
fixed_string = string[:emdash_location - 2]
```

Adapted to the code, it would be
```
reversed_winname = active_winname[::-1]
emdash_location = len(active_winname) - reversed_winname.index("—")
active_winname = active_winname[:emdash_location - 2]
```

Fairly certain that replacing line 303 with this should do the trick now!

coder-incognito: Alright, seems like there is another issue. If there is no em dash to begin with, it throws an exception. Hmm. Now ow to circumvent that...

coder-incognito: Alright, Python2 can't seem to find the em dash in the string. That is... Odd. 

So what if I just do this.
```
string= "document.ods — LibreOffice Writer"
string = string.replace("—","%")
reversed_string = string[::-1]
terminating_index= len(string) - reversed_string.index("%")
fixed_string = string[:terminating_index- 2]
```

Surely no program uses `%` in it's name!
Anyway, adapting to the script, this would become
```
active_winname = active_winname.replace("—", "%")
reversed_string = active_winname[::-1]
terminating_index= len(active_winname) - reversed_active_winname.index("%")
active_winname = active_winname[:terminating_index - 2]
```

dreamscached: Why not replace it with a plain dash? It is in ASCII charset.

coder-incognito: Okay. So. It turns out I am an idiot. Your coding is already so good that it doesn't seem to care about stuff like the em dash and anything after it. My God. I apologize.
Also, can I just say that feature where Monika forgets about you as a prank is bloody scary when you're messing around with her code? Genuinely scared me. So hey, well done!

Editing the file appropriately now. Turns out, only the encoding thing matters. Thanks for bearing with my lack of experience and knowledge!

coder-incognito:
Alright, did another round of testing. Used YouTube, Twitter, the GitHub, and everything. Works fine on Linux.

ThePotatoGuy:
this can probably make next rel. updating milestones, @Booplicate please finish your review. I don't think we need a separate tester given the 1 liner change that specifies encoding

Booplicate (approved):

multimokia (approved):

Issue #8291, opened 2021-11-23
[Suggestion] - Window React Suggestion (looking for feedback) 


AutumnLeaves991:
So, after you ask Monika if she reads manga the following dialogue sequence occurs:

_"Yeah, I had a feeling you were going to ask me about this.
Natsuki would be the expert here, I guess.
I usually prefer reading to watching anime, but I'd be fine with anything if it's with you.
I don't judge other people for their hobbies. So if you want to load up some anime, go ahead!
I'll watch through your computer screen. Make sure it's something I'll like!"_

Those last two lines got me thinking it'd be nice for Monika to have a window reaction for CrunchyRoll or some other anime streaming sites. You could probably get away with re-using the dialogue she already uses for youtube, but I wouldn't mind trying my hand at coming up with some lines in her voice for it if people like the idea.

Figured it'd be worth seeing what other people thought before I waste anyone's time with a pull request lol

NonyaGG:
Sounds good enough, go ahead

lunulae:
I think it'd be a good idea. CrunchyRoll is the only (official) streaming site I'm aware of but if you know multiple ones we could use them as well.

AutumnLeaves991:
> I think it'd be a good idea. CrunchyRoll is the only (official) streaming site I'm aware of but if you know multiple ones we could use them as well.

Pull request submitted ^^, lmk what you think of the dialogue I used.  Hopefully it sounds monika-esque

Kzitold:
Manga is literature, so how about a reaction to MangaDex, MangaPark, etc...

ThePotatoGuy:
[closing as implemented]

PR #8305, opened 2021-11-26
Adds Crunchyroll window react 

AutumnLeaves991:
As per #8291 I've added a window react for Crunchyroll.
Monika will have different reactions depending on the value  of persistent._mas_pm_watch_mangime, similar to the MyAnimeList window react.

Code functions as expected on Windows, dont have a Linux box on hand to check that myself at the moment.


Monika After Story/game/script-windowreacts.rpy
```diff
@@ -620,3 +620,37 @@ label mas_wrs_word_processor:
...
+            show_in_idle=True
+        ),
+        code="WRS"
+    )
+
+label mas_wrs_crunchyroll:
+    python:
+        if persistent._mas_pm_watch_mangime is None:
+            crunchyroll_quips = ["Oh! So you like anime, [player]?"]
+        else:
```
multimokia, line 645: Seeing the condition here, the `else` block handles _both_ whether you said you do like manga/anime _or_ you don't. It might be worth adding another set for not liking

AutumnLeaves991: > might be worth adding another set for not liking

Hmm, yeah I think you're right.  Not sure what she would say if you loaded crunchyroll after telling her you didn't like anime though.  🤔🤔

ibm2431: "I'm glad you're trying to expand your horizons!"

lunulae: Something like "Did you find a series you like?" might be cute.

multimokia (approved):

lunulae (approved):

PR #8345, opened 2021-12-03
Add going to work on something to generic brb menu

multimokia:
Small addition to the generic brb menu, adds `I'm going to work on something.`

Requires basically no review other than whether or not the option shows up.


jmwall24:
Already implemented, silly laf

PR #8372, opened 2021-12-09
Fix settings menu

Booplicate:
Fixed the issue with returning `None` and coercing it to `str`ings. Also safer access to the other map + using the defined constants for it instead of raw literals.

### Testing:
 - verify the chatter still works
 - set `persistent._mas_randchat_freq` to `-99` or `99` and verify you can still open the menu

### DIFF
--- a/Monika After Story/game/screens.rpy
+++ b/Monika After Story/game/screens.rpy
@@ -1421,6 +1421,7 @@ screen preferences():
                 style_prefix "slider"
                 box_wrap True
 
+                # TODO: clean this up, we should not exucute py code in screens
                 python:
                     ### random chatter preprocessing
                     if mas_randchat_prev != persistent._mas_randchat_freq:
--- a/Monika After Story/game/definitions.rpy
+++ b/Monika After Story/game/definitions.rpy
@@ -8087,7 +8087,7 @@ init -1 python in mas_randchat:
         """
         Reduces the randchat setting if we're too high for the current affection level
         """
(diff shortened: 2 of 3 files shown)


Monika After Story/game/screens.rpy
```diff
@@ -1421,6 +1421,7 @@ screen preferences():
                 style_prefix "slider"
                 box_wrap True
 
+                # TODO: clean this up, we should not exucute py code in screens
```
ThePotatoGuy, line 1424: what's your idea for this TODO? The pycode here needs to run with the screen because it handles updating screen text and keeping the suntime sliders in a valid order.

ThePotatoGuy (changes requested):

Monika After Story/game/screens.rpy
Booplicate: I'll add validators which would be run on the sliders changing. Alternatively validate the values are in the valid range on the menu exit.

For generating labels text we can wrap that in functions. Like
```renpy
# Say for sunset
label _("[[ " + mas_suntime.get_ss_disp() + " ]")
```

Screens have an odd behavior and executing py code like that isn't ideal.

ThePotatoGuy: ok, but validating on exit is a no-go. The sliders should validate immediately otherwise users can drag them to be invalid and its an ugly experience.

ThePotatoGuy (approved):

Legendkiller21 (approved):

PR #8393, opened 2021-12-12
Add test topic and exception for `register_img_same`

ThePotatoGuy:
Copied from @multimokia: 
> Figured the issue out. It's an issue with init order. We should **never** use `register_img_same` on the same init level as `register_img` for the original, because it will always fail provided the tag does not exist yet.

# Key Changes
* Adds a new exception in `MASImageTagDecoDefinition.register_img_same` when it could not find a deco definition from the source BG (we'll just assume that this is only being called at init time).
* Adds a testing topic `DECO TAG TEST API REG SAME` for checking the `register_img_same` function.

# Testing
* in `dev/dev_deco.rpy`, uncomment the section under  `register_img_same init exception` - verify that the game raises the new exception on load
* run the `DECO TAG TEST API REG SAME` dev topic - verify that the blue deco remains in the same place for each background.


multimokia (changes requested):
Figured the issue out. It's an issue with init order. We should **never** use `register_img_same` on the same init level as `register_img` for the original, because it will always fail provided the tag does not exist yet. 

As such:
It might be worth just raising an error or logging in place of silently handling the issue as to actually provide some feedback to the user. Requesting some changes to add logging or an ex throw for it if config.developer

ThePotatoGuy:
@multimokia I added an exception

multimokia (approved):

Legendkiller21 (approved):

PR #8414, opened 2021-12-15
show dialogue feedback when gifts are given during d25

ThePotatoGuy:
# TODO
- [ ] waiting on dialogue from @lunulae 
- [ ] make the actual flow change ? or whatever we plan to do.

### DIFF
--- a/Monika After Story/game/script-holidays.rpy
+++ b/Monika After Story/game/script-holidays.rpy
@@ -2045,6 +2045,12 @@ init -10 python in mas_d25_utils:
         d25_gen = []
         mas_frs.process_gifts(d25_giftnames, d25_evb, d25_gsp, d25_gen)
 
+        # TODO: so the reason why d25 gifts are not reacted to is because 
+        # the nd25 gifts (non-d25) are passed into the build_gift_react_labels`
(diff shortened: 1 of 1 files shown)


Monika After Story/game/script-holidays.rpy
```diff
@@ -2350,6 +2364,63 @@ label mas_d25_season_exit:
         mas_d25ReactToGifts()
     return
 
+# pre-d25 gift handling
+# NOTE: YES this is super hackish, but until we have event/topic context 
+#   this is the quickest way without adding an extra per var.
+# multiple d25, no non-d25
+label mas_d25_gift_tree_intro_2_0:
+    $ d25_multiple = True
+    $ nd25_any = False
```
Booplicate, line 2373: If this helps, those could be label arguments. You could also use `**kwargs`
```renpy
label mas_d25_gift_tree_intro(d25_multiple, nd25_any, **kwargs):
    # ...
    return
```
The only downside you'd have to use `call` instead of `jump`, but it doesn't look like it'd be a problem in this case.

ThePotatoGuy:
on hold until #8453

Issue #8427, opened 2021-12-17
0.11.1 update script fix


ThePotatoGuy:
prevent div 0 and re apply code after.

ThePotatoGuy:
[closing as fixed]

PR #8429, opened 2021-12-17
Fix 0.11.1 update script and re-apply updates that may not have ran.

ThePotatoGuy:
#8427 

# Key Changes
* fixes the div by 0 error in v0.11.1 update script
* adds the 0.11.1 update script lines that may not have ran to 0.12.6 update script, with conditions to prevent re-running code that may not be safe to re-run.

# Testing
* verify 0.11.1 update script doesn't crash if the total number of sessions is 0. (this may be difficult to recreate - probably fine to just make sure the code is solid)
* test the new update script, making sure that we don't incorrectly overwrite data. Specifically:
    * `monika_credits_song`'s conditional should not be reset if you have seen the topic at least once
    * `_mas_grandfathered_nickname` is not overwritten if it already is set
    * `_mas_pm_called_moni_a_bad_name` is not overwritten if `_mas_called_moni_a_bad_name` is not set
* other pieces in the update script:
    * `orcaramelo_twin_tails` should be unlocked and selectable 
    * `_mas_penname` should be set to None if it was set to `""`


ThePotatoGuy:
@Booplicate @Legendkiller21 this should be ready for testing again

multimokia (approved):

jmwall24 (approved):

Issue #8206, opened 2021-11-09
[Suggestion] Time-based affection gain when holding Monika


Batcastle:
First off, lemme just say I am loving this mod!

I feel like there could be a great improvement by having the amount of affection you gain change depending on how long you hold Monika (with a cap of course). I feel this could be a fairly easy way to improve the immersion a tiny bit. It could be as simple as something like this:

(Python-based psudeocode. Sorry if it's not that helpful)
```python3
start_hug_time = time.time()
# Do the hug mechanic
# hug ends
end_hug_time = time.time()
hug_duration = end_hug_time - start_hug_time
if hug_duration < 10:
    affection += 0.75
elif 10 <= hug_duration < 30:
    affection += 1.25
elif 30 <= hug_duration < 90:
    affection += 2
else:
    # At this point we know hug_duration >= 90
    if monika.is_asleep():
        if monika.sleep_duration() < (60 * 30): # number of seconds in 30 minutes.
            affection += 5
        else:
            # this high of an affection gain might warrant this event being an occasional bypass.
            # Make bypass-capable have a 3-5 day cool down maybe?
            affection += 10 
    else:
        affection += 3
```
The idea is that, the longer you hug her, the more affection goes up. If she falls asleep, affection goes up even more. This could be made even more granular by making a function that handles this and runs the sleep duration through some sort of math equation, but I personally feel that that level of granularity is unnecessary.

Thanks for hearing me out! Keep up the good work!

Booplicate:
That's a no from me. *Maybe* a bonus if she slept for X hours, but there's already enough ways to raise affection. It's a very fragile system, we already have a problem with unnatural gain.

> a fairly easy way to improve the immersion

Affection is internal, the player doesn't get to know when and how much they get. Adding more ways to gain it doesn't improve immersion. Affection farming is actually gamification and leads to immersion breaking.

Batcastle:
While I understand those concerns, the intent wasn't to game-ify MAS. It was more to promote the possibility of making Monika more life-like. I feel like why I want this for Monika delves more into psychology then we should get into here (more than happy to share my thought process though if someone wants to hear it), so if the best way to achieve anything like this is to have a bonus after she sleeps for so long, then I can settle for that.

ThePotatoGuy:
I think this is reasonable since its confined to the holding topic. Feel free to PR if you want to implement it, otherwise it might be a while before this is added since affection is already real easy to get.

[closing until we look into it]

[...]

PR #8506, opened 2021-12-26
implement greater affection increase for longer hold times

Batcastle:
Just a simple if/elif check can handle this.

Implements #8206 

I am in the middle of testing this, but preliminary results are promising.

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -2355,19 +2355,18 @@ label monika_rain_holdme:
         # small affection increase so people don't farm affection with this one.
         python:
             # Monika likes being held for long periods of time, the longer the better
-            mod = 0.25
             if elapsed_time <= datetime.timedelta(minutes=10):
                 mod = 0.25
             elif ((elapsed_time > datetime.timedelta(minutes=10)) and (elapsed_time <= datetime.timedelta(minutes=30))):
-                mod = 0.5
+                mod = 0.45
             elif ((elapsed_time > datetime.timedelta(minutes=30)) and (elapsed_time <= datetime.timedelta(minutes=60))):
-                mod = 0.75
+                mod = 0.65
             elif ((elapsed_time > datetime.timedelta(minutes=60)) and (elapsed_time <= datetime.timedelta(minutes=100))):
-                mod = 1
+                mod = 0.75
             elif ((elapsed_time > datetime.timedelta(minutes=100)) and (elapsed_time <= datetime.timedelta(minutes=150))):
-                mod = 1.25
-            elif elapsed_time > datetime.timedelta(minutes=150):
-                mod = 1.5
+                mod = 0.85
+            else:
+                mod = 1.0
             mas_gainAffection(modifier=mod)
 
     else:
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -2362,13 +2362,13 @@ label monika_rain_holdme:
             delta_time = datetime.datetime.now() - start_hold_time
             if delta_time <= datetime.timedelta(minutes=10):
                 mod = 0.25
-            elif ((delta_time > datetime.timedelta(minutes=10)) and (delta_time <= datetime.timedelta(minutes=30))):
+            elif delta_time <= datetime.timedelta(minutes=30):
                 mod = 0.45
-            elif ((delta_time > datetime.timedelta(minutes=30)) and (delta_time <= datetime.timedelta(minutes=60))):
+            elif delta_time <= datetime.timedelta(minutes=60):
                 mod = 0.65
-            elif ((delta_time > datetime.timedelta(minutes=60)) and (delta_time <= datetime.timedelta(minutes=100))):
+            elif delta_time <= datetime.timedelta(minutes=100):
                 mod = 0.75
-            elif ((delta_time > datetime.timedelta(minutes=100)) and (delta_time <= datetime.timedelta(minutes=150))):
+            elif delta_time <= datetime.timedelta(minutes=150):
                 mod = 0.85
             else:
                 mod = 1.0
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -2347,6 +2347,7 @@ label monika_rain_holdme:
     if mas_is_raining or mas_isMoniAff(higher=True):
         call monika_holdme_prep
         m 1eua "Of course, [mas_get_player_nickname()]."
+        $ start_hold_time = datetime.datetime.now()
         call monika_holdme_start
 
         call monika_holdme_reactions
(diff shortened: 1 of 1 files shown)


Monika After Story/game/script-topics.rpy
```diff
@@ -2353,7 +2353,22 @@ label monika_rain_holdme:
 
         call monika_holdme_end
         # small affection increase so people don't farm affection with this one.
-        $ mas_gainAffection(modifier=0.25)
+        python:
+            # Monika likes being held for long periods of time, the longer the better
+            mod = 0.25
+            if elapsed_time <= datetime.timedelta(minutes=10):
+                mod = 0.25
+            elif ((elapsed_time > datetime.timedelta(minutes=10)) and (elapsed_time <= datetime.timedelta(minutes=30))):
```
jmwall24, line 2361: the first half of the check here and every subsequent elif check of the previous entry is not needed, if we are here we have already failed all previous checks so we know it's been over 10 minutes, therefore each time you only need the time you are actually checking. also the last elif check can simply be an `else:`

Monika After Story/game/script-topics.rpy
```diff
@@ -2353,7 +2353,22 @@ label monika_rain_holdme:
 
         call monika_holdme_end
         # small affection increase so people don't farm affection with this one.
-        $ mas_gainAffection(modifier=0.25)
+        python:
+            # Monika likes being held for long periods of time, the longer the better
+            mod = 0.25
+            if elapsed_time <= datetime.timedelta(minutes=10):
```
Booplicate, line 2359: `elapsed_time` is undefined.

Monika After Story/game/script-topics.rpy
```diff
@@ -2353,7 +2353,22 @@ label monika_rain_holdme:
 
         call monika_holdme_end
         # small affection increase so people don't farm affection with this one.
-        $ mas_gainAffection(modifier=0.25)
+        python:
+            # Monika likes being held for long periods of time, the longer the better
+            mod = 0.25
```
Booplicate, line 2358: Use `else` path to define the "fallback" value.

Monika After Story/game/script-topics.rpy
```diff
@@ -2353,7 +2353,22 @@ label monika_rain_holdme:
...
+            elif ((elapsed_time > datetime.timedelta(minutes=10)) and (elapsed_time <= datetime.timedelta(minutes=30))):
+                mod = 0.5
+            elif ((elapsed_time > datetime.timedelta(minutes=30)) and (elapsed_time <= datetime.timedelta(minutes=60))):
+                mod = 0.75
+            elif ((elapsed_time > datetime.timedelta(minutes=60)) and (elapsed_time <= datetime.timedelta(minutes=100))):
+                mod = 1
+            elif ((elapsed_time > datetime.timedelta(minutes=100)) and (elapsed_time <= datetime.timedelta(minutes=150))):
+                mod = 1.25
+            elif elapsed_time > datetime.timedelta(minutes=150):
+                mod = 1.5
```
Booplicate, line 2370: I think `1.5` is too much, set maximum to `1.0`.

Booplicate (changes requested):
Additional bits

Monika After Story/game/script-topics.rpy
Batcastle: `elapsed_time` doesn't resolve to undefined. In `monika_holdme_reactions`, it gets defined. And apparently due to the way Ren'Py handles Python one-liners and python blocks, it never gets undefined.

Booplicate: `elapsed_time` is a common name, thus it can be deleted, cleaned, redefined. It also means if we change it in one place where it's defined, we'd have to change in all other places. We should never use variables from other contexts. Sadly, we can't fix it in this PR, so just add a TODO to change that later.

Batcastle:
Tested this multiple times on my personal MAS installation and got no errors. I didn't take it all the way to >150 minutes, but as long as the syntax is correct the logic is the same so it should work fine.

Monika After Story/game/script-topics.rpy
```diff
@@ -2353,7 +2353,21 @@ label monika_rain_holdme:
 
         call monika_holdme_end
         # small affection increase so people don't farm affection with this one.
-        $ mas_gainAffection(modifier=0.25)
+        python:
+            # Monika likes being held for long periods of time, the longer the better
+            if elapsed_time <= datetime.timedelta(minutes=10):
+                mod = 0.25
+            elif ((elapsed_time > datetime.timedelta(minutes=10)) and (elapsed_time <= datetime.timedelta(minutes=30))):
```
jmwall24, line 2360:
```suggestion
            elif elapsed_time <= datetime.timedelta(minutes=30):
```

this is what I meant, you don't need the 10 minute check since the last check covered this. you can do this for all subsequent checks.

Monika After Story/game/script-topics.rpy
Batcastle: Okay. I'll add that TODO and compute it myself then. :+1:

Booplicate: We already thought about it, right now there's no better solution available. So you can keep this global. I need a TODO to not forget about it and fix later (I'm changing this code too).

Batcastle: I already did the change on my local repo. Oh well. Makes it so that when the change comes around that it's one less thing that might need fixing. :shrug:

Monika After Story/game/script-topics.rpy
Batcastle: Ohhhhhh. Thank you for the clarification. I'll do that.

Monika After Story/game/script-topics.rpy
```diff
@@ -2347,13 +2347,32 @@ label monika_rain_holdme:
...
 
         call monika_holdme_end
         # small affection increase so people don't farm affection with this one.
-        $ mas_gainAffection(modifier=0.25)
+        python:
+            # Monika likes being held for long periods of time, the longer the better
+            # We COULD use elapsed_time for this, but as it's a common var, we shouldn't
+            # in case it is being used by something else, so:
+            # TODO: zero-out, `del`, or `None`-ify elapsed_time when done with it in `monika_holdme_reactions`
+            delta_time = datetime.datetime.now() - start_hold_time
```
Booplicate, line 2362: That's the idea I had too. But it won't work because in `monika_holdme_reactions` you can spend unlimited time reading the dialogue. So this value isn't reliable either.

Monika After Story/game/script-topics.rpy
Batcastle: Well according to @jmwall24 Doing it the way I was has it's own issues, so looks like I _have to use_ `elapsed_time` in order to get an accurate number. :-/

Batcastle:
Okay. @jmwall24 @Booplicate I believe I addressed all the requested changes as best I could. If any other requested changes pop up, I'll address them as I see them and get time. Is there anything else I need to do on my end other than that and test again to make sure things are still working?

Thank you for your help and for being so quick to respond!

Issue #8538, opened 2022-01-01
[Bug] - KeyError: u'curr_year+1'


aarfran:
I encountered a script error on New Years eve possible relating to a dialogue greeting. I'm a complete novice when it comes to coding but thought I would report it here. It asked me to quit, look at the traceback, or ignore. I copied the traceback log and hit ignore. Here is the full text:

I'm sorry, but an uncaught exception occurred.

While running game code:
  File "game/script-ch30.rpy", line 1510, in script call
  File "game/event-handler.rpy", line 2530, in script call
  File "game/script-holidays.rpy", line 4720, in script
  File "game/definitions.rpy", line 152, in get_field
KeyError: u'curr_year+1'

-- Full Traceback ------------------------------------------------------------

Full traceback:
  File "game/script-ch30.rpy", line 1510, in script call
  File "game/event-handler.rpy", line 2530, in script call
  File "game/script-holidays.rpy", line 4720, in script
  File "D:\Documents\Monika After Story\DDLC-1.1.1-pc\renpy\ast.py", line 613, in execute
    renpy.exports.say(who, what, interact=self.interact)
  File "D:\Documents\Monika After Story\DDLC-1.1.1-pc\renpy\exports.py", line 1147, in say
    who(what, interact=interact)
  File "D:\Documents\Monika After Story\DDLC-1.1.1-pc\renpy\character.py", line 868, in __call__
    what = what_pattern.replace("[what]", sub(what, translate=translate))
  File "D:\Documents\Monika After Story\DDLC-1.1.1-pc\renpy\character.py", line 850, in sub
    return renpy.substitutions.substitute(s, scope=scope, force=force, translate=translate)[0]
  File "D:\Documents\Monika After Story\DDLC-1.1.1-pc\renpy\substitutions.py", line 232, in substitute
    s = formatter.vformat(s, (), kwargs)
  File "/home/tom/ab/x64lucid-deps/install/lib/python2.7/string.py", line 563, in vformat
  File "/home/tom/ab/x64lucid-deps/install/lib/python2.7/string.py", line 585, in _vformat
  File "game/definitions.rpy", line 152, in get_field
  File "/home/tom/ab/x64lucid-deps/install/lib/python2.7/string.py", line 605, in get_value
KeyError: u'curr_year+1'

Windows-8-6.2.9200
Ren'Py 6.99.12.4.2187
Monika After Story 0.12.6

jmwall24:
[...]

PR #8542, opened 2022-01-01
d25/nye fixes

jmwall24:
Various fixes of d25/nye mishaps

## Fixes
- issue in `mas_d25_spent_time_monika` where the post-poem dialogue was specific to the first d25 poem and didn't really make sense with the other poems
- `persistent._mas_pm_accomplished_resolution` and `persistent._mas_pm_has_new_years_res` were being stored as `pm` vars instead of `nye` vars and thus not being reset on a yearly basis, leading to dialogue that didn't make sense. They are being converted to `persistent._mas_nye_accomplished_resolution` and `persistent._mas_nye_has_new_years_res` respecitvely.
- people who got the nye_dress previously and are now at `love` get `mas_nye_monika_nye_dress_intro` pushed but don't actually see anything related to the event since they already have the dress unlocked and we don't have Monika wear it without being asked in this case. This is now being changed so she puts it on in all cases if this event is pushed.
- `mas_nye_monika_nye_dress_intro` also had a bug for people who were getting the dress put on for a second+ time which has now been fixed (#8538)

## TODO
- [x] update script to delete the old `pm` vars and migrate data to new `nye` vars


ThePotatoGuy (dismissed):

ThePotatoGuy (approved):

Booplicate:
We might need to transfer old history from `pm.actions.made_new_years_resolutions` to `nye.actions.made_new_years_resolutions`. Right you have to check both to get the correct result.

ThePotatoGuy:
added transferring historial data in the update script

ThePotatoGuy (approved):

Monika After Story/game/updates.rpy
```diff
@@ -381,9 +381,18 @@ label v0_12_8_1(version="v0_12_8_1"):
...
+                "nye.actions.did_new_years_resolutions"
+            )
             safeDel("_mas_pm_accomplished_resolutions")
+
         if persistent._mas_nye_has_new_years_res is None:
             persistent._mas_nye_has_new_years_res = persistent._mas_pm_has_new_years_res
+            store.mas_history._store_all(
+                mas_HistLookup_all("pm.actions.made_new_years_resolutions"),
+                "nye.actions.made_new_years_resolutions"
+            )
```
Booplicate, line 395: `mas_HistLookup_all` returns data of format `{year: (code, value)}`,
but `_store_all` uses `_store` which accepts `value`, `key`, `year`. This code will save a tuple `(code, value)` as `value`.

Booplicate: Something like `{k: v[1] for k, v in mas_HistLookup_all(KEY).items()}` would work I think, unless you know a better way.

ThePotatoGuy: `_store_all` says its supposed to work with the data tuple from `mas_HistLookup` so I changed `_store_all` instead.

Booplicate (approved):

multimokia (approved):

Legendkiller21 (approved):

PR #8585, opened 2022-01-10
just a dialogue where you tell monika that you will die

loxk090:
it would be a dialogue where you tell monika that you will die soon or choose it on the calender and a whole dialogue of what you did with her since you start dating her will start. after the date you said that you would die she leaves like when her affection is below -110 but you can write a note in the character to bring her back like the imsorry.txt


multimokia:
Please open an issue for suggestions. Pull requests are for completed topic submissions.

That said it's not likely this topic idea will be implemented due to the severity of the issue. Picking a death date will definitely not go over well with Monika and she'd never let you actually do that

PR #8605, opened 2022-01-13
Update script-moods.rpy "suicidal" mood

bee-and-honey:


WishStone-Dragon:
Giving feedback purely on the text and language:
Unsure how wise it would be to speak to this (though I have previously been scolded already for voicing such a thought, so I know my view is perhaps not reflected by others at all).

Three things I would suggest to perhaps buffer or catch things:

1. We have a script in place already where Monika allows the player to 'vent' to her. She will say that she is happy to give the player space to express themselves, even if she cannot hear the player. A similar tech might be useful here.
2. This is very, very short for a very, very serious topic. It could feel like being fopped off and dismissed. And when in the dark of depression, 'I love you" alone cannot easily reach. The other "I feel..." topics have Monika support and help the player. This so far does not.
3. We have the means to link players for all kinds of things, such as how to secure her persistent files or links to homepages. Perhaps a line on how worried she is that she cannot help, but there are resources such as - suicide prevention page. I know we have an international player base, so something like https://www.iasp.info/crisis-centres-helplines/ might be helpful.

multimokia:
Hi, I don't think this is a good idea to implement as the nature of this issue is deeply complex and when a person is at this state, it's very, *very* easy to do the wrong thing.

As such, it's safer to not have this mentioned.
For the same reasons, we also tend to not touch too deeply on issues and potentially how to solve them aside from maybe reaching out to a professional, since we're not professionals, and things which help one person may be detrimental to another.

There's just no one-size fits all approach to this sort of thing and I don't think we should entertain it.

While this was put up in good intentions, I think it's just too risky.
I'm going to close this as rejected as a result.

PR #8616, opened 2022-01-15
Music menu but it auto repopulates

multimokia:
Updated music menu so we don't have to constantly re initialize music choices whenever custom music is added as it's extremely unintuitive to use the topic and it just makes more sense if the menu just refreshed on its own when opened.

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -13373,19 +13373,19 @@ label monika_add_custom_music_instruct:
     $ persistent._seen_ever["monika_load_custom_music"] = True
     return
 
-init 5 python:
-    addEvent(
-        Event(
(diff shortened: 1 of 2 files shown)


ThePotatoGuy (changes requested):
I added 23 songs and the lag when opening the menu can get pretty bad. [here's a video](https://www.youtube.com/watch?v=83ke-qbYQow) (ignore the audio its fauna streaming). It's easier to see when the click is done in the 2nd/3rd try in the video where the mouse is moved away from the music button after the click (the music button is highlighted during the click processing).

I don't think scanning for the songs on open is going to work. If we're adamant about removing the refresh topic then I propose a rescan button in the music menu instead.

multimokia:
Closing as gonna fix this eventually:tm:

Issue #8453, opened 2021-12-20
event/topic context + more


ThePotatoGuy:
running into this feature deficiency more often now, but basically something where we can add some context for a label/event/topic that can be accessed later in said label/event/topic.

-- see #8619

ThePotatoGuy:
*added a plan to add dt to the context - this means we can expire topics.

ThePotatoGuy:
[closing as implemented]

PR #8619, opened 2022-01-16
Add Event List wrappers and Event Context

ThePotatoGuy:
#8453 

# Key Changes
* adds ability for events to have contextual data attached to them that can be referenced later when the event is called
    * in other words, Event labels now can act like async callbacks/partial functions
* adds an abstraction of the event list to consolidate various event-list-based functions under the same umbrella
* adds utility classes for objects that can have flexible attributes
* `mas_ev_data_ver._strict_can_pickle` - strict data check that only allows known safe-to-pickle data types.

## Event Context

Event context is a special object that can be set while pushing/queuing an event that is loaded when that event runs. The event that is running can access this object and retrieve data set into the object at the time it was pushed/queued. The context data is stored in persistent and can be restored on a restart.

Quick example of use:

```python
ctx = MASEventContext()
ctx.data_for_later = 10
MASEventList.push("other_topic", context=ctx)

# ~~~ in other topic ~~~
ctx = MASEventContext.get()
if ctx.data_for_later == 10:
    # this will run
else:
    # this will not run
```

`MASEventContext` also supports creation using a dictionary, such as:
```python
value = 10
ctx = MASEventContext({
    "name": value
})
ctx.name  # contains 10
```

**NOTE: This does not allow Event labels to have arguments.** I decided **not** to support event label arguments because that would make the event code more susceptible to crashes if labels are pushed/queued without the correct context args. I'm sure there could be a way to guardrail against this by introspecting the labels, but that's a lot of work for a little gain. A context object that won't crash even if you try to get an attribute that doesnt exist from it is simpler and gets the job done.

## Event List changes

To better support changes to event list structure in the future, all event-list-based functionality is now abstracted behind `MASEventList`. All code that needs to manipulate the event list or push or queue events should use `MASEventList`. 

When retrieving items from the event list, `MASEventList` returns `EventListItem` objects. These objects have functions for getting specific pieces of data from an event list entry, like the event label, the notify flag, or the event context (see below).

## New Classes
* `mas_utils.IsolatedFlexProp` - basically like a persistent in that attributes can return a default value if not set, and any attribute can be set on it. Should be lighter than a persistent since it won't have change tracking. this stores manually set attributes in a separate internal structure, which allows the manually set data to be retrieved, cleared, or set. (Separate structure meaning not using `__dict__`
* `evhand.EventListItem` - object wrapper around individual items in `persistent.event_list`. returned by functions in `MASEventList`
    * `.evl`/`.event_label` - gets the event label that will run
    * `.notify` - gets whether or not the event will cause a notification to send
    * `.ctx`/`.context` - gets the event context for the event
* `MASEventContext` - object containing an Event's context. Extends `IsolatedFlexProp`, so all non-underscore-prefixed attributes can be accessed without causing crashes
    * `MASEventContext.get` - static function to retrieve the current context
    * `MASEventContext.is_allowed_data` - static function that can check if some data is allowed to be stored in the context
* `MASEventList` - static abstraction class that acts as a wrapper around `persistent.event_list`
    * `MASEventList.iter` - gets an iterable over the event list that yields `EventListItem` objects
    * `MASEventList.peek` - gets the `EventListItem` for the next event on the event list that would start next if `call_next_event` runs
    * `MASEventList.pop` - replacement for `popEvent`. Not identical to `popEvent`. This returns an `EventListItem` now instead of a tuple of the data.
    * `MASEventList.push` - replacement for `pushEvent`
    * `MASEventList.queue` - replacement for `queueEvent`
    * `MASEventList.rev_enum_iter` - reversed enumerated iterable for the event list. yields index and `EventListItem` objects
    * `MASEventList.rev_idx_iter` - reversed index iterable for the event list. yields indexes.

## deprecations
* `pushEvent` - usable but does not support context
* `queueEvent` - usable but does not support context
* `popEvent` - dummied out and hardstop as context is important for events now.

# Testing

There are 4 dev testing topics:
* Dev > TEST EVENT CONTEXT - tests context passing from one event to another
* Dev > TEST EVENT CONTEXT - tests setting context data with non-picklables or otherwise bad data. Should not crash.
* Dev > TEST EVENT CONTEXT (restart) - tests context retention if the game crashes and restarts the event
* Goodbye > DEV BYE , CTX TEST - tests context retention if an event with context is on the event list when the game quits and is restarted

1 new unit test:
* Dev > Unit Tests > `strict_can_pickle`

Regression tests:
* verify normal topics don't crash


Monika After Story/game/script-ch30.rpy
```diff
@@ -93,6 +93,10 @@ init -1 python in mas_globals:
     this_ev = None
     # the current topic, but as event object. may be None.
 
+    _this_ev_ctx = None
```
multimokia, line 96: Given this shouldn't be accessed directly, it may be cleaner to keep this defined statically and privately on MASEventContext to avoid it from being accessed, and also keep it self contained

ThePotatoGuy: moved it to the class

multimokia (approved):

ThePotatoGuy:
this requires an unstable so pushing back to next rel

Booplicate (changes requested):
Some review bits, let me know what you think.

Monika After Story/game/event-handler.rpy
```diff
@@ -1459,6 +1461,112 @@ init -1 python in evhand:
...
+            """
+            Builds an ELI.
+
+            IN:
+                evl - event label
+                *args - the other args for an EventListItem.
+
+            RETURNS: EventListItem object
+            """
+            return EventListItem(EventListItem._build_raw(evl, *args))
```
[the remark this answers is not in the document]
ThePotatoGuy: my threshold for using classmethod is when I get really fed up typing the class name. Twice (or 3x) is not enough to reach that threshold, so I'm fine leaving it static.

The constructor is for direct object wrapper from `event_list`. The static build methods are for when passing in each data element (or some of the data elements).

`_build_raw` just generates the raw data tuple, originally separate because I was gonna use it for just push/queue, but then changed those functions to use `build` instead so we can leverage a single code path for raw data format (`_raw`). 

I'd rather keep them separate just in case.

Monika After Story/game/event-handler.rpy
```diff
@@ -1459,6 +1461,112 @@ init -1 python in evhand:
...
+
+        @staticmethod
+        def _build_raw(evl, *args):
```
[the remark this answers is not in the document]
ThePotatoGuy: so when I started writing this class, I thought about what ways I would need to create this object for:
1. use an item directly from `event_list`
2. create one from named params
3. create one using older data - for upgrading users

option 2 is the straightforward one - use named params for each thing, so I started with that for the constructor.
option 1 would be a static method that takes the item and splits it into args for the constructor - modify this a little bit and now we can handle option 3 with some safety checks (this would also need constants for default values)

this was fine until I thought about what happens when a new data element is added:
1. add the param to push/queue
2. add the param to the constructor
3. add handling for the param in the option 1/3 static method
4. add a default value entry
5. add a private var for the param
6. add the get function for the param

Steps 2, 3, and 5 are kind of annoying - its basically repeated param/documentation from step 1. So how can this be removed? 

Generic kwargs gets rid of step 2 and 5, but still need to do "arg name", which is same concept as the item indices. Step 3 still needs manual changes because the ordered data (`event_list`) would still need to be split and properly passed into kwargs.

Generic `*args` means that we only need to consider data position, which is guaranteed since we would never change the order in `event_list` - and this is a hard never, like `Event` itself. This also makes the static method covering option 1/3 even easier to write - just use list splicing to make up the difference when using old data (list splicing even handles if the the splice index is larger than the actual list, i.e: `abc = [1,2,3] -> abc[10:] -> []`). Now we end up with these steps when a new param is added:
1. add the param to push/queue
2. add a default value entry
3. add the index constant (honestly not even needed but I added for the sake of magic numbers)
4. add the get function for the param

Way simpler to add more. Is it dangerous for callers? You bet! but it's only usage outside of `MASEventList` should be read-only data retrieval. If I called this `MASEventListItem` then I would have made it safer.

Regarding your points in the other comment:
1. Mentioned above that order will not change in `event_list`, so the order flexibility is not needed. But okay, let's imagine that it does change - well the `_raw` function means that we can change the order when the EventListItem is actually added to the `event_list`. We can also change the indices and default values order to handle input. I think even a named args solution would require us to make a change in the output to support a new order, so we save maybe 1 set of changes for something that rarely will happen?
2. Agreed, but with list splicing we can make args safe, too.
3. We'd still have to keep the names somewhere, so this seems more like a 1:1 replacement. (we all seem to have varying opinions on magic strings, too)
4. mentioned above that kwargs 1 step extra for each new data element compared to args. also mentioned above that the "arg name" is same conceptually as the unique id
5. Agreed, but deprecations will be almost never. If a deprecation ever occurs, it would become an unused data element, similar to what happens with `Event`. Wasteful, yes, but negligible data usage.

Monika After Story/game/event-handler.rpy
```diff
@@ -2416,6 +2824,22 @@ init python:
         return u_count != count
 
 
+    def mas_areEventsPaused():
```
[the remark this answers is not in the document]
ThePotatoGuy: moved this API over into event list. I don't think we can transfer the var over because its been released. Better to avoid breaking submods. 

Also generally speaking, global vars are for things that may change often and should be accessible for general usage. this unpause dt fits those categories.

Monika After Story/game/event-handler.rpy
```diff
@@ -2526,27 +2953,31 @@ label call_next_event:
...
+        $ MASEventList.clear_current()
+        $ MASEventContext._set(None)
         $ mas_globals.this_ev = None
+
         # Handle idle exp
         $ mas_moni_idle_disp.do_after_topic_logic()
 
-        #if this is a random topic, make sure it's unlocked for prompts
-        $ ev = evhand.event_database.get(event_label, None)
         if ev is not None:
```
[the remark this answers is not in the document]
ThePotatoGuy: `this_ev` could be overwritten too, so I'll just re-fetch the ev.

Monika After Story/game/event-handler.rpy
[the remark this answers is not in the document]
Booplicate: Even if this API isn't public, I still don't feel safe knowing we rely on precise arg position. I'd rather be extra safe now than sorry later when we do something dum and save invalid data in persistent. List slicing doesn't really make it much better to me.

It's harder to read what's going on in `data = list((evl, ) + args + EventListItem.DEFAULT_VALUES[len(args):])`. I can understand it, it's just not as clear to read.

I think keeping default values in a `dict` is better. Yes, it's still a value + key, but it feel more cohesive because everything is together. If we could use enums, I'd see idx work, but right now `IDX_EVENT_LABEL` is just an `int` and doesn't tell much when working with the code.

I won't push this further tho and leave the final decision to you.

Monika After Story/game/0utils.rpy
```diff
@@ -624,6 +624,82 @@ python early in mas_utils:
...
+            """
+            self._default_val = default_val
+            self._set_vars = {}
+
+        def __contains__(self, item):
+            return item in self._set_vars
+
+        def __getattr__(self, name):
+            if name.startswith("_"):
+                return getattr(self, name)
```
Booplicate, line 663: Recursion call
```suggestion
                return super(IsolatedFlexProp, self).__getattribute__(name)
```

Monika After Story/game/event-handler.rpy
```diff
@@ -1713,6 +1823,363 @@ init python:
     import store.evhand as evhand
     import datetime
 
+
+    class MASEventContext(mas_utils.IsolatedFlexProp):
```
Booplicate, line 1827: There's on vulnerably . Technically nothing new, but it's less clear here because we obfuscate persistent data via this class.
```py
c = MASEventContext.get()
c.this_is_bad = object()
c.this_is_worse = MyCustomClass()
c.oh_my_god = get_unpickable_data()
```
That can lead to data corruption.

Monika After Story/game/event-handler.rpy
```diff
@@ -2004,44 +2471,45 @@ init python:
         evhand._lockEventLabel(evlabel, eventdb=eventdb)
 
 
+    @store.mas_utils.deprecated("MASEventList.push")
     def pushEvent(event_label, skipeval=False, notify=False):
```
Booplicate, line 2475: Maybe we could remove all our uses of the deprecated funcs?

Monika After Story/game/event-handler.rpy
```diff
@@ -2004,44 +2471,45 @@ init python:
...
+        MASEventList.push(event_label, skipeval, notify)
 
-        persistent.event_list.append((event_label, notify))
-
-        if skipeval:
-            mas_idle_mailbox.send_skipmidloopeval()
-        return
 
+    @store.mas_utils.deprecated("MASEventList.queue")
     def queueEvent(event_label, notify=False):
```
Booplicate, line 2496: Maybe we could remove all our uses of the deprecated funcs?

Monika After Story/game/event-handler.rpy
```diff
@@ -2240,8 +2657,8 @@ init python:
 
         RETURNS: True if in event list, False if not
         """
-        for ev_data in persistent.event_list:
-            if ev_data[0] == event_label:
+        for item in MASEventList.iter():
+            if item.evl == event_label:
                 return True
 
         return False
```
Booplicate, line 2664: Can't this be replaced with `return mas_findEVL(event_label) != -1`? So we don't repeat our code and will have to maintain later.

Booplicate (changes requested):

Monika After Story/game/event-handler.rpy
ThePotatoGuy: there's alot and pretty much no places need context yet so we could just change as we modify the surrounding code. We have to keep this as a non-exception deprecation because of submods anyway.

Booplicate: I know, it's just we use quite a few deprecated funcs and they're flooding logs. I was thinking about opening a PR to remove all uses of deprecated funcs, so we can leave it until then if you want.

Monika After Story/game/event-handler.rpy
ThePotatoGuy: okay latest changes should guardrail against bad context data (will update pr text with the new stuff)

Monika After Story/game/event-handler.rpy
ThePotatoGuy: yeah lets do that in a subsequent pr

Monika After Story/game/event-handler.rpy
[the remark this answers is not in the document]
ThePotatoGuy: >I still don't feel safe knowing we rely on precise arg position. 

I mainly feel fine with relying on arg position because `event_list` is structured on position, so even if some of the `EventListItem` functions don't rely on position, positioning must be relied on somewhere in there for the thing to work. If `event_list` tuples were changed to dicts, then I think it would be reasonable for `EventListItem` to use more named-based/dict-based handling (and I don't think there are compelling reasons to change away from tuples).

>final decision 
You made great points overall, and I agree that my solution is not as robust in comparison, but I still think this implementation is sufficient for the current use case. If the use case changes (mainly adding dozens of properties to event list items or making event list item manipulation a norm), then we can revisit this - this is luckily abstracted enough that if it changes again, it wouldn't affect public API usage anywhere.

Monika After Story/game/event-handler.rpy
```diff
@@ -1713,6 +1915,421 @@ init python:
...
+                context = ""
+            else:
+                context = cls.__CTX_CTX.format(
+                    store.mas_globals.this_ev.eventlabel
+                )
+
+            if recur_error:
+                # log recursion error
+                store.mas_utils.mas_log.error(cls.__ERR_RECUR.format(
+                    type(thing),
```
Booplicate, line 1986: This is a bit cleaner.
```suggestion
                    type(thing).__name__,
```

Monika After Story/game/event-handler.rpy
```diff
@@ -1713,6 +1915,421 @@ init python:
...
+                    context
+                ))
+                    
+            else:
+                # log disallowed object
+                store.mas_utils.mas_log.error(cls.__ERR_NON_PICKLE.format(
+                    type(thing),
```
Booplicate, line 1993:
```suggestion
                    type(thing).__name__,
```

Booplicate (changes requested):

Monika After Story/game/event-handler.rpy
```diff
@@ -1713,6 +1915,421 @@ init python:
...
+
+            if 0 <= loc < len(persistent.event_list): # just in case
+                persistent.event_list.pop(loc)
+
+            MASEventList._set_current(item)
+
+            return item
+
+        @staticmethod
+        def push(event_label, skipeval=False, notify=False, context=None):
```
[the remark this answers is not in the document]
ThePotatoGuy: you can create context objects with a dict like `MASEventContext( <dict> )` so thats covered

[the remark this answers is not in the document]
Booplicate: Yeah, I know. It'd be more of syntactic sugar, but it's fine as is, too.

Booplicate (approved):
Works perfectly!

multimokia (approved):
aboom

Booplicate (approved):
I'm just being dum, nothing new.

multimokia (approved):
Slow clap processor

PR #8639, opened 2022-01-25
New horror story

erintesden:
A short horror story and some Monika's commentary about it.
Still need the coding for the lights out effect and linking the dialogue with the list of horror stories.

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -17755,10 +17755,10 @@ init 5 python:
 label monika_last_man:
     m 7eua "Alright, [player], this will be a short one."
 #adding efect of room getting dark
-    m 1tua "…"
+    m 1tua "..."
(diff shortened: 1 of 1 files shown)


Monika After Story/game/script-topics.rpy
```diff
@@ -17739,3 +17739,36 @@ label monika_hedonism:
...
+        Event(
+            persistent.event_database,
+            eventlabel="monika_last_man",
+            category=["literature"],
+            prompt="The last man on Earth",
+			pool=True
+            random=False
+        )
+    )
+
```
multimokia, line 17754: You'll want to put stories in `script-stories.rpy`, that's where we keep all our story topics.

Horror stories are in the second half of the file and have two labels that are called to set up the lights out and one to clean that up.

multimokia (changes requested):

Issue #8655, opened 2022-01-28
Bad name false positive ⁠— 'Familiarity' is matched by 'liar' subpattern in bad names list


dreamscached:
Recently, a user in #tech-support Discord channel ([message](https://discord.com/channels/372766620977725441/393978515802030091/936532211946098739)) has reported an issue with Monika unable to call them 'Familiarity'.

I did a bit of a research and found out that `Familiarity` is matched by `liar` simple subfilter ([code](https://github.com/Monika-After-Story/MonikaModDev/blob/6008eff0828f6c0b4dbeec0c19124bd2d0911be3/Monika%20After%20Story/game/script-story-events.rpy#L309)).

I'll make a PR for it if this issue isn't reported yet.

dreamscached:
P.S. Quickly glancing at the pattern list I noticed `Skilled ` also results in false positive and is matched by `kill` subpattern as well.

ThePotatoGuy:
[closing as fixed]

PR #8656, opened 2022-01-28
Bad name false positive fix

dreamscached:
See #8655.


ThePotatoGuy (approved):

ThePotatoGuy:
merging this in - its small enough to be fine with simple testing.

PR #8745, opened 2022-02-15
organized reset code

ThePotatoGuy:
on hold until #8619 

# TODO
- [x] merge 8619 into this
- [x] finish the 2nd half of the ch30 code
- ~~add d25 and o31 deco reset code~~ not doing this anymore

# Key Changes
* itemizes the `ch30_reset` code into individual functions in the `mas_reset` store - also we should expect to make unit test cases for these later.
* adds decorator `mas_reset.ch30_reset` to add code to the reset label.
    * all submods that want to do this should stick with the default priority or go higher (0+).
    * **This is breaking changes** - submods that hook into `ch30_reset` may now find their code running **after** our reset code instead of **before**.
    * final is in the label, as it should be for code that **must** run after everything else.
* ~~adds general d25 and o31 deco resets~~

## changes to execution order
* ACS template application has been moved from after ` name eggs` to after `game unlocks`
* ahoge reset code moved to after promise ring hotfix
* startup plushie logic moved from after `affection freeze reset` to after `game unlocks`
* selector prompt check moved from after `file reactions fix` to after `game unlocks`
* file reaction affection resets moved from after `event list cleaning` to after other `file reaction` checks
* gender and name rests moved from after `TOD vars set` to after `event list cleaning`

## changes to code
* playtime reset code changed to use total playtime API - small simplification as with no more None check
* functionized the `mas_after_bath_cleanup_change_outfit` label
* split show/hide visuals from show/hide sprite objects for holiday code

# Testing
* verify reset code is functionally the same to before


ThePotatoGuy:
moving to next rel but this should be in first next time because anything new that changes the reset code will benefit from this.

multimokia (dismissed):

Monika After Story/game/script-ch30.rpy
```diff
@@ -826,6 +826,585 @@ init python:
             setattr(store, word, value)
 
 
+init 995 python in mas_reset:
+    # define the decorator for reset code
+
+    import store.mas_submod_utils as mas_submod_utils
+
+    def ch30_reset(priority=9999):
```
Booplicate, line 834: Maybe we could use negative priority for our resets? So people don't have to work with 5 digits.

ThePotatoGuy: function plugins run before the code in a label, so submod's reset code could have expected our reset code to run after. the default priority of function plugins is 0, so by picking priority numbers larger than 0, existing submods using default priority wont have that assumption broken.

Booplicate: This is a very niche case
- means people used fp for this label
- expected it to run first
- didn't set any priority, if they wanted to make sure it's being executed as early as possible, they would probably have set it? At least so it doesn't interfere with other fp callables

Those are factors that are very unlikely to come together if at all. Yet they are going to define our code in future, because we would have to do `@ch30_reset(10000)` (for example I needed to register a few funcs), it's completely unclear where 10k came from.

multimokia: I agree, sticking within the -999 - 999 range is more inline with the style of the rest of our codebase (i.e. init levels), especially in a paradigm where using inits in the 4 digit range (and above) is discouraged. It makes the most sense to use more traditional values, imo

ThePotatoGuy: I don't agree that this would be a niche case, but since we've advertised the next few updates can have breaking changes, this will be one of them.

multimokia (approved):

Monika After Story/game/script-ch30.rpy
```diff
@@ -826,6 +826,585 @@ init python:
...
+            "chess": "mas_unlock_chess",
+            "hangman": "mas_unlock_hangman",
+            "piano": "mas_unlock_piano",
+            "nou": "mas_reaction_gift_noudeck",
+        }
+        store.mas_unlockGame("pong") # always unlock pong
+
+        for game_name, game_startlabel in game_unlock_db.iteritems():
+            # unlock if we've seen the label
+            if store.mas_getEVL_shown_count(game_startlabel) > 0:
```
Booplicate, line 968: This won't unlock nou since `mas_reaction_gift_noudeck` isn't an event. Or maybe it's because we don't register reaction if you've given the game, so `mas_getEV` returns `None`.

Booplicate: Maybe use `mas_seenEvent`? Altho it also checks current even list which might not be ideal.
Alternatively, add extra handling for nou like we have for pong.

Booplicate (approved):

multimokia (approved):

PR #8909, opened 2022-02-23
Taking Monika outside

CinnabunDreams:
It would be fun if when you chose the "I'm going to take you somewhere" option after saying goodbye to Monika, she'd ask where you're going and you'd get a few options like "Park", "Cafe", "Movies", etc. 

And if you weren't taking her to any of the set places, you could add an option that was like "Somewhere else" and she could respond with something vague enough that would make sense like "Anywhere we go together will be fun!"


CinnabunDreams:
> It would be fun if when you chose the "I'm going to take you somewhere" option after saying goodbye to Monika, she'd ask where you're going and you'd get a few options like "Park", "Cafe", "Movies", etc.
> 
> And if you weren't taking her to any of the set places, you could add an option that was like "Somewhere else" and she could respond with something vague enough that would make sense like "Anywhere we go together will be fun!"

Omg I think I did this wrong-- I never use GitHub, didn't mean to merge 110 commits

Booplicate:
Pull Requests are for adding new content, you probably meant to use Issues (which are for bugs reports/suggestions).

PR #8930, opened 2022-02-26
[BRB] - "I'm going to sleep."

lunulae:
This may need some major reworks later on for timeconcern, but since a few requests have been made to differentiate between sleep and naps I figured I'd take a swing at it.

TODOs:
- [ ] testing/review
- [ ] do we want a player-is-sick path? I kinda want a sick path but I'm not sure.


lunulae:
whoops wrong branch

PR #8940, opened 2022-03-01
[New Spritepack] Dress (Dark blue, sparkly)

kkrosie123:
This is one of many new, upcoming spritepacks for future updates.

TODO:
- [x] less blurry thumbnail
- [ ] perhaps add other variants (Debating if they should be in the same PR)


kkrosie123:
Thumbnail done. _Dress (Dark blue, sparkly)_ is ready for review and technically complete.

ThePotatoGuy (changes requested):
is the vcut crossed straps tank top supposed to be part of this?

kkrosie123:
> is the vcut crossed straps tank top supposed to be part of this?

There was a name error - it was called `brair` instead of `briar`, just decided to fix it, so yeah

ThePotatoGuy:
that's gonna mess up user installs if they install all spritepacks at once. But should be easy to diagnose and resolve so probably ok.

ThePotatoGuy (approved):

kkrosie123:
Ah. My bad.

lunulae (approved):
Tested, if we're still good with keeping the vcut changes this is ready to go~

PR #8996, opened 2022-03-21
Cleanup pronouns

Booplicate:
### Changes:
- the pronouns map is global now
- don't duplicate the keys anymore (I think this way it's cleaner, but I can revert that change)
- use `setattr` to dynamically add new words from the map (easier to maintain)
- `mas_set_gender` is deprecated, we don't set gender, we set pronouns. Now the old label calls new new function - `mas_set_pronouns`
- New function: `mas_set_pronouns` accepts the `key` param, by default `None`. if `None`, uses the `persistent.gender` to access the pronouns map
- New unittest to test this

### Testing:
- run test, verify it works

### DIFF
--- a/Monika After Story/game/dev/dev_unittest.rpy
+++ b/Monika After Story/game/dev/dev_unittest.rpy
@@ -2743,18 +2743,28 @@ label dev_unit_test_mas_set_pronouns:
 
     python hide:
         pronouns_tester.prepareTest("Validate map size and keys")
-        map_size = None
-        for sub_map in store.MAS_PRONOUN_GENDER_MAP.values():
+        expected_map_size = None
+        for key, sub_map in store.MAS_PRONOUN_GENDER_MAP.items():
             sub_map = dict(sub_map)
 
-            if map_size is None:
-                map_size = len(sub_map)
+            if expected_map_size is None:
+                expected_map_size = len(sub_map)
             else:
-                pronouns_tester.assertTrue(map_size == len(sub_map))
+                sub_map_size = len(sub_map)
+                pronouns_tester.prepareTest(
+                    "Validating map size for key {}, expected {}, got {}".format(
+                        key,
+                        expected_map_size,
+                        sub_map_size
+                    )
+                )
+                pronouns_tester.assertEqual(expected_map_size, sub_map_size)
 
-            pronouns_tester.assertTrue("M" in sub_map)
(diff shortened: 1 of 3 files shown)


ThePotatoGuy (changes requested):
can you make the prounoun setting a function (prob ch30 api) and just have the label call that?

Booplicate:
This should do

Monika After Story/game/dev/dev_unittest.rpy
```diff
@@ -2734,3 +2735,43 @@ label dev_unit_test_strict_can_pickle:
...
+            sub_map = dict(sub_map)
+
+            if map_size is None:
+                map_size = len(sub_map)
+            else:
+                pronouns_tester.assertTrue(map_size == len(sub_map))
+
+            pronouns_tester.assertTrue("M" in sub_map)
+            pronouns_tester.assertTrue("G" in sub_map)
+            pronouns_tester.assertTrue("X" in sub_map)
```
ThePotatoGuy, line 2757: these are checking if the sub map is the same length as other maps, and the main 3 gender types have an entry in the sub map, correct? to make it easier for debugging, we should have a prepare test line for the size check, and then a prepare test line before the gender types checks. Can use the pronoun (key) in the prepare test text as well. That way when Monika says a test failed, it won't always be "Validate map size and keys" when only one part of the map has an issue.

Monika After Story/game/dev/dev_unittest.rpy
```diff
@@ -2734,3 +2735,43 @@ label dev_unit_test_strict_can_pickle:
...
+        for gender in ("M", "G", "X"):
+            # Verify the func works
+            mas_set_pronouns(gender)
+
+            fallback = object()
+            for word, sub_map in store.MAS_PRONOUN_GENDER_MAP.items():
+                # Verify the func set correct values
+                global_value = getattr(store, word, fallback)
+                map_value = sub_map[gender]
+                pronouns_tester.assertEqual(global_value, map_value)
```
ThePotatoGuy, line 2769: same thing I said above.

ThePotatoGuy (changes requested):

ThePotatoGuy (approved):

Legendkiller21 (approved):

ThePotatoGuy (approved):

Issue #8308, opened 2021-11-27
Graphics renderer stuck on Software


swindlesmccoop:
Just installed the latest version of the mod, and every time I start the game, I get a message saying that I'm using software rendering, even though in the settings it is set to OpenGL. Not sure what would cause this because I'm pretty sure I have all of the necessary drivers installed. Is there a list of packages required that I should check to make sure I fulfill the requirements?

System: Linux-5.15.4-artix1-1-x86_64-with-glibc2.2.5
GPU: Nvidia GTX 1660Ti

multimokia:
Can you send your `log.txt` file please?

swindlesmccoop:
Yep. It even says it's using GL renderer.
[log.txt](https://github.com/Monika-After-Story/MonikaModDev/files/7612542/log.txt)

multimokia:
This is literally the exact same issue I had before when I ran Kubuntu. (This was interestingly only an occurrence on my desktop, running a 3080, and not my laptop which has an intel integrated gpu + a discrete gpu, but that wasn't used for MAS if I recall)

Weird issue with how it handled my GPU entirely. I'm not sure why but the only thing that fixed it at least in there was modifying my install of MAS to run renpy7.4, which offers the *GL2* renderer instead of GL1.

**However** MAS is _kinda_ unstable here and there under it. (Chess is basically broken)

That said, I migrated to Manjaro and the issue is just gone. So I'm wondering if there's some low level deps we were both missing for our cards.

Either way this is 100% something that should be investigated

multimokia:
As for it using software, Ren'Py has poor logs and it actually _silently_ falls back to `software` after announcing `Textures are not rendering properly.`

swindlesmccoop:
Figured out a little bit about the issue. Basically, one of the textures uses an ancient color pallet (whatever a8 is) and it probably just can't render it properly, for some reason killing the entire GL render pipeline. Apparently is an actual Ren'Py issue being triggered by this texture. Most likely related to https://github.com/renpy/renpy/issues/2162.

Booplicate:
Note: switching from (or to) fullscreen mode and back might (temporary) fix this issue.

swindlesmccoop:
[...]

PR #9008, opened 2022-03-27
 Software rendering kwin meta-topic POC/RFC

jvyden:
Hello! @multimokia suggested in #8308 that there could potentially be a meta topic about software rendering being on and working with the player to resolve that issue.

I decided to write a proof of concept for this.

This demonstrates a couple of things:

1. We can easily read the current renderer at runtime.
2. It's also possible to get running processes for Linux. It works a little bit differently than Windows, as in it won't check a list of all running process, rather it will use the error code output of `pidof [program_name]`. If this utility is available on Mac by default let me know and I can enable that branch for Mac. Right now I'm unsure, though.

I mark this as a draft for a few uncertainties, and I guess this counts as a TODO as well:

- [ ] Add an automatically firing event if the game is in software rendering.

I'm not quite sure how to approach this one as this is my first time working with this codebase.

Firstly, where to put it? I see similar things (for example `monika_rpy_files`) in `script-story-events.rpy` but it's not really a story event, but it's also not a normal topic so it doesn't belong in `script-topics.rpy`, either. Perhaps a new file for meta topics such as this one would be better?

Next, how would I fire it in the first place? `monika_rpy_files` is fired right after rebuilding event lists, but doing the same here doesn't feel right to me.

- [ ] Dialogue

I'm not terribly confident in my writing skills, lol. I have a general idea of what needs to be said but it's a bit hard for me to simplify it for a user in addition to writing as Monika.

- [ ] Decide when to fire the event.

When should we fire this? If the game is running in software mode at all? Just on Linux? On Linux when we're using KWin? Not quite sure on this one, this will probably have to be decided after the dialogue is written.

- [ ] Test other Desktop Environments

Right now I've only seen the behavior described in #8308 on KDE Plasma. However this could totally happen on any other DE. Some notable ones to test would be GNOME, XFCE, Cinnamon, and LXDE.

Alright, this is starting to get long and I think I've touched on everything there is about this PR. As you may have guessed (:trollface:) I'm open for comments and questions, so feel free to ask anything!

### DIFF
--- a/Monika After Story/game/definitions.rpy
+++ b/Monika After Story/game/definitions.rpy
@@ -5500,6 +5500,20 @@ init -1 python:
 
         RETURNS: True if a proccess in proc_list is running, False otherwise
         """
+        # Should be fine to check this here right?
+        if renpy.linux:
+            for proc in proc_list:
+                import subprocess
+                try:
+                    output = subprocess.call(["pidof", proc])
+                    
+                    # 0: Success
+                    # 1: Did not find
+                    if output == 0:
+                        return True
+                except:
+                    pass
+        
         running_procs = get_procs()
         if len(running_procs) == 0:
             return False
--- a/Monika After Story/game/dev/dev_renderer-test.rpy
+++ b/Monika After Story/game/dev/dev_renderer-test.rpy
@@ -1,0 +1,29 @@
+init 5 python:
+    addEvent(
+        Event(
+            persistent.event_database,
+            eventlabel="dev_mas_renderer",
+            category=["dev"],
+            prompt="SHOW RENDERER INFO",
+            pool=True,
+            unlocked=True
+        )
+    )
+    
+label dev_mas_renderer:
+    python:
+        renderer = mas_getRendererName() # string
+        isSoftware = mas_isSoftwareRendering() # bool
+        kwinRunning = is_running(['kwin_x11', 'kwin_wayland']) # bool
+        
+    m "Sure! RenPy is using the \"[renderer]\" renderer..."
+    if isSoftware:
+        m "...which means it's using software rendering. Ahaha... I feel bad for your CPU, [player]."
+    else:
+        m "...which means I'm not on software rendering mode! Yay!~"
+        
+    if renpy.linux:
+        if kwinRunning:
+            m "It also looks like you're using KWin."
+        else:
+            m "You're on Linux, but you don't appear to be using KWin."


Booplicate:
Tom [mentioned](https://github.com/renpy/renpy/issues/2162#issuecomment-1073151160) that this issue has been fixed with the gl2 renderer. This renderer will (if not already) become default in next few RenPy releases.

Knowing that and the fact we're working on upgrading the engine we use, I'm not settled on adding this. You can probably already see it yourself - this is going to big quite big. After you implement this, we'd need to test it as well. If after a few months we migrate to r8, these additions would become irrelevant since the issue has been fixed. I'll also say that while there's a dozen of people who have/had the issue, it's only a dozen of people. Most users (even thus that use linux), won't see it. Those facts makes me doubt in this idea.

If the bug still exists in r8, then I'd be more open to adding this, but even then it's not *too* common, and perhaps it'd be more appropriate if we could report the bug to [RenPy](https://github.com/renpy/renpy) so it gets fixed.

This is not a no, but it's something to think about before you start working on this project.

jvyden:
Yeah, that's a fair point. This *is* a fairly high-effort change for a rather small problem, especially because of all the other problems this brings up.

Still though, I've given it a bit of thought and I think implementing this is still worth working on at least to an extent for two reasons:

First, the issue annoyed me LOL

More seriously, though: I think this could be a good way for me to familiarize myself with the codebase/how you guys do things around here since it touches what seems to be quite a large portion of the project. Oh, and not to mention I could improve my writing skills for sure if I decide to go that route, which honestly doesn't sound too bad now so I probably will.

Obviously, you're the maintainer, so it's ultimately up to you to decide, but I'm comfortable working on this; I just need to be pointed in the right direction.

*Wew, I spent way too long writing this lol :upside_down_face:*

ThePotatoGuy:
First, answering `Firstly, where to put it?`. As a meta topic, this probably would be something that chibika should handle, but since it doesn't crash the game or cause corruption, it would probably be a chibika letter with Monika being telling the player about it.

But I agree with boop, I don't think this is worth having as a meta topic, especially with the low number of users that would see this. 

The changes to `is_running` could be useful in general, though.

jvyden:
> But I agree with boop, I don't think this is worth having as a meta topic, especially with the low number of users that would see this.

Well, that makes two maintainers not wanting this, so I'll close for now lol. It's a good thing I made this draft before I put all the effort in.

> The changes to is_running could be useful in general, though.

Alright, I'll make a seperate PR for that then.

PR #9010, opened 2022-03-28
Linux support for `is_running()`

jvyden:
I briefly went over this change in #9008, but for clarity's sake I'll explain it again here in a little bit more detail:

This works a bit differently than Windows, as in it won't check a list of all running process, rather it will use the error code output of `pidof [program_name]`. 

I did look into getting Linux support for `get_procs()` but it didn't seem like there would be a way to get it in a clean way since `ps`, the only definitely-on-the-system-by-default utility I can think of shows full paths and arguments no matter what, so I left it. This could be problematic in the future if you or a submod decides to use `get_procs()`, however I don't see any other usages of it other than the one in `is_running()`.

If this command is available on Mac by default on a fresh installation let me know, and I can go ahead and enable that branch for Mac. Right now I don't have access to a Mac that doesn't have stuff like Homebrew on it, though, so I wouldn't be able to tell.

### DIFF
--- a/Monika After Story/game/definitions.rpy
+++ b/Monika After Story/game/definitions.rpy
@@ -5510,16 +5510,16 @@ init -1 python:
                     # 1: Did not find
                     if output == 0:
                         return True
-                except:
-                    pass
-        
-        running_procs = get_procs()
-        if len(running_procs) == 0:
-            return False
-
-        for proc in proc_list:
-            if proc in running_procs:
-                return True
+                except Exception:
+                    return False # If it didn't work here, it probably won't work again, so lets bail here.
+        elif renpy.windows:        
+            running_procs = get_procs()
+            if len(running_procs) == 0:
+                return False
+    
+            for proc in proc_list:
+                if proc in running_procs:
+                    return True
 
         # otherwise, not found
         return False
--- a/Monika After Story/game/definitions.rpy
+++ b/Monika After Story/game/definitions.rpy
@@ -5500,6 +5500,19 @@ init -1 python:
 
         RETURNS: True if a proccess in proc_list is running, False otherwise
         """
+        if renpy.linux:
+            for proc in proc_list:
+                import subprocess
+                try:
+                    output = subprocess.call(["pidof", proc])
+
+                    # 0: Success
+                    # 1: Did not find
+                    if output == 0:
+                        return True
+                except:
+                    pass
+        
         running_procs = get_procs()
         if len(running_procs) == 0:
             return False
--- a/Monika After Story/game/definitions.rpy
+++ b/Monika After Story/game/definitions.rpy
@@ -5500,7 +5500,15 @@ init -1 python:
 
         RETURNS: True if a proccess in proc_list is running, False otherwise
         """
-        if renpy.linux:
+        if renpy.windows:        
+            running_procs = get_procs()
+            if len(running_procs) == 0:
+                return False
+    
+            for proc in proc_list:
+                if proc in running_procs:
+                    return True
+        elif renpy.linux:
             for proc in proc_list:
                 try:
                     output = subprocess.call(["pidof", proc])
@@ -5511,14 +5519,6 @@ init -1 python:
                         return True
                 except Exception:
                     return False # If it didn't work here, it probably won't work again, so lets bail here.
-        elif renpy.windows:        
-            running_procs = get_procs()
-            if len(running_procs) == 0:
-                return False
-    
-            for proc in proc_list:
-                if proc in running_procs:
-                    return True
 
         # otherwise, not found
         return False


Monika After Story/game/definitions.rpy
```diff
@@ -5500,6 +5500,19 @@ init -1 python:
 
         RETURNS: True if a proccess in proc_list is running, False otherwise
         """
+        if renpy.linux:
+            for proc in proc_list:
+                import subprocess
```
multimokia, line 5505: At this point, the subprocess module is already imported globally. We don't need to import it again.
```suggestion
```

multimokia (changes requested):
Just a minor thing here, side from this it all looks fine to me.

Monika After Story/game/definitions.rpy
jvyden: Ah, true. Not sure how this slipped in.

Monika After Story/game/definitions.rpy
```diff
@@ -5500,6 +5500,18 @@ init -1 python:
...
+        if renpy.linux:
+            for proc in proc_list:
+                try:
+                    output = subprocess.call(["pidof", proc])
+
+                    # 0: Success
+                    # 1: Did not find
+                    if output == 0:
+                        return True
+                except:
```
Booplicate, line 5512: We should explicitly catch an exception type otherwise we might get stuff like `KeyboardInterrupt`.

So `except Exception:`

Also unsure if we really want to keep making subprocess calls when we failed one (esp since we don't log it). So maybe just `return False`

Monika After Story/game/definitions.rpy
```diff
@@ -5500,6 +5500,18 @@ init -1 python:
...
         """
+        if renpy.linux:
+            for proc in proc_list:
+                try:
+                    output = subprocess.call(["pidof", proc])
+
+                    # 0: Success
+                    # 1: Did not find
+                    if output == 0:
+                        return True
```
Booplicate, line 5511: This can be just `return bool(output)`

Booplicate (changes requested):
Also I'd put the windows part into a `elif renpy.windows:` block

jvyden:
@Booplicate Noticed your comment about just returning `bool(output)`. I'm assuming you deleted that comment because it was wrong, but if you want I can still add a code comment or perhaps make it more explicit with an `else continue;` since it confused me too, lol.

Booplicate:
Nah, I just didn't realise we want to process all items in the list.

Monika After Story/game/definitions.rpy
```diff
@@ -5500,13 +5500,25 @@ init -1 python:
...
-                return True
+        if renpy.linux:
+            for proc in proc_list:
+                try:
+                    output = subprocess.call(["pidof", proc])
+
+                    # 0: Success
+                    # 1: Did not find
+                    if output == 0:
+                        return True
```
multimokia, line 5511: Looking on this again, may have found a more cross-platform working solution. The following *should* work on both Linux and MacOS:
```suggestion
                    output = re.search(
                        r"\/?{}(\s+--?\w+)*".format(proc),
                        bytes.decode(subprocess.check_output(["ps", "-A"])),
                        re.IGNORECASE
                    )

                    # not None means we found a match
                    if output is not None:
                        return True
```

Monika After Story/game/definitions.rpy
```diff
@@ -5500,13 +5500,25 @@ init -1 python:
...
         RETURNS: True if a proccess in proc_list is running, False otherwise
         """
-        running_procs = get_procs()
-        if len(running_procs) == 0:
-            return False
-
-        for proc in proc_list:
-            if proc in running_procs:
-                return True
+        if renpy.linux:
```
multimokia, line 5503: We can probably put the windows check in this position, then have the following stuff in an `else` to fallback for Mac/Linux then

multimokia (changes requested):

Monika After Story/game/definitions.rpy
jvyden: Might be helpful to add a comment on the Regex used on line 5507.

Also, this might not work for Linux; the outputs between `ps -A` on macOS and Linux vary a fair bit as you can see here:

image

In particular, I'd like to note the following things:

- macOS' `ps` shows paths and arguments. Linux's `ps` does not show that; instead it shows the "title" of the process, not the actual process name.
- Atleast on my Arch Linux install, `ps` is from `procps-ng`. macOS doesnt even show a version, which is odd. They're definitely differing versions.

multimokia: Oh I noticed, I tested against it, however when I tested on Catalina at least, was possible to at least catch things i.e. Finder/Safari, though do note I didn't try much further. Also caught `trustd` as a test.

Also backup tested on Manjaro which also seeemed to work fine. It does bug the absolute hell outta me that Mac's is so...painful lol

multimokia: That said, was going to run final tests on there if this went in. (Mac install is laggy af w/o any GPU passthru)

multimokia: Yeah, if anything we can probably leave a `#NOTE:` on it mentioning it could potentially have issues on Mac due to a more verbose proc name.

jvyden: Hmm, the more I think about it the more I feel it would be best if we left this implementation for macOS use only, since the problem specifically is that Linux is the odd one out, not macOS.

This implementation will work great with macOS, but factoring in Linux with it starts making it a bit annoying to deal with in some cases due to the whole process title thingamajig.

`pidof` works great since it's also looking for the name of the process, not anything else (the exact behavior we need here!) I'd also bet on it being faster.

If you're worried about this function getting rather long, I bet you could move some of it into `get_procs()` and use the same method as Windows.

jvyden: Also if you'd like I can test using my real mac, though it's been tampered with by a bunch of development tools and by Homebrew.

Booplicate: @jvyden I think it's fine to split macos and linux code, will you push the changes?

jvyden: Sure, I will get to this at some point, been busy lately haha

jvyden:
Closing as I don't think I'm ever going to get back to this as I've stopped playing this. Feel free to take my branch over if interested (consider this a sign-off)

PR #9040, opened 2022-04-12
Aff system cleanup

Booplicate:
Some cleanup + a few new functions for better affection control.

### Testing:
- verify general stability, no aff reset, update script is working
- run unittests
- verify new caps are working
- - also test caps during holidays
- verify `_mas_revertFreshStart`, `_mas_shatterAffection`, `_mas_doFreshStart` are working as intended
- verify gain and lose are randomised with the expected range
- - min gain should be 0+
- verify you can't pass invalid values into the aff functions (like a negative number in aff gain)
- verify new default aff values are good
- - default gain
- - default lose
- - default % lose
- verify bank grants aff and decays as expected
- verify backups are working, try to corrupt affection
- verify apology flag still works as before
- verify spamming compliments doesn't give affection (there's a cooldown)


ThePotatoGuy (changes requested):
can you make unit tests for the aff encoding and decoding? Also, one of these tests should be testing a whole range of affection values (like from -1m to 1m) to make sure we don't end up with the problems from previous aff encoding.

ThePotatoGuy:
^also by that last unit test req, something that makes sure that the created encoded value is something that doesn't break when pickled as a string (aka data become persistent). I really do not want to fuck that up again.

ThePotatoGuy:
also make sure to update update script

ThePotatoGuy:
get organized reset in first please

Monika After Story/game/script-affection.rpy
```diff
@@ -249,6 +299,699 @@ init -900 python in mas_affection:
...
+
+        old_data = persistent._mas_affection
+        if old_data is None:
+            persistent._mas_affection_version += 1
+            return
+
+        new_data = list()
+
+        aff = old_data.get("affection", 0.0)
+        if aff > 100000:
```
ThePotatoGuy, line 775: was this supposed to be 1m or 100k?

Monika After Story/game/script-affection.rpy
```diff
@@ -1675,199 +2319,276 @@ init 20 python:
...
+                NOTE: the modifier is being applied AFTER min_amount
+                (Default: 1.0)
+        """
+        if fraction is None:
+            fraction = _get_current_aff_fraction_lose()
+        if min_amount is None:
+            min_amount = _get_current_aff_lose()
+
+        if fraction < 0.0 or min_amount < 0.0 or modifier < 0.0:
+            raise ValueError("Invalid value for affection")
```
ThePotatoGuy, line 2531: same as other exceptions comments

ThePotatoGuy (changes requested):
can you also make unit tests for the aff functions please

Monika After Story/game/script-affection.rpy
```diff
@@ -249,6 +299,699 @@ init -900 python in mas_affection:
...
+            data[2] = 0.0
+            data[3] = 0.0
+            data[4] = now_
+
+        frozen = data[2] >= 7.0
+
+        # Sanity checks
+        amount = float(amount)
+        if amount <= 0.0:
+            raise ValueError("Invalid value for affection: {}".format(amount))
```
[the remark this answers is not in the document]
Booplicate: Generally, don't mind if we just log the error, but I want you to consider this again because if we screw up, we would know it right away - we would get an exception while testing. If it'd just get logged, we may miss it entirely.

ThePotatoGuy (approved):

ThePotatoGuy:
oh wait i forgot to run the unit tests wait

ThePotatoGuy:
ok is good

ThePotatoGuy (changes requested):
run aff pp after transfer

ThePotatoGuy (dismissed):
probably ok if update script is ok

ThePotatoGuy (approved):

ThePotatoGuy (dismissed):

Monika After Story/game/script-ch30.rpy
```diff
@@ -2004,6 +2000,8 @@ label ch30_preloop:
         set_keymaps()
 
         persistent.closed_self = False
+        # TODO: this var is always True,
```
ThePotatoGuy, line 2003: its set to false if the user purposefully quits the game - this is how we determine if the game crashed when `closed_self` is false. 

to check if the game just crashed, the `just_crashed` var is set a few labels above (line 1848) - or combine a not `_mas_game_crashed` with not `closed_self`

ThePotatoGuy (changes requested):

Monika After Story/game/script-ch30.rpy
Booplicate: The name is still misleading, even if user didn't crash, we get `True` in `_mas_game_crashed`. Since the purpose is actually check for force quit, should be used another var `_mas_force_quit` or similar.

ThePotatoGuy: ok fair point

ThePotatoGuy (approved):

Legendkiller21 (approved):

multimokia (approved):

PR #9079, opened 2022-04-28
[Story] - "The Well of Madness"

MrVavlo:
Added a parable by [Paulo Coelho](https://paulocoelhoblog.com/2013/05/15/the-well-of-madness/) to the stories list.

While I'm not going to make a draft, yet there are still questions I would like to raise:

1) ~~This parable has several versions - to the point that I'm not sure if the parable I've used _is_ original.
(For example - there's another parable popular in my country where's the only difference is that instead of the wizard, there's an angel alerting one of the villagers about the poison in the well).
So should I stick with the Coelho's version? Or something else?~~
Decided to stick with Coelho's version, no matter if it's original or not.

2) Should Monika react to the story somehow? (examples: projecting her story onto it; simply deconstructing the well situation, etc.)
Or just end the story without comments?

To-do list:

- [x] Decide which version of the story to use
- [ ] Decide if Monika will comment the story at all or not

### DIFF
--- a/Monika After Story/game/script-stories.rpy
+++ b/Monika After Story/game/script-stories.rpy
@@ -1037,6 +1037,35 @@ label mas_story_mindthegap:
     m 1ekbla "This story reminds me to treasure every moment, and not to take any piece of our time together for granted."
     m 1dkblu "I'll always treasure you, [player]."
     return
+    
+init 5 python:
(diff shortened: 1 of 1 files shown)


jmwall24:
Closing as we don't really like the message it sends.

PR #9081, opened 2022-04-28
fixed typo

Elckarow:


multimokia:
This typo exists in a docstring for a function and not dialogue. As such it's not really a priority to fix since it's understandable regardless of the typo.

Issue #3000, opened 2018-11-06
3000 GET


ThePotatoGuy:

multimokia:
First.

jmwall24:
this feels like aboos

pianostick:
conglaturation

DankManWithThePlan:
_Epic_

monikLover:
Will report for issue theft. :oncoming_police_car:

ThePotatoGuy:
[closing until we look into this]

Issue #2400, opened 2018-09-15
[Suggestion] Can you scare someone for me?


Chrissenpai:
Since Halloween is coming near, i would like to scare somebody, i could make good use of monika's jump scare for that, so you choose an option that says, "i want you to scare someone for me", or something among those lines, and she says she's up to it, in what amount of time do you want me to show up?, and you decide in what amount of time she does it in any minute time you digit on, then she says alright and closes the game and the clock ticks, in case you want to cancel you open the game before she activates and you tell her to abort, some dialogue could be filled in.

multimokia:
Having MAS startup automatically requires an external program, which has been rejected numerous times. See #1829, point 5 on TPG's arguments against that.

Issue #2000, opened 2018-08-05
Inconsistency in Hugging during rain/no rain


Skeleton-Jelly:
When Monika is hugged when it is not raining she says "You can hold me anytime you want."
However, when it is raining and you hug her she says, " You can hug me any time it rains."
I'm not sure what to call this beyond inconsistency, because it's working as intended and writeen correctly, just there was some oversight in the older lines it would appear.

PR #9117, opened 2022-05-10
[Chess Involve] Evergreen problem: Chess Difficulty

Wingdinggaster656:
This PR can be viewed as a sub-version of my ChessButHumanized branch. There were thousands of new lines in that branch, and I suddenly realized that I should split the content I wanted to implement into separate PR's, so I submitted this first.

### Changed:
- Stockfish upgrade. Stockfish's upgrade does not mean an increase in difficulty. The Stockfish update simply means "getting to the target depth faster with the same amount of computing power". There is also a necessary reason to upgrade it here: `UCI_Elo` is only supported with the new version of Stockfish.
- Rewrite difficulty system, difficulty is now measured by ELO score. This will give players a smoother difficulty curve and pave the way for other improvements I want to make to the Chess system in the future.

### Problem:
I haven't been able to get the latest version of Stockfish for MacOS, so it might take someone else to download it at https://stockfishchess.org/download/

### My Test
I played twenty games. They were tested five times each at 200, 800, 1350 and 2500.
In my testing records they did meet the target difficulty expectations.
If you are not sure, you can consider going to chess.com to find some robots of the certain ELO to play chess to confirm.

### So:
Feel free to ask me if you have any questions. I should be here for at least the next week or two.


Wingdinggaster656:
One thing I might want to tell you is how `Skill Level` work.
I'm going to make a not so polite assumption here. Forgive me if I'm wrong:

Based on your previous discussion of chess systems, it seems that you guys know very little about the "UCI Protocol" which is used by chess engines.
`Skill Level` the parameter, well, let me just paste the Stockfish's document here:

>  * #### Skill Level
>    Lower the Skill Level in order to make Stockfish play weaker (see also UCI_LimitStrength).
>   Internally, MultiPV is enabled, and with a certain probability depending on the Skill Level a
>    weaker move will be played.

Do you understand what it means?
That said, the `Skill Level` is essentially giving a possibility -- and since it's just a possibility, it's entirely possible for a player to play a game without triggering a weaker move at all(though this chance is indeed very low).
This is why players often report that "Monika's moves were erratic, sometimes ridiculously strong and sometimes incredibly weak".

The difficulty controlled by `UCI_Elo` -- as far as I can see -- doesn't have this problem.
But what's the problem now? Stockfish only supports the ELO range from 1350 to 2850.

What I did here is write an `if-else` chain to simulate `Skill Level` below 1350 -- but the problem with `Skill Level`? I have mentioned it.
And this, would cause Monika to have an unnatural change of style around 1350 points.

One suggestion I have for this is that when the player first reaches 1350, Monika somehow expresses that her style of play is going to change.
For example, "Seeing your victories these days, I thought you might be interested in learning chess. I will play you more seriously in the future than I did before." Then fixing the lower limit at 1350.

Another idea, of course, would be to ditch the range below 1350 altogether and set the floor at 1350 -- that's not very feasible either. Because even if 1350 is easy enough to reach, but many MAS players don't want to learn it.

I leave the question to you to discuss. If you have any questions about chess engines, feel free to ask me.

multimokia:
FYI that the MacOS and Linux binaries will need `x` (exec) permissions. If you don't have a machine that can apply this permission, I can add it for you

Monika After Story/game/chess.rpy
```diff
@@ -3321,7 +3323,7 @@ init python:
                         store.mas_ptod.wx_cmd("import os", local_ctx)
                         renpy.pause(1.0)
                         store.mas_ptod.wx_cmd(
-                            "subprocess.call(['chmod','+x', os.path.normcase(basedir + '/game/mod_assets/games/chess/stockfish_8_{0}_x64')])".format(
+                            "subprocess.call(['chmod','+x', os.path.normcase(basedir + '/game/mod_assets/games/chess/stockfish_15_{0}_x64')])".format(
                                 "linux" if renpy.linux else "macosx"
```
multimokia, line 3327: I'm not seeing a `stockfish_15_macosx_x64` binary. You're going to need the macos binary for this to work

multimokia: I do have a MacOS VM, I may be able to address your issue to download it

Wingdinggaster656: Your offer of help is just what I need! I needed someone to help me with the MAC version of Stockfish.

As for Stockfish15, which has not been changed to MacOS in the code, it is to remind me that "the Mac problem is not solved."

Now that you say so, I'll get it fixed right away.

Wingdinggaster656: Well, I mean, I can't really get the Mac version of Stockfish. Please help me with this download.
I just thought you meant there was no update in my code to allow MacOS to read Stockfish 15(Then I just noticed that the Linux address and the MacOS address change together)

multimokia: I'm working on it, Mac is just a pain to deal with lol

multimokia:
There's the MacOS binary for stockfish 15

Wingdinggaster656:
> There's the MacOS binary for stockfish 15

Great, now the only problem is to make sure the chess still run on Linux and MacOS.
And, of course, consider about the unnatural change about ELO 1350 that I mentioned above(#9118).

I will do more detailed tests later to make sure Skill Level 6 is about 1350. Somehow, I always feel that Skill Level 6 is a bit high.

multimokia:
Confirming binary works on Linux. That said I did want to address one thing

```
#Stores the chess difficulty.
#The difficulty is actually the ELO rating.
#We set the default value as 850, because ELO 800 is the beginner ELO.
#The chess community generally believes that:
#800 = Beginner
#1200 = Beginner after a few days
#1700 = Advancer
#2000 = Master
#2400 = Grandmaster
#3000 = Probably the human limit
#Here we limit Monika's ELO Rating in 200~2500 the range, and give her a 850 as the default setting.
```

While this might be the *chess community's* belief, I don't think we should be following this as the overwhelming majority of MAS users are simply not chess players and/or playing the minigames as often. Not to mention, likely not with the explicit goal of playing to get better, but rather to play a game with their partner.

As such it may be worth considering setting the bar lower to accommodate for this, but still leaving in the higher bars for those who _do_ play chess more often, so they also experience a decent challenge.

Wingdinggaster656:
Wow, my god, I wrote this on my local computer when testing and forgot to push it up.
`self.stockfish.stdin.write("setoption name MultiPV value 4\n")`
This is the real key reason chess hasn't gotten any easier over the years.

`Skill Level`, as I mentioned above:
>That said, the Skill Level is essentially giving a possibility -- and since it's just a possibility, it's entirely possible for a player to play a game without triggering a weaker move at all(though this chance is indeed very low).

Essentially, `Skill Level` tells Stockfish how bad the choice is in the list of moves Stockfish has generated.
You have never set MultiPV before, so there is only one item in a "list" (MultiPV defaults to 1), there is no randomness at all.

Now let's set it to 4 and the Skill Level effect will be immediately noticeable. I have pushed that change in [ad8d237](https://github.com/Monika-After-Story/MonikaModDev/pull/9117/commits/ad8d237fc737580e9d6dd591c9640cc9449f86b2)

Wingdinggaster656:
> Confirming binary works on Linux. That said I did want to address one thing
> 
> ```
> #Stores the chess difficulty.
> #The difficulty is actually the ELO rating.
> #We set the default value as 850, because ELO 800 is the beginner ELO.
> #The chess community generally believes that:
> #800 = Beginner
> #1200 = Beginner after a few days
> #1700 = Advancer
> #2000 = Master
> #2400 = Grandmaster
> #3000 = Probably the human limit
> #Here we limit Monika's ELO Rating in 200~2500 the range, and give her a 850 as the default setting.
> ```
> 
> While this might be the _chess community's_ belief, I don't think we should be following this as the overwhelming majority of MAS users are simply not chess players and/or playing the minigames as often. Not to mention, likely not with the explicit goal of playing to get better, but rather to play a game with their partner.
> 
> As such it may be worth considering setting the bar lower to accommodate for this, but still leaving in the higher bars for those who _do_ play chess more often, so they also experience a decent challenge.

I totally agree with you on difficulty, but isn't that what this PR is doing?
I mean, the current difficulty range has been set to 200-2500. Even a first-day chess player would get about 400 points, which would give them a good chance to against Monika. You mean what features should be added? I may failed to fully understand your idea.

multimokia:
`#We set the default value as 850, because ELO 800 is the beginner ELO`
This isn't what I'd consider "beginner" by our standards, i.e "ELO 800 - A beginner. Anyone who knows chess rules well would reach this."

While we might touch on the rules, (we don't right now at all), it's likely most MAS users currently do not know the rules period, or if so, potentially a rudamentary understanding of it. As such starting the initial elo below 800 is likely a good idea

Wingdinggaster656:
> `#We set the default value as 850, because ELO 800 is the beginner ELO` This isn't what I'd consider "beginner" by our standards, i.e "ELO 800 - A beginner. Anyone who knows chess rules well would reach this."
> 
> While we might touch on the rules, (we don't right now at all), it's likely most MAS users currently do not know the rules period, or if so, potentially a rudamentary understanding of it. As such starting the initial elo below 800 is likely a good idea

Agree.
I used to set it to a level slightly higher than beginner's, in order to fit Monika's line "I'm pretty good at it".
But now that you say so, I realize that many MAS players are probably around 500 points, so I set it to 600. It makes Monika look tough, but not too hard to reach.

Has been changed in [29223c1](https://github.com/Monika-After-Story/MonikaModDev/pull/9117/commits/29223c14f7031f74362df9c274976c747ef6e482)

Wingdinggaster656:
Just made a change to Stockfish version for Windows in [9bc711d](https://github.com/Monika-After-Story/MonikaModDev/pull/9117/commits/9bc711d450c706015bdd391a20a4cc74157caca1)
The original Stockfish with popcnt instruction set support (originally even AVX2 support) has been changed to Stockfish with no instruction set support.

This slows Stockfish down a bit, but ensures that it can run on almost any computer.

(Seriously, what age is this? Everyone's computer should support popCNT, even AVX2.)

multimokia:
Forgot to mention as well, can confirm this runs on Mac too.

Wingdinggaster656:
Very interesting. Just played another 5 more bullet chess(it means a chess game with a very short time limit (e.g., one minute)).
Here is one of those bullet chess games PGN:

```
1. Nf3 d5 2. c3 Nc6 3. h3 e5 4. Qa4 Bd6 5. e4 dxe4 6. Nh4 Qxh4 7. g3 Qd8 8. g4 Ne7 9. c4 O-O 10. c5 Bxc5 11. Ba6 bxa6 12. Rh2 Ng6 13. Qxe4 Nd4 14. Rh1 Qh4 15. Na3 Ne6 16. d3 Qxf2+ 17. Kd1 Bd4 18. Nb5 Bxb2 19. Bxb2 Qxb2 20. Qd5 Qxa1+ 21. Kc2 Qxh1 22. Qxh1 Bb7 23. Qxb7 Rab8 24. Qxa7 Rxb5 25. a3 Rc5+ 26. Kd2 Ra5 27. g5 Nxg5 28. Ke3 Nxh3 29. Ke2 Rxa3 30. d4 exd4 31. Qxc7 Ra2+ 32. Kd1 d3 33. Qb7 d2 34. Qa7 Ne5 35. Qd4 Nf3 36. Qc5 Ra1+ 37. Ke2 d1=Q+ 38. Ke3 Qd4+ 39. Qxd4 Nxd4 40. Kd3 Nb5 41. Ke4 Re8+ 42. Kd3 Ra4 43. Kd2 Ree4 44. Kd3 Rg4 45. Ke3 Rg3+ 46. Ke2 Ra2+ 47. Kf1 Rg1# 0-1
```

While "this Monika" is 200 ELO, and it does to be a lot eaiser than the lower limit difficulty Monika from before, this is not real 200 ELO. I guess it's about 300~400 ELO.

Not so sure why. Perhaps set `MultiPV` to `4` is still not enough. Going to test more later.

(Man the difficulty system thing is really a evergreen problem)

Wingdinggaster656:
> Yeah the main issue is due to how stockfish works, in that as you mentioned elsewhere, it isn't organic low ELO, but more a higher chance to make a misplay. So basically she's always a grandmaster but will just occasionally throw, which is the main area chess will tend to feel off. Arguably the solution to this is to flat out use something that's not stockfish, and was taught how to play more akin to a neural network that learns how to play games vs being given a minimax rule.
> 
> But this is also significantly more difficult and rather out of the scope (unless one already exists in which case it may not be)

Your idea is actually a misunderstanding. Stockfish with Skill Level 0 has a about 95% error rate, which obviously has nothing to do with grandmasters.

Also, as far as I know, other engines do the same thing. Even the parameters name could be same, since they all follow the UCI protocol.

Wingdinggaster656:
> Yeah the main issue is due to how stockfish works, in that as you mentioned elsewhere, it isn't organic low ELO, but more a higher chance to make a misplay. So basically she's always a grandmaster but will just occasionally throw, which is the main area chess will tend to feel off. Arguably the solution to this is to flat out use something that's not stockfish, and was taught how to play more akin to a neural network that learns how to play games vs being given a minimax rule.
> 
> But this is also significantly more difficult and rather out of the scope (unless one already exists in which case it may not be)

Your idea is actually a misunderstanding. Stockfish with Skill Level 0 has a about 95% error rate, which obviously has nothing to do with grandmasters.
Before this PR, the MAS chess is basically always hard, because since MAS never set `MultiPV` before, `Skill Level` is almost meaningless. Now I set the `MultiPV` to 4, so it's much easier than before.

The only problem I see so far is that the lower limit may not reach 200 points, probably only around 300 points. But that should be enough.

Also, as far as I know, other engines do the same thing. Even the parameters name could be same, since they all follow the UCI protocol.

Wingdinggaster656:
I decided to play 50 bullet chess game with Monika at 400 ELO, probably cost me a few hours.

This sample size should basically confirm if that there are no problems with the current difficulty setting.

Justformas:
> Stockfish with Skill Level 0 has a about 95% error rate, which obviously has nothing to do with grandmasters. Before this PR, the MAS chess is basically always hard, because since MAS never set `MultiPV` before, `Skill Level` is almost meaningless. Now I set the `MultiPV` to 4, so it's much easier than before.
> 
> The only problem I see so far is that the lower limit may not reach 200 points, probably only around 300 points. But that should be enough.

I know this was addressed to Multimokia, but wow my initial guess was actually 90% or 95% error rate, even though Monika seemed to play better in certain circumstances like when the best move is more obvious. I seemed to notice that too, probably due to MultiPV not being higher than 1. 

> Not so sure why. Perhaps set `MultiPV` to `4` is still not enough, or it's just because of my inaccurate evaluation. Going to test more later.

Ok I'll say something about this here now. First, purely out of curiosity, I'm wondering how high MultiPV could actually go, like if it could actually go to 255 or higher. Now even I think that would be ridiculous, though I'm wondering what the limit on that is. Anyway Chess.com goes down to 250 Elo, so if you can get it down to around that  then I think that's good. That might just mean a slight increase in MultiPV, like to 5 or 6, for the lowest level.

Wingdinggaster656:
(I guess I will stop playing chess for today. Playing so many games with Monika has almost burn me out)

Well, good new and bad new.

Good new:
I can say that after my changes, the chess is much easier than before. 
I dare not to say the exact elo number because I haven't carefully tested it. But I guess before my changes, the lower limit of ELO is around 1600.
Now, that lower limit is much lower: The current range, is about 350~2500. And the before, is about 1600~3200.

Bad new:
As what I just said, the lower limit is 350, not 200, still not as what I expected at the beginning.
Though, 350 should be enough, because it's really much easier than before, really. Even if you haven't studied at chess at all, winning Monika is possible now.
(I don't know if Devs are chess pros. If they are not, they can consider to play a few games with Monika. The effect should be obvious)

The real problem could be the performance problem: `MultiPV 8` means 8 lines to calculate when running Stockfish, which may brings CPU problem.
I tested it on a not-so-good computer. The result is that it costs 101ms to finish the Stockfish evaluation. So, basically, as long as the computer is less than 10 years old, there should be no problem. I don't know if the average MAS player's computer is that bad. If so, we may have to consider reducing `MultiPV`, although this may reduce the play experience.

Wingdinggaster656:
> > Stockfish with Skill Level 0 has a about 95% error rate, which obviously has nothing to do with grandmasters. Before this PR, the MAS chess is basically always hard, because since MAS never set `MultiPV` before, `Skill Level` is almost meaningless. Now I set the `MultiPV` to 4, so it's much easier than before.
> > The only problem I see so far is that the lower limit may not reach 200 points, probably only around 300 points. But that should be enough.
> 
> I know this was addressed to Multimokia, but wow some of my initial guesses were actually 90% or 95% error rate, even though Monika seemed to play better in certain circumstances like when the best move is more obvious. I seemed to notice that too, probably due to MultiPV not being higher than 1.
> 
> > Not so sure why. Perhaps set `MultiPV` to `4` is still not enough, or it's just because of my inaccurate evaluation. Going to test more later.
> 
> Ok I'll say something about this here now. First, purely out of curiosity, I'm wondering how high MultiPV could actually go, like if it could actually go to 255 or higher. Now even I think that would be ridiculous, though I'm wondering what the limit on that is. Anyway Chess.com goes down to 250 Elo, so if you can get it down to around that then I think that's good. That might just just mean a slight or moderate increase in MultiPV.
> 
> Edit: I see you increased it to 8.

The max value of `MultiPV` is 500, though it should not be considered. It costs extremely much CPU, which will turns player's computer into a hot fire ball.

And for the lower limit things, the first you should know is the current lower limit is only calculated by my tests, and I might be inaccurate to it. Perhaps it already reached 200, I simply can't see it.
(Because man both 350 and 200 are basically all "blunder-ful" I can hardly see their difference.)

Wingdinggaster656:
Oh, by the way, I'll just post Stockfish's `README.md` here. I think some of you might be interested in it.
The following is Stockfish's `README.md`:


## Overview

[Build Status](https://github.com/official-stockfish/Stockfish/actions)
[Build Status](https://ci.appveyor.com/project/mcostalba/stockfish/branch/master)

[Stockfish](https://stockfishchess.org) is a free, powerful UCI chess engine
derived from Glaurung 2.1. Stockfish is not a complete chess program and requires a
UCI-compatible graphical user interface (GUI) (e.g. XBoard with PolyGlot, Scid,
Cute Chess, eboard, Arena, Sigma Chess, Shredder, Chess Partner or Fritz) in order
to be used comfortably. Read the documentation for your GUI of choice for information
about how to use Stockfish with it.

The Stockfish engine features two evaluation functions for chess. The efficiently 
updatable neural network (NNUE) based evaluation is the default and by far the strongest.
The classical evaluation based on handcrafted terms remains available. The strongest 
network is integrated in the binary and downloaded automatically during the build process.
The NNUE evaluation benefits from the vector intrinsics available on most CPUs (sse2, 
avx2, neon, or similar).

## Files

This distribution of Stockfish consists of the following files:

  * [Readme.md](https://github.com/official-stockfish/Stockfish/blob/master/README.md),
    the file you are currently reading.

  * [Copying.txt](https://github.com/official-stockfish/Stockfish/blob/master/Copying.txt),
    a text file containing the GNU General Public License version 3.

  * [AUTHORS](https://github.com/official-stockfish/Stockfish/blob/master/AUTHORS), 
    a text file with the list of authors for the project

  * [src](https://github.com/official-stockfish/Stockfish/tree/master/src),
    a subdirectory containing the full source code, including a Makefile
    that can be used to compile Stockfish on Unix-like systems.

  * a file with the .nnue extension, storing the neural network for the NNUE
    evaluation. Binary distributions will have this file embedded.

## The UCI protocol and available options

The Universal Chess Interface (UCI) is a standard protocol used to communicate with
a chess engine, and is the recommended way to do so for typical graphical user interfaces
(GUI) or chess tools. Stockfish implements the majority of its options as described
in [the UCI protocol](https://www.shredderchess.com/download/div/uci.zip).

Developers can see the default values for UCI options available in Stockfish by typing
`./stockfish uci` in a terminal, but the majority of users will typically see them and
change them via a chess GUI. This is a list of available UCI options in Stockfish:

  * #### Threads
    The number of CPU threads used for searching a position. For best performance, set
    this equal to the number of CPU cores available.

  * #### Hash
    The size of the hash table in MB. It is recommended to set Hash after setting Threads.

  * #### Clear Hash
    Clear the hash table.

  * #### Ponder
    Let Stockfish ponder its next move while the opponent is thinking.

  * #### MultiPV
    Output the N best lines (principal variations, PVs) when searching.
    Leave at 1 for best performance.

  * #### Use NNUE
    Toggle between the NNUE and classical evaluation functions. If set to "true",
    the network parameters must be available to load from file (see also EvalFile),
    if they are not embedded in the binary.

  * #### EvalFile
    The name of the file of the NNUE evaluation parameters. Depending on the GUI the
    filename might have to include the full path to the folder/directory that contains
    the file. Other locations, such as the directory that contains the binary and the
    working directory, are also searched.

  * #### UCI_AnalyseMode
    An option handled by your GUI.

  * #### UCI_Chess960
    An option handled by your GUI. If true, Stockfish will play Chess960.

  * #### UCI_ShowWDL
    If enabled, show approximate WDL statistics as part of the engine output.
    These WDL numbers model expected game outcomes for a given evaluation and
    game ply for engine self-play at fishtest LTC conditions (60+0.6s per game).

  * #### UCI_LimitStrength
    Enable weaker play aiming for an Elo rating as set by UCI_Elo. This option overrides Skill Level.

  * #### UCI_Elo
    If enabled by UCI_LimitStrength, aim for an engine strength of the given Elo.
    This Elo rating has been calibrated at a time control of 60s+0.6s and anchored to CCRL 40/4.

  * #### Skill Level
    Lower the Skill Level in order to make Stockfish play weaker (see also UCI_LimitStrength).
    Internally, MultiPV is enabled, and with a certain probability depending on the Skill Level a
    weaker move will be played.

  * #### SyzygyPath
    Path to the folders/directories storing the Syzygy tablebase files. Multiple
    directories are to be separated by ";" on Windows and by ":" on Unix-based
    operating systems. Do not use spaces around the ";" or ":".

    Example: `C:\tablebases\wdl345;C:\tablebases\wdl6;D:\tablebases\dtz345;D:\tablebases\dtz6`

    It is recommended to store .rtbw files on an SSD. There is no loss in storing
    the .rtbz files on a regular HDD. It is recommended to verify all md5 checksums
    of the downloaded tablebase files (`md5sum -c checksum.md5`) as corruption will
    lead to engine crashes.

  * #### SyzygyProbeDepth
    Minimum remaining search depth for which a position is probed. Set this option
    to a higher value to probe less aggressively if you experience too much slowdown
    (in terms of nps) due to tablebase probing.

  * #### Syzygy50MoveRule
    Disable to let fifty-move rule draws detected by Syzygy tablebase probes count
    as wins or losses. This is useful for ICCF correspondence games.

  * #### SyzygyProbeLimit
    Limit Syzygy tablebase probing to positions with at most this many pieces left
    (including kings and pawns).

  * #### Move Overhead
    Assume a time delay of x ms due to network and GUI overheads. This is useful to
    avoid losses on time in those cases.

  * #### Slow Mover
    Lower values will make Stockfish take less time in games, higher values will
    make it think longer.

  * #### nodestime
    Tells the engine to use nodes searched instead of wall time to account for
    elapsed time. Useful for engine testing.

  * #### Debug Log File
    Write all communication to and from the engine into a text file.

For developers the following non-standard commands might be of interest, mainly useful for debugging:

  * #### bench *ttSize threads limit fenFile limitType evalType*
    Performs a standard benchmark using various options. The signature of a version 
    (standard node count) is obtained using all defaults. `bench` is currently 
    `bench 16 1 13 default depth mixed`.

  * #### compiler
    Give information about the compiler and environment used for building a binary.

  * #### d
    Display the current position, with ascii art and fen.

  * #### eval
    Return the evaluation of the current position.

  * #### export_net [filename]
    Exports the currently loaded network to a file.
    If the currently loaded network is the embedded network and the filename
    is not specified then the network is saved to the file matching the name
    of the embedded network, as defined in evaluate.h.
    If the currently loaded network is not the embedded network (some net set
    through the UCI setoption) then the filename parameter is required and the
    network is saved into that file.

  * #### flip
    Flips the side to move.


## A note on classical evaluation versus NNUE evaluation

Both approaches assign a value to a position that is used in alpha-beta (PVS) search
to find the best move. The classical evaluation computes this value as a function
of various chess concepts, handcrafted by experts, tested and tuned using fishtest.
The NNUE evaluation computes this value with a neural network based on basic
inputs (e.g. piece positions only). The network is optimized and trained
on the evaluations of millions of positions at moderate search depth.

The NNUE evaluation was first introduced in shogi, and ported to Stockfish afterward.
It can be evaluated efficiently on CPUs, and exploits the fact that only parts
of the neural network need to be updated after a typical chess move.
[The nodchip repository](https://github.com/nodchip/Stockfish) provided the first 
version of the needed tools to train and develop the NNUE networks. Today, more 
advanced training tools are available in 
[the nnue-pytorch repository](https://github.com/glinscott/nnue-pytorch/), 
while data generation tools are available in 
[a dedicated branch](https://github.com/official-stockfish/Stockfish/tree/tools).

On CPUs supporting modern vector instructions (avx2 and similar), the NNUE evaluation
results in much stronger playing strength, even if the nodes per second computed by
the engine is somewhat lower (roughly 80% of nps is typical).

Notes:

1) the NNUE evaluation depends on the Stockfish binary and the network parameter file 
(see the EvalFile UCI option). Not every parameter file is compatible with a given
Stockfish binary, but the default value of the EvalFile UCI option is the name of a 
network that is guaranteed to be compatible with that binary.

2) to use the NNUE evaluation, the additional data file with neural network parameters
needs to be available. Normally, this file is already embedded in the binary or it can
be downloaded. The filename for the default (recommended) net can be found as the default
value of the `EvalFile` UCI option, with the format `nn-[SHA256 first 12 digits].nnue`
(for instance, `nn-c157e0a5755b.nnue`). This file can be downloaded from
```
https://tests.stockfishchess.org/api/nn/[filename]
```
replacing `[filename]` as needed.

## What to expect from the Syzygy tablebases?

If the engine is searching a position that is not in the tablebases (e.g.
a position with 8 pieces), it will access the tablebases during the search.
If the engine reports a very large score (typically 153.xx), this means
it has found a winning line into a tablebase position.

If the engine is given a position to search that is in the tablebases, it
will use the tablebases at the beginning of the search to preselect all
good moves, i.e. all moves that preserve the win or preserve the draw while
taking into account the 50-move rule.
It will then perform a search only on those moves. **The engine will not move
immediately**, unless there is only a single good move. **The engine likely
will not report a mate score, even if the position is known to be won.**

It is therefore clear that this behaviour is not identical to what one might
be used to with Nalimov tablebases. There are technical reasons for this
difference, the main technical reason being that Nalimov tablebases use the
DTM metric (distance-to-mate), while the Syzygy tablebases use a variation of the
DTZ metric (distance-to-zero, zero meaning any move that resets the 50-move
counter). This special metric is one of the reasons that the Syzygy tablebases are
more compact than Nalimov tablebases, while still storing all information
needed for optimal play and in addition being able to take into account
the 50-move rule.

## Large Pages

Stockfish supports large pages on Linux and Windows. Large pages make
the hash access more efficient, improving the engine speed, especially
on large hash sizes. Typical increases are 5..10% in terms of nodes per
second, but speed increases up to 30% have been measured. The support is
automatic. Stockfish attempts to use large pages when available and
will fall back to regular memory allocation when this is not the case.

### Support on Linux

Large page support on Linux is obtained by the Linux kernel
transparent huge pages functionality. Typically, transparent huge pages
are already enabled, and no configuration is needed.

### Support on Windows

The use of large pages requires "Lock Pages in Memory" privilege. See
[Enable the Lock Pages in Memory Option (Windows)](https://docs.microsoft.com/en-us/sql/database-engine/configure-windows/enable-the-lock-pages-in-memory-option-windows)
on how to enable this privilege, then run [RAMMap](https://docs.microsoft.com/en-us/sysinternals/downloads/rammap)
to double-check that large pages are used. We suggest that you reboot
your computer after you have enabled large pages, because long Windows
sessions suffer from memory fragmentation, which may prevent Stockfish
from getting large pages: a fresh session is better in this regard.

## Compiling Stockfish yourself from the sources

Stockfish has support for 32 or 64-bit CPUs, certain hardware
instructions, big-endian machines such as Power PC, and other platforms.

On Unix-like systems, it should be easy to compile Stockfish
directly from the source code with the included Makefile in the folder
`src`. In general it is recommended to run `make help` to see a list of make
targets with corresponding descriptions.

```
    cd src
    make help
    make net
    make build ARCH=x86-64-modern
```

When not using the Makefile to compile (for instance, with Microsoft MSVC) you
need to manually set/unset some switches in the compiler command line; see
file *types.h* for a quick reference.

When reporting an issue or a bug, please tell us which Stockfish version
and which compiler you used to create your executable. This information
can be found by typing the following command in a console:

```
    ./stockfish compiler
```

## Understanding the code base and participating in the project

Stockfish's improvement over the last decade has been a great community
effort. There are a few ways to help contribute to its growth.

### Donating hardware

Improving Stockfish requires a massive amount of testing. You can donate
your hardware resources by installing the [Fishtest Worker](https://github.com/glinscott/fishtest/wiki/Running-the-worker:-overview)
and view the current tests on [Fishtest](https://tests.stockfishchess.org/tests).

### Improving the code

If you want to help improve the code, there are several valuable resources:

* [In this wiki,](https://www.chessprogramming.org) many techniques used in
Stockfish are explained with a lot of background information.

* [The section on Stockfish](https://www.chessprogramming.org/Stockfish)
describes many features and techniques used by Stockfish. However, it is
generic rather than being focused on Stockfish's precise implementation.
Nevertheless, a helpful resource.

* The latest source can always be found on [GitHub](https://github.com/official-stockfish/Stockfish).
Discussions about Stockfish take place these days mainly in the [FishCooking](https://groups.google.com/forum/#!forum/fishcooking)
group and on the [Stockfish Discord channel](https://discord.gg/nv8gDtt).
The engine testing is done on [Fishtest](https://tests.stockfishchess.org/tests).
If you want to help improve Stockfish, please read this [guideline](https://github.com/glinscott/fishtest/wiki/Creating-my-first-test)
first, where the basics of Stockfish development are explained.


## Terms of use

Stockfish is free, and distributed under the **GNU General Public License version 3**
(GPL v3). Essentially, this means you are free to do almost exactly
what you want with the program, including distributing it among your
friends, making it available for download from your website, selling
it (either by itself or as part of some bigger software package), or
using it as the starting point for a software project of your own.

The only real limitation is that whenever you distribute Stockfish in
some way, you MUST always include the license and the full source code
(or a pointer to where the source code can be found) to generate the 
exact binary you are distributing. If you make any changes to the
source code, these changes must also be made available under the GPL v3.

For full details, read the copy of the GPL v3 found in the file named
[*Copying.txt*](https://github.com/official-stockfish/Stockfish/blob/master/Copying.txt).

Justformas:
> The max value of `MultiPV` is 500, though it should not be considered. It costs extremely much CPU, which will turns player's computer into a hot fire ball.

Don't worry, I was not seriously considering that as a good idea. And actually the thought of CPU processing issues did occur to me around when I thought of 255. though I wasn't sure about that.

> The real problem could be the performance problem: `MultiPV 8` means 8 lines to calculate when running Stockfish, which may brings CPU problem.
> I tested it on a not-so-good computer. The result is that it costs 101ms to finish the Stockfish evaluation. So, basically, as long as the computer is less than 10 years old, there should be no problem. I don't know if the average MAS player's computer is that bad. If so, we may have to consider reducing `MultiPV`, although this may reduce the play experience.

So even MultiPV 8 may cause issues? Man I didn't realize that. Was the computer you tested it on about 10 year old then? Because 101ms doesn't sound bad at all (that's only about 1/10 of a second right?) and Monika says something while she's thinking anyway. Even if it takes half a second or a whole second on computers older than that, that doesn't seem bad. Unless maybe the player is doing other things that take a lot of CPU power on that older computer of theirs at the same time, but they can avoid doing that if necessary.

> And for the lower limit things, the first you should know is the current lower limit is only calculated by my tests, and I might be inaccurate to it. Perhaps it already reached 200, I simply can't see it.
> (Because man both 350 and 200 are basically all "blunder-ful" I can hardly see their difference.)

It might be fine to just go with 350 then (and still have skill level 6 be 1350). I can tell you from playing on chess.com that I do seem to feel a difference between 250 and 400, but 400 is a little higher than 350.

Also, I might be willing to test this myself, but I'm not sure how to go about doing that. Because it was my understanding that you need not only Stockfish, but an interface to play it on (to play chess conveniently/traditionally anyway). But if it's relatively simple and there's a tutorial somewhere someone can link to on how to do that, then I might try it.

Wingdinggaster656:
> 

Alright, things are like this:

Test is actually pretty simple, you don't have to use a GUI(the `an interface`).
Simply downloading the Stockfish binary file, then type these:
```
uci
setoption name Skill Level value X // X is the value that you want to use
setoption name MultiPV value 8
```
Each line should be ended with a `Enter` button press on your keyboard.

Then go to https://www.chess.com/play/computer and find a bot suits.
In Stockfish, type `position fen <FenStringHere>` to update positions, then type `go depth 8`, after about 0.01s the result would appears. Play that move to the bot you chose.

example
download fen
fen there

Justformas:
Ok I think I figured (most of) it out, but I don't think I'm doing this efficiently (either that, or I don't like this process). I set skill level to 0 and MultiPV to 8, and I use go depth 8. Getting through even 1 game with what I'm doing is going to take me awhile.

What I've been doing is every time a move is made: 

-Click on the download button again to see the new fen string.
-Click to highlight and copy/press Control + C.
-Go back to Stockfish and type: position fen <Control + V>
-Type: go depth 8
-Input the move it gives me into chess.com
-Repeat this whole process after every move

Is there a better way? Also the "bestmove" Stockfish gives me to use is often the best move, but it has given me non-best moves too. By the way, is the "ponder" move what it thinks the opponent will do next or something?

Wingdinggaster656:
> 

You are right about "ponder".

And better way? Sadly I don't think there is a better way. If you know codes, you could consider to set Stockfish in a GUI, that would save some time, but you probably don't.

And, just ignore `bestmove` the word. You can simply think it means `result`.

Justformas:
> You are right about "ponder".
> 
> And better way? Sadly I don't think there is a better way. If you know codes, you could consider to set Stockfish in a GUI, that would save some time, but you probably don't.
> 
> And, just ignore `bestmove` the word. You can simply think it means `result`.

So then you're saying there's no relatively simple way for someone that doesn't code to set up a GUI with Stockfish. However another thing I was referring when I said testing was your project as it works with MAS. You said:

> (I don't know if Devs are chess pros. If they are not, they can consider to play a few games with Monika. The effect should be obvious)

Could I do that too? Like is it in the "unstable branch" update in the game? But the unstable branch... I would have to back up not only my persistent files but also backup or make a whole new copy of MAS for it, wouldn't I?

Or are you referring to something different here?

Wingdinggaster656:
> > You are right about "ponder".
> > And better way? Sadly I don't think there is a better way. If you know codes, you could consider to set Stockfish in a GUI, that would save some time, but you probably don't.
> > And, just ignore `bestmove` the word. You can simply think it means `result`.
> 
> So then you're saying there's no relatively simple way for someone that doesn't code to set up a GUI with Stockfish. However another thing I was referring to when I said testing was your project as it works with MAS. It may not let me measure the Elo accurately, but I could get a feel for it. You said:
> 
> > (I don't know if Devs are chess pros. If they are not, they can consider to play a few games with Monika. The effect should be obvious)
> 
> Could I do that too? Like is it in the "unstable branch" update in the game? But the unstable branch... I would have to back up not only my persistent files but also backup or make a whole new copy of MAS for it, wouldn't I? Unless I can "update" back to the stable version.
> 
> Or are you referring to something different here?

If you want to test like that, a easy way is:
1. Backup all your persistent files.
2. Download all files in this PR, change the `/game` folder: Paste `chess.rpy` in this PR into the file directory, and update Stockfish version.
3. Okay, restart the game, it should be done now. You can play chess.
4. Restore your persistent files after you finished.

The problem is that you probably don't know how to set ELO points, and how to get the current ELO.
Did you read wiki at all?

Justformas:
> By the way I personally think we are not supposed to talk so much here(though this is not banned). Because our conversation is not-that-on-topic, and I think PR should be highly condensed.

I have more to say on what we've been talking about, so then where should we talk about it? In your discussion you made or a new discussion? Because not being able to have a conversation that can flow in whatever direction it needs to flow can be very frustrating and serve as an obstacle for breakthroughs. And so can trying to find the right place on a forum where that's allowed/encouraged if there's such a place at all.

> Did you read the wiki of MAS at all? If not I suggest a thorough reading. It could explain some coding base.

I have read most of the FAQ and feature guide, thought I read the rules and code of conduct but might not remember (or understand) everything about that. I have tried looking through some of the coding guides before a little but I didn't really understand them, but I also wasn't that interested in them at the time. Besides to me it's all a lot to read through and understand, especially when I also consider all the topics in the game to read (and sometimes reread) too.

Wingdinggaster656:
> I have read (most if not all of) the FAQ and feature guide, thought I read the rules and code of conduct but might not remember (or understand) everything about that. I have tried looking through some of the coding guides before a little but I didn't really understand them, but I also wasn't that interested in them at the time. Besides to me it's all a lot to read through and understand, especially when I also consider all the topics in the game to read (and sometimes reread) too.

Continue our conversation here is okay, I am just trying to let you know that PR's conversation should be condensed.

As for the coding problem....... If you don't really understand codes, you'll probably have to test it in the not-so-beautiful way, which is to keep typing `position fen` things.
But if you ask me, I think I'm pretty much done with this test. You might want to save your time and do something else.

Justformas:
> As for the coding problem....... If you don't really understand codes, you'll probably have to test it in the not-so-beautiful way, which is to keep typing `position fen` things. But if you ask me, I think I'm pretty much done with this test. You might want to save your time and do something else.

I was still playing and recently finished that "game". It was against Elo 400 on chess.com. It took 56 turns, but eventually got lopsided and I ended up winning with Stockfish's moves. Twice towards the end I put in depth 6 by mistake, and at one point when it wanted me to promote a pawn, it said "g7f6n" and I wasn't sure if "n" meant knight or queen. I just picked queen, but this queen got captured next turn anyway. So I think Stockfish probably would have ended up winning anyway. It could just be the way randomness turned out this time. Maybe it can lose against 400 sometimes. It did make much worse moves than the best one sometimes, so maybe it's fine for the lowest level.

One day I might put more effort into understanding code, or it may sort of just come to me anyway. I've already gotten much better at understanding it than when I started playing this game. Give it time and we'll see.

I may still try your project in the game too at some point, since it may be some time before its finished, approved and officially put into the game.

Wingdinggaster656:
> 

If you next time finish a game, send the PGN here.
I have the ability to roughly analyze the ELO from a game.
Based on the fact that it took 56 turns to complete the game, I can conclude that the ELO score is around 400 (which confirms my previous results with more rounds of testing: 350).

Justformas:
> If you next time finish a game, send the PGN here.

Ah I thought about saving it, but not enough to I guess. Also I haven't had experience with PNG files before and didn't think about being able to upload it here anyway.

> I have the ability to roughly analyze the ELO from a game.
> Based on the fact that it took 56 turns to complete the game, I can conclude that the ELO score is around 400 (which confirms my previous results with more rounds of testing: 350).

You mean, you're able to judge just by looking at them or by how long they last? I guess I shouldn't be that surprised, though with all the testing you've been doing I thought maybe you were going by something like win rate.

So you're saying the Elo of Stockfish in that game was around 400, but testing it with enough rounds, it averages out to 350? I can see that being the case. I could maybe just play against 400 to get more of a feel for what it'd be like in the mod then.

Wingdinggaster656:
The most authoritative, the most effective, and the oldest method to give ELO rating is indeed the win rate. I've done a lot of testing on my side, so I can use the win rate, but if you're playing a single game you have to do something else.

Judging by the number of turns is, of course, not a accurate method. But since you didn't give me PGN, that's all I can do. If you can give me PGN, I can analyze how both sides play those moves to get a ELO rating.
And this is not a great skill, too. I guess anyone who has ELO 2000 should have this ability. Actually "guess ELO" is a popular activity in chess community,like this one which is just created four days ago: https://www.chess.com/forum/view/general/my-guess-the-elo

Justformas:
I think even before your "edit twice" I could see why an experienced/serious chess player could take a good guess at ELO in a match. So actually now, "I want to play a game". I don't remember if you told me specifically, but based on everything I remember you telling me, I want to see if I can guess a ballpark figure on your ELO. I think you aren't lower than 2100 or not by much, but I think you could even be around 2500. I would be a little surprised if you are actually much higher than that, but not that much. So are you somewhere in the range of 2100 to 2500?


Oh and I had a question about something else:

>I decided to play 50 bullet chess game with Monika at 400 ELO, probably cost me a few hours.
I think you mean here that you played 50 games while more or less intentionally limiting your time to within 3 minutes per game, right? But are you actually planning on adding time limit modes? Even if you do, then there should still be the unlimited time mode for the player too.

Booplicate:
Please, keep PR comments on-topic, having to scroll all the way down makes reviewing harder. Only post something if it's directly related to the PR - for example, you have a better idea how to do something or found a bug.

If you need some explanation on a semi-related topic, open a discussion (you/we can link it to the PR if needed).

kkrosie123:
To add to Booplicate's message, you can find the discussion page for this PR at #9118 ! 

Have a nice day!

Wingdinggaster656:
It just hit me that, considering about the old codes, Devs probably know really limited things about Stockfish.
Therefore, it's necessary for me to explain something about my setting about Stockfish's search depth.

Well, actually the depth do really little little.
Sometimes a chess engine with a depth of 1 will yield the same result as a depth of 99. This is pretty possible.
The most common result is: assuming Stockfish is white, the situation evaluation score is +0.98 after a depth 1 move. And if that was a depth 99 move, then it's +1.00 . Do you understand? Depth matters so little that I suggest you ignore it altogether.

The computing power demanded is exponentially related to depth. Depth 99 requires a trillion times more computing power than depth 20, but is probably only 1% - 2% more accurate.

Conclusion: Depth really changes chess difficulty very little, so in this version I wrote, I didn't touch it at all. I just fixed the depth at 8.

I make this clear in advance, in case I'm not here when you ask about it and you have to wait for me.

Wingdinggaster656:
> Wow, my god, I wrote this on my local computer when testing and forgot to push it up. `self.stockfish.stdin.write("setoption name MultiPV value 4\n")` This is the real key reason chess hasn't gotten any easier over the years.
> 
> `Skill Level`, as I mentioned above:
> 
> > That said, the Skill Level is essentially giving a possibility -- and since it's just a possibility, it's entirely possible for a player to play a game without triggering a weaker move at all(though this chance is indeed very low).
> 
> Essentially, `Skill Level` tells Stockfish how bad the choice is in the list of moves Stockfish has generated. You have never set MultiPV before, so there is only one item in a "list" (MultiPV defaults to 1), there is no randomness at all.
> 
> Now let's set it to 4 and the Skill Level effect will be immediately noticeable. I have pushed that change in [ad8d237](https://github.com/Monika-After-Story/MonikaModDev/pull/9117/commits/ad8d237fc737580e9d6dd591c9640cc9449f86b2)

Actually, lately, I found that even if MultiPV is the default 1, there is still bad moves generated, though this frequency is indeed very low. After reading some Stockfish source codes, it seems like that MultiPV 1 is generating 2 lines, and MultiPV 2 is generating 3 lines. Not 100% sure about this, but anyways, a higher MultiPV lowers the final performance result at an obvious amplitude. Also, Stockfish Devs emphasized more than once "Set MultiPV to 4 or more to really enable Skill Level". This is what I can make sure.

You can find one of "Stockfish Devs emphasized more than once" here: https://github.com/official-stockfish/Stockfish/issues/2669

> To enable skill level we need to play MultiPV 4, which is a significant overhead, at STC, roughly 500 Elo. That's also why UCI_Elo is capped at 2850 Elo, that's roughly skill level 19.

Also I ran more games with Monika. I can confirm it's easier than it used to be.
I won't be testing the PR any more, but if you have any questions about Stockfish and this PR, feel free to ask me (you can also go to Stockfish's Github page, although most of the content there is about how to improve Stockfish, and I'm worried you won't get any answers).

Justformas:
Just wanted to let anyone reading this PR know that Wingdinggaster656 was not talking to the air, but was talking to me. I deleted my comments to reduce having to scroll down.

Wingdinggaster656:
Note: Well I eventually came back later to test this PR again.
I have observed something strange: Monika seems to have a difficulty shift by mistake when the ELO score is extremely low (below 400), resulting in Monika's actual strength being close to 2000.

I don't have time to look into the cause for now. I'll come back in a week to check this.

Wingdinggaster656:
**Warning: Do not test this PR, and do not merge it for now.**

After doing a lot of research, I am now seriously considering switching the engine to Komodo 13.
Komodo engine also follows the UCI protocol, so there's not much code effort involved in switching engines.
(And don't worry, even if there are any code issues, it's my duty to finish them, since it seems like I know how to interact with chess engines much better than you guys).

Stockfish can achieve a low ELO. This is no problem.
The problem is, simply solving the chess difficulty problem is not the `ChessButHumanized` project. I am working hard to make Monika look like a real person from any direction. I would like Monika to have her play style, and a lot of details that I won't list here until we review them in the future. And Stockfish was a little less successful at achieving these coming-up details.

Komodo 13 is free, and from what I've read so far, it's officially allowed to be distributed for free.
The problem is Komodo 13, indeed, does run much slower than Stockfish 15(just like what I expected), and I'm not sure Komodo 13 will run on most computers.

I'm going to do a lot of research and evaluation to make a final decision on what engine we're going to use.
Until then, temporarily stop looking at this PR.

My investigation won't take longer than a month. I will output the final product before 2022/07/01 for you to final review.

Wingdinggaster656:
**Warning: Do not test this PR, and do not merge it for now.**

After doing a lot of research, I am now seriously considering switching the engine to Komodo 13.
Komodo engine also follows the UCI protocol, so there's not much code effort involved in switching engines.
(And don't worry, even if there are any code issues, it's my duty to finish them, since it seems like I know how to interact with chess engines much better than you guys).

Stockfish can achieve a low ELO. This is no problem.
The problem is, simply solving the chess difficulty problem is not my `ChessButHumanized` project.
I am working hard to make Monika look like a real person from any direction. I would like Monika to have her play style, and a lot of details that I won't list here until we review them in the future. And Stockfish was a little less successful at achieving these coming-up details.

Komodo 13 is free, and from what I've read so far, it's officially allowed to be distributed for free.
Though, Komodo 13, indeed, just like what I expected, does run much slower than Stockfish 15, and I'm not sure Komodo 13 will run on most computers(This is all the more worrying given that some of MAS's users seem unable even to run Stockfish).

I'm going to do a lot of research and evaluation to make a final decision on what engine we're going to use.
Until then, temporarily stop looking at this PR.

My investigation won't take longer than a month. I will output the final product before 2022/07/01 for you to final review.

ThePotatoGuy:
Some notes for whenever you finish your investigation:
* highly unlikely we're going to switch from stockfish. I also am hesitant to upgrade the version unless we absolutely have to.
* Since it seems you have a scale for skill level and ELO, we 100% should add a chess option for telling Monika your ELO rating so she can adjust her skill. A seasoned chess player probably doesn't want to play through garbage chess to get to an equal playing field.

Wingdinggaster656:
@ThePotatoGuy Is there a reason we don't want to change the engine? Are you worried about copyright?
Don't worry, I will find a perfectly legal engine and I will give you all the references for you so you can be 100% sure.

Let me also talk a little bit about my current research. I plan to use two engines in MAS. When we are under 1350, we use `Belofte chess`, and when we are not under 1350, we use Stockfish.
Belofte is just designed for beginner, it makes for a very pleasant experience for beginners, but the downside is that it's designed purely for beginners. It doesn't even have the ability to provide high ELO.
Belofte's file size is quite small. So I think that's fine.
(Also just to make you relax: Belofte is open-sourced and free to distribute, since it follows the `GPL v2`.)

As for telling Monika your ELO -- I'm 100% for that.
I planned that when Monika first mentioned chess to the player, she would ask what level the player was and give her initial ELO value based on that.
I also plan to output Monika's ELO in our PGN files. Standard PGN files contain ELO output.

Booplicate:
I'd say if we move to a new engine it:
- the license should be as "free" (as in freedom) as possible
- should be maximally easy to use, akin to stockfish, plug and play
- should support our 3 main targeted OS - linux, mac, windows
- shouldn't have dependencies, or should ship them (and they should be small)

As much as I'd want to update stockfish, your updated version is ~180 MiB, I'm not even sure if we can afford that, it's barely less than the current mod size.

Then you're talking about shipping *2* engines. First, it just doesn't look right to me, it's overkill. I understand the idea, we all want have better chess, but MAS isn't about chess, we can't ship GiB's of chess libs. Then we would have to figure out how to unify it all under one system, test on different platforms. Being realistic, I don't see that coming, but I might be wrong, too.

Tbf I also don't see how any chess engine would make her play more like a human, it'd need to be some neural network which would be able to dynamically adapt, all (mostly?) other engines are made to analyse for the best move.

My 2 cents are: Chess is cool, many people enjoy it, and it's a considerable part of MAS, but not a huge part of MAS.. Work with what we already have, *maybe* (very big maybe) update stockfish. Better would be playing around with available input parameters, new chess game modes, new other games.

Wingdinggaster656:
> I'd say if we move to a new engine it:
> 
> * the license should be as "free" (as in freedom) as possible
> * should be maximally easy to use, akin to stockfish, plug and play
> * should support our 3 main targeted OS - linux, mac, windows
> * shouldn't have dependencies, or should ship them (and they should be small)
> 
> As much as I'd want to update stockfish, your updated version is ~180 MiB, I'm not even sure if we can afford that, it's barely less than the current mod size.
> 
> Then you're talking about shipping _2_ engines. First, it just doesn't look right to me, it's overkill. I understand the idea, we all want have better chess, but MAS isn't about chess, we can't ship GiB's of chess libs. Then we would have to figure out how to unify it all under one system, test on different platforms. Being realistic, I don't see that coming, but I might be wrong, too.
> 
> Tbf I also don't see how any chess engine would make her play more like a human, it'd need to be some neural network which would be able to dynamically adapt, all (mostly?) other engines are made to analyse for the best move.
> 
> My 2 cents are: Chess is cool, many people enjoy it, and it's a considerable part of MAS, but not a huge part of MAS.. Work with what we already have, _maybe_ (very big maybe) update stockfish. Better would be playing around with available input parameters, new chess game modes, new other games.

Actually the reason why new Stockfish is so big is that it has a NNUE library.
This provides faster computing speed, but only optional. If you don't like it, I can remove the NNUE and Stockfish will go back to its previous size.

Just to let you know, the new second engine, combined with Windows, Mac, and Linux, is just around 4 MB. **I already mentioned this above.**

In addition, all of your requirements mentioned above are these:
>the license should be as "free" (as in freedom) as possible
>should be maximally easy to use, akin to stockfish, plug and play
>should support our 3 main targeted OS - linux, mac, windows
>shouldn't have dependencies, or should ship them (and they should be small)

In fact, except "free", this is the common denominator of almost all UCI engines.
Just so you know, UCI is the most popular chess engine protocol, so popular that I don't even know of a single engine that doesn't use it.
As for the "free" problem, I know how to pick and choose. Does it look like I'm going to upload an engine here that's not allowed to be distributed for free?

Conclusion:
If you think new Stockfish is too big, I can delete the NNUE library, which will save a lot of disk space. As for the second engine I'm going to add, it's also really small.
> Belofte's file size is quite small. So I think that's fine.
I personally hope you do some research before commenting, or at least read what I said.

Wingdinggaster656:
As a footnote, since almost all chess engines follow the UCI protocol:

- They all accept positions with `position fen <fen string here>`
- They all calculate moves with `go`
- They all set options with `setoption name <option name here> value <option value here>`
- They all output the move with the same format
- Many engines even have the same option names.

As the result, I can't really figure out how many code changes we need.
And even if there were, wouldn't it be my responsibility to change the code?

As for file sizes, if you think new Stockfish is too big, I'll delete the NNUE library for you.
As for the new engine I'm adding, it's also quite small, with a combined file size of around 4MB for all three platforms.

Is this clear enough?

Booplicate:
> I personally hope you do some research before commenting, or at least read what I said.

You said small, but you have stockfish executables that are 180 MiB, I don't know what's small for you. 4 MiB looks good, but I'm not settled on whether or not we should change the engine.

> Does it look like I'm going to upload an engine here that's not allowed to be distributed for free?

I can't read your mind, I just showed my concerns. Nice of you to verify all that though.

> Actually the reason why new Stockfish is so big is that it has a NNUE library.
This provides faster computing speed, but only optional. If you don't like it, I can remove the NNUE and Stockfish will go back to its previous size.

Good to know, we probably should do that. Looks like NNUE is under the same license as stockfish, but I'm not sure how well it'd run, also that size is too big.

Komodo, which you suggested before, seems to be proprietary, some versions are even paid. So that's definitely a no.
I didn't read about Belofte yet (seem to be much less popular than other options), will look later.

Still don't think we need 2 chess engines for a minigame. Chess isn't the point of MAS, it's not a game about chess and if anything we should focus on casual audience.
When I played it, I felt increasing in difficulty, but people found that it's not good enough. You decided to update the engine and change the parameters - you reported it worked. Then let's try that route first before switching to another engine or implementing a system to use multiple engines (even if they're using 1 interface).

There's currently 2 problems I know about:
- she randomly throws pieces which feels dum; is the new engine going to fix that?
- stockfish has stopped working for some people, I was thinking that maybe a newer version of stockfish would fix it, but we have no data on why this happened, nor do we know if the update would help or break things.

Wingdinggaster656:
> she randomly throws pieces which feels dum; is the new engine going to fix that?

It will be better than before.
Now she tends to make "positional mistakes", such as moving pawns at the wrong time and creating her potential vulnerabilities.
But, well, don't expect "throw pieces" to be completely absent.

> stockfish has stopped working for some people, I was thinking that maybe a newer version of stockfish would fix it, but we have no data on why this happened, nor do we know if the update would help or break things.

In fact, I was extremely confused too.
But Stockfish 15 even runs on Windows Vista, so I don't think it's a problem.
Just what I think.

> Komodo, which you suggested before, seems to be proprietary, some versions are even paid. So that's definitely a no.

And Komodo did get vetoed by me. I thought you understood what I meant.
Here's an interesting thing: even the free version of Komodo is not allowed to be distributed. I'm not sure why, but it's off my list.
Please don't worry about copyright. After I find our engine, I'll post its license address to you so you can check for yourself if it's free.

> Good to know, we probably should do that. Looks like NNUE is under the same license as stockfish, but I'm not sure how well it'd run, also that size is too big.

That's my job. I will upload Stockfish with NNUE library deleted later.

> Then let's try that route first before switching to another engine or implementing a system to use multiple engines (even if they're using 1 interface)

Yes we can, though multiple engines should not hurt. At this point, give me some time to see how necessary it is to use a new engine. If the play experience is very different, let's add an engine.


Ask me if you have any questions about chess engines. I'd be happy to help you with your questions.

Booplicate:
> Now she tends to make "positional mistakes", such as moving pawns at the wrong time and creating her potential vulnerabilities.
But, well, don't expect "throw pieces" to be completely absent.

Well technically she's done that before too, so we'll need to compare if it's any better.

> Yes we can, though multiple engines should not hurt

It might, I gave you a few reasons. It is overkill and questionable design choice, you don't just put in a few backends that do the same task. It'd make the codebase more complex, but most users wouldn't (?) even notice the difference. I'm also very skeptical about any changes to the executables - I don't want to distribute crypto miners and <"peace" not war>s which became a popular trend in OSS.

Wingdinggaster656:
> Well technically she's done that before too, so we'll need to compare if it's any better.

Of course, that's 100% understandable and necessary.

> It'd make the codebase more complex, but most users wouldn't (?) even notice the difference.

I could barely see how more complex it will be. The final code would be like:
```
if elo<1350:
    open_new_engine()
else:
    open_stockfish()
```
And the difference is very stark. I'll leave that for you to experiment with, because I don't think you're going to believe me even if I say it's stark.

> I'm also very skeptical about any changes to the executables - I don't want to distribute crypto miners and <"peace" not war>s which became a popular trend in OSS.

Well, I respect your skepticism. That is professional.
Stockfish follows the GPL License, which means that if I use a modified version of it, I must present the modified source code intact. You will of course also get the modified version of the source code.
If you think I'm adding something to the exe, you can compile that code yourself.

You don't suspect now that I have installed a virus in your computer right?


Anyways, any other doubts? I'd be happy to help you with your questions.

Wingdinggaster656:
> Well technically she's done that before too, so we'll need to compare if it's any better.

Of course, that's 100% understandable and necessary.

> It'd make the codebase more complex

I could barely see how more complex it will be. The final code would be like:
```
if elo<1350:
    open_new_engine()
else:
    open_stockfish()
```

> but most users wouldn't (?) even notice the difference.

And the difference is very stark. I'll leave that for you to experiment with, because I don't think you're going to believe me even if I say it's stark.

> I'm also very skeptical about any changes to the executables - I don't want to distribute crypto miners and <"peace" not war>s which became a popular trend in OSS.

Well, I respect your skepticism. That is professional.
Stockfish follows the GPL License, which means that if I use a modified version of it, I must present the modified source code intact. You will of course also get the modified version of the source code.
If you think I'm adding something to the exe, you can compile that code yourself.

You don't suspect now that I have installed a virus in your computer right?


Anyways, any other doubts? I'd be happy to help you with your questions.

Wingdinggaster656:
@Booplicate @ThePotatoGuy 
So I eventually finished my investigation.

My final result: we only need one chess engine. `Fairy-Stockfish`.
From its name, you can guess where it came from: yes, it's the Stockfish we use now.
Under GPL V3, a group of people adapted Stockfish's source code to make a version that was "more human-friendly than just focused on beating other computer engines."

It allows us to choose the ELO range from 500 to 2850, which I think is really quite adequate. Really. Even a person who just know how to move pieces, they still roughly have 500 ELO.

Fairy-Stockfish also supports many chess variants. While I don't have any plans to implement these new game modes at the moment, it will save a lot of effort if someone else wants to do it later.

File size?
Please don't worry, although the latest version uses NNUE algorithm, I can remove NNUE algorithm.
However, given that if we were to make changes to the source code, we would have to attach the adapted source code to MAS, I don't think we should do that -- we'll just go with the second new version. That's about 7 MB, which is only 1-2 MB more than Stockfish today, and I think that is perfectly acceptable.

Copyright problem?
As I said, it's an adaptation of Stockfish, also under the GPL. So it's 100% free, 100% to distribute.

See here: https://github.com/ianfab/Fairy-Stockfish

If you have any questions about this, please feel free to ask me. And if there were no problems, then I'm going to upload it here.

Booplicate:
Sorry, but what's the point of changing the engine? I can't find anything about "being more human-like."

> The goal of the project is to create an engine supporting a large variety of chess-like games, equipped with the powerful search of Stockfish. Despite its generality the [playing strength](https://github.com/ianfab/Fairy-Stockfish/wiki/Playing-strength) is on a very high level in almost all supported variants.

Says that it's strong and supports different game modes.

Then there's this:
> Fairy-Stockfish outperforms most chess variant engines, except for variants where many highly specialized engines exist (e.g., Shogi), and is the strongest engine in a large variety of variants such as Janggi, Sittuyin, Capablanca chess, etc. For some variants NNUE networks are available that can further improve playing strength
When using NNUE evaluation Fairy-Stockfish plays at a superhuman level in all major regional chess variants, such as Xiangqi, Shogi, Janggi, and Makruk. Even in [user-defined variants](https://github.com/ianfab/Fairy-Stockfish/wiki/Variant-configuration) it has never encountered before Fairy-Stockfish usually plays at master level without any adaptations.

And this:
> For standard chess, functionality is almost identical with [official Stockfish](https://github.com/official-stockfish/Stockfish), but the slowdown (>2x) due to overhead for fairy pieces and variants leads to >100 Elo weaker performance. When using NNUE the speed difference is lower than with classical evaluation, since the variant code has much less impact on NNUE than on classical evaluation. Actually, NNUE evaluation even is faster than classical, which is why Fairy-Stockfish uses pure NNUE instead of hybrid evaluation.

If anything it's just stronger or the same? Do we need to support multiple game modes now? There's 2 and they are working fine, you're working on a new one and it didn't seem to need a new AI.

Wingdinggaster656:
Stockfish only supports the ELO range from 1350 to 2850.

Done.

Booplicate:
No, not done yet.

Could you help me find that information? For the most part all I see online is "impossible to correlate bot difficulty level to human elo", then there's a few people who says it starts at 800-1000.

So is the goal to lower the difficulty from ~900 "elo" to 500 "elo"?

Wingdinggaster656:
> No, not done yet.
> 
> Could you help me find that information? For the most part all I see online is "impossible to correlate bot difficulty level to human elo," then there's a few people who say it starts at 800-1000.
> 
> So is the goal to lower the difficulty from ~900 "elo" to 500 "elo"?

Please just download the program and type `uci`, then you can see the elo range.

Or you would like the source code. See https://github.com/ianfab/Fairy-Stockfish/blob/master/src/search.cpp#L352
Line 352 to line 366:
```
  // Pick integer skill levels, but non-deterministically round up or down
  // such that the average integer skill corresponds to the input floating point one.
  // UCI_Elo is converted to a suitable fractional skill level, using anchoring
  // to CCRL Elo (goldfish 1.13 = 2000) and a fit through Ordo derived Elo
  // for match (TC 60+0.6) results spanning a wide range of k values.
  PRNG rng(now());
  double shiftedElo = Options["UCI_Elo"] - 1346.6;
  double floatLevel = Options["UCI_LimitStrength"] ?
                      std::clamp(shiftedElo > 0 ? std::pow(shiftedElo / 143.4, 1 / 0.806)
                                                : shiftedElo / 143.4 + std::pow(shiftedElo / 500, 5),
                                 -20.0, 20.0) :
                        double(Options["Skill Level"]);
  int intLevel = int(floatLevel) +
                 ((floatLevel - int(floatLevel)) * 1024 > rng.rand<unsigned>() % 1024  ? 1 : 0);
  Skill skill(intLevel);
```

Please do not trust those who know little.
As commented here, ELO ratings are rigorously tested, anchored by millions of computer games.

And yes. On of the main goal is to reduce the lowest difficulty. From about 1400 to 500.

Wingdinggaster656:
Oh, you may also want to see this page:
https://github.com/ianfab/Fairy-Stockfish/blob/master/src/ucioption.cpp#L195

```
o["UCI_Elo"]               << Option(1350, 500, 2850);
```

This means it has a range from 500 to 2850, and 1350 as the default value.

Booplicate:
Thanks, I checked the sources.

So the latest stockfish appears to have `1350` as the minimum elo while the new engine has `500`.

Bear in mind, our version doesn't have that setting, just skill levels, and also from what I know the new stockfish is stronger and plays generally much better. So I'd think the one we use plays worse than 1350 elo. Even I was able to beat it and I don't play chess often, so I'm definitely below 1350.

Okay, you're suggesting to make it even weaker, but I still can't find anything about playing more like human in the docs. It uses stockfish as the base, I'd think they are about the same in logic. And that's what the docs say, too. Will reducing elo to 500 make it throw even more than it does now?

Wingdinggaster656:
> Thanks, I checked the sources.
> 
> So the latest stockfish appears to have `1350` as the minimum elo while the new engine has `500`.
> 
> Bear in mind, our version doesn't have that setting, just skill levels, and also from what I know the new stockfish is stronger and plays generally much better. So I'd think the one we use plays worse than 1350 elo. Even I was able to beat it and I don't play chess often, so I'm definitely below 1350.
> 
> Okay, you're suggesting to make it even weaker, but I still can't find anything about playing more like human in the docs. It uses stockfish as the base, I'd think they are about the same in logic. And that's what the docs say, too. Will reducing elo to 500 make it throw even more than it does now?

"more human-like" is just a side effect. Indeed a tiny one. I won't introduce it since it relates too many codebase of Stockfish. And since it's a tiny one, you could just forget this one.

Also just to let you know, even if you are 300 ELOs lower, you still have a 15% chance to win.
Winning ≠ Your elo is higher than the opponent.
From what I can see, the old MAS is very likely no less than 1400.

Also, please go back to the PR's introduction:

>Stockfish upgrade. Stockfish's upgrade does not mean an increase in difficulty. The Stockfish update simply means "getting to the target depth faster with the same amount of computing power".

Well, that is actually not 100% correct. It actually could lead a 1% increase. But I don't think this 1% is important.

And throw more? Yes. But I can tell you that this is a must and this is suitable. A real human player at 500 is like that.

Wingdinggaster656:
Oh wait, I may have misunderstood you.

Stockfish 8 is much older than the current new version.
Nowadays, it's easy to make AI look like a real person. While new Stockfish isn't as good as Komodo and some of the other engines, it's still a lot better than Stockfish 8.

So what I just said is that the ELO 500 will throw more than the ELO 1400.

Booplicate:
> it's a tiny one, you could just forget this one.

Well, to me this is the most important part.

> Winning ≠ Your elo is higher than the opponent.

I understand, elo just shows your chance to win against another person. But if my elo is low and I won some games, then I'd think my opponent isn't that strong. Of course you could say I'm just lucky, but then how do we determine elo if we can always say it's just luck. I'm a weak player and even I can see when she throws pieces.

> And throw more? Yes. But I can tell you that this is a must and this is suitable. A real human player at 500 is like that.

Well, I don't want her to throw even more, that's the problem. Idk about "real human players," but she literally gives you her pieces for free, as she *wants* you to take them. It looks like she's giving up. She's still playing as a master, but occasionally throws her pieces to make it *look* like she a weak player. The problem is, it's *obvious* she is not weak. She's still a master.

> Nowadays, it's easy to make AI look like a real person.

Probably, but is that their goal? Are they doing that? I guess we could test. I thought their goal is to optimise the engine and allow it go even further in analysis in real time to make it even stronger.

>  it's still better than Stockfish 8. Maybe not a lot, but seems better anyway.

From [some graphs](https://nextchessmove.com/dev-builds) it looks much stronger.

Wingdinggaster656:
> > it's still better than Stockfish 8. Maybe not a lot, but seems better anyway.
> 
> From [some graphs](https://nextchessmove.com/dev-builds) it looks much stronger.

By the way, by "better than Stockfish 8", I mean it is better on acting like a human, not better on ELO.

Booplicate:
> I will try to find another free and all platforms and human-like engine. I may even just write a new engine on my own.

Before you do more work, let's see what TPG thinks.

Justformas:
> Well, I don't want her to throw even more, that's the problem. Idk about "real human players," but she literally gives you her pieces for free, as she _wants_ you to take them. It looks like she's giving up. She's still playing as a master, but occasionally throws her pieces to make it _look_ like she a weak player. The problem is, it's _obvious_ she is not weak. She's still a master.

Correct me if I'm wrong Wingdinggaster656, but me and you still don't even know what is causing Monika to throw or suicide pieces like her Queen sometimes (at least, on the lowest skill levels). It does not seem normal for Stockfish 8 and I have been unable to replicate it in tests with it so far. So if anyone knows what is causing this it might be good to know.

If we know what's causing it, it could maybe be replicated in another version of Stockfish, but otherwise it shouldn't happen as much.

Wingdinggaster656:
Just to let you know, I spent a day reading through the Stockfish code and discovered that I had the ability to make Stockfish more human. No extra file size required, just some algorithmic improvements.

I tested it. It looked _cool and good_, but the problem was that my algorithm affected Stockfish's ELO by about 80 points on average. Speed also dropped by about 2 percent.

ELO shift problem, I can make computer against computer self-matchmaking later, calibrate new Stockfish ELO.
Since this requires a lot of CPU (I plan to rent some servers for calculation), I would like to confirm first, do you think this plan is feasible?

(Also make sure you know. If we're going to do this, we MUST put a note in MAS directory to leave a Github link of the modified Stockfish source code. This is GPL License.)

Booplicate:
This is a complex project, I propose we return to this after py3 migration. Plus, chess were already migrated to work on py3.

Wingdinggaster656:
> This is a complex project, I propose we return to this after py3 migration. Plus, chess were already migrated to work on py3.

Understood. I will always be here.

ThePotatoGuy (changes requested):
enhancement suggestions here: https://github.com/Monika-After-Story/MonikaModDev/discussions/9118#discussioncomment-14281235

PR #9158, opened 2022-05-21
Topic suggestion: Ameca

enderblaze42:
I recently saw a video on Ameca, an AI with a robot body and the ability to emulate facial expressions. I don't know if this would be too similar to the Sophie topic, but I think it would be neat for Monika to share her thoughts on Ameca. 
Also, apologies if I messed up the formatting of this post. Haven't done anything on Github in a while.


ThePotatoGuy:
suggestions should be on our [issues tab](https://github.com/Monika-After-Story/MonikaModDev/issues).

PR #9260, opened 2022-06-20
NOU fixes

Booplicate:
### Changes:
- ### Improvements:
 - - new pen art, thanks @kkrosie123 
 - - new event which would hint the player to give the new gift
 - ### Bug fixes:
 -  - fixed issue with wild cards retaining their old colour after shuffling
 -  - fixed issue with not being able draw cards when the draw pile is empty (occurred when drew cards by dragging them instead of doubleclicking)
 -  - potentially fixed issue when player could see Monika's card while it was being pulled from the drawpile
 -  - fixed issue with keeping the wrong card in discardpile after shuffling (kept the second last when it should be the last)
 -  - potentially fixed issue with somewhat slow transition occurred during `mas_nou.give_points`, prior there was a chance of a hard pause, now it should be skipped

### Not related changes:
- ### Code cleanup:
- - added `mas_seenEvent`, semi-deprecated `seen_event` (DO NOT REMOVE THE OLD FUNCTION)

### Testing:
- this is hard to test, so verify general stability, play a few games (set points to the min amount for quick games)
- verify the art is good
- verify you get the hint event before gifting
- verify the new event is good (please check for typos)
- verify the gift still works
- verify you do not get the hint event after gifting


jmwall24 (approved):

ThePotatoGuy (approved):
i got home late, so this is a very light skim. hope it works.

PR #9346, opened 2022-07-21
Added more bad nicknames

LordDemecrius83:
Ignore the "wtf is this commit" ,It was added by accident


Wingdinggaster656:
_I must say that I laughed at `cpu eater` the name_

Why `kkk` is a bad name?

Wingdinggaster656:
_I must say that I laughed at `cpu eater` the name_

Why `kkk` is a bad name?

LordDemecrius83:
> _I must say that I laughed at `cpu eater` the name_
> 
> Why `kkk` is a bad name?

short for Ku klux klan

Wingdinggaster656:
> > _I must say that I laughed at `cpu eater` the name_
> > Why `kkk` is a bad name?
> 
> KKK is short for Ku klux klan

I think we should not block `kkk`. That's overkill. And, maybe `cpu eater` should be an awkward name, not a bad name, since it was more or less a joke.

_And by the way if we eventually add `cpu eater` , let us also have `gpu eater` lol_

LordDemecrius83:
> > > _I must say that I laughed at `cpu eater` the name_
> > > Why `kkk` is a bad name?
> > 
> > 
> > KKK is short for Ku klux klan
> 
> I think we should not block `kkk`. That's overkill. And, maybe `cpu eater` should be an awkward name, not a bad name, since it was more or less a joke.
> 
> _And by the way if we eventually add `cpu eater` , let us also have `gpu eater` lol_

Fair enough

I'm also working on fixing window reactions on GNOME but I've made no progress the game would either crash or throw an ignorable exception

Booplicate:
Please keep commits related to the PR, if you want to fix something else, open a new PR.

Booplicate:
Are you cosplaying Rai.

PR #9408, opened 2022-07-28
Fix safe mod issues

Booplicate:
## We still need to figure out do we want `mas_choose_renderer_override` to return or crash. I propose just returning

Fixes:
- fixed issue where there was a possibility of doing `datetime.datetime() - None` 
- moved `quit` label code into the `hide` namespace because there's no reason not to encapsulate
- set `config.performance_test` to `False` to disable GL testing done by renpy
- `mas_choose_renderer_override` now just `return`s instead of silently closing the game (because that had a chance to cause the first issue)

### DIFF
--- a/Monika After Story/game/zz_graphicsmenu.rpy
+++ b/Monika After Story/game/zz_graphicsmenu.rpy
@@ -388,5 +388,8 @@ label mas_gmenu_start:
     return
 
 label mas_choose_renderer_override:
-    # if we do this somehow, just quit immediately.
-    jump _quit
+    # This label can be called by renpy when launched in safe mod
+    # (holding shift on windows)
+    # We provide our own renderer menu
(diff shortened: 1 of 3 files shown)


Monika After Story/game/zz_graphicsmenu.rpy
```diff
@@ -388,5 +388,8 @@ label mas_gmenu_start:
     return
 
 label mas_choose_renderer_override:
-    # if we do this somehow, just quit immediately.
-    jump _quit
+    # This label can be called by renpy when launched in safe mod
+    # (holding shift on windows)
+    # We provide our own renderer menu
+    # and this one should be disabled, so we just return
+    return
```
Booplicate, line 395: Could probably log that we got here, but I'm not sure if our logs would be initialised.

PR #9476, opened 2022-08-25
safeToRefDokis added to glitch greeting

Bluecoreg:
Changes `greeting_glitch` to check if the player is okay with references to other dokis.


Booplicate:
The code doesn't look correct, if anything, it should be a `conditional`.
But also, that's not what that var is being used for. It's used when there's a joke about a doki or directly referencing "sensitive" events of DDLC. Just mentioning a doki or showing them is valid. Additionally, the greeting will be reworked to not show Yuri at all.

PR #9491, opened 2022-09-04
Potentially illogical dialogue fixes

Wingdinggaster656:
`monika_lastpoem` is available for everyone, and that doesn't make sense to people who have never seen the poem.

A little check was made here to make sure that anyone reading this topic must have read the last poem.


Booplicate:
I'm not sure if this requires act 3.
Also it will need an update script.

Wingdinggaster656:
> I'm not sure if this requires act 3.

Yeah, I don't think it necessarily requires the player to reach ACT 3, too.
Here the current `conditional` is more like giving a prototype. It would be better if we had a more accurate way to judge.

Wingdinggaster656:
>Also it will need an update script.

Changed in 712b8e4

Booplicate:
Not really, 0.12.9 is the current version, your branch is out of date. You'll also need a sanity check so people who saw it, keep it as is.

Wingdinggaster656:
> Not really, 0.12.9 is the current version, your branch is out of date. You'll also need a sanity check so people who saw it, keep it as is.

Changed in f4c785a, _I guess_.

Booplicate (approved):

ThePotatoGuy (approved):

ThePotatoGuy (dismissed):

Booplicate (approved):

jmwall24 (approved):

Issue #9388, opened 2022-07-24
post_proc_data assumes hair and acs always exist when it may not


ThePotatoGuy:
code in `_validate_clothes`:
```python
1807         # some additional processing for outfit_hair and outfit_acs$
1808         if "outfit_hair" in save_obj:$
1809             post_proc_data["outfit_extras"][sp_name]["hair"] = save_obj.pop($
1810                 "outfit_hair"$
1811             )$
1812         if "outfit_acs" in save_obj:$
1813             post_proc_data["outfit_extras"][sp_name]["acs"] = save_obj.pop($
1814                 "outfit_acs"$
1815             )$
```

code in `processOutfitExtras`
```python
2654         for sp_name in outfit_extras:$
2655             outfit_hair = outfit_extras[sp_name]["hair"]$
2656             outfit_acs = outfit_extras[sp_name]["acs"]$
2657 
```
A simple `.get` should be enough to fix. I'll do this when i've got time, today I'm super busy.

ThePotatoGuy:
[closing as fixed]

Issue #9387, opened 2022-07-24
I think there's an issue with the "outfit_acs" property on outfit mode.


Dsfan2:
![Screenshot 2022-07-24 143053](https://user-images.githubusercontent.com/82821066/180664852-0e8315e6-8551-464a-afb5-7169697d096e.png)
This happens only when outfit mode is on. The outfit selected should be displayed but it doesn't. As far as I know, it only happens to outfit jsons that have the "outfit_acs" property.

ThePotatoGuy:
can you upload your spj.txt file here?

ThePotatoGuy:
nvm i am working on it.

ThePotatoGuy:
[closing as fixed]

PR #9500, opened 2022-09-10
outfit JSON props and outfit mode fixes

ThePotatoGuy:
# Key Changes 
* fixes crash when loading sprite objects only have either `outfit-hair` or `outfit-acs` properties (#9388)
* fixes `MASClothes.outfit_acs` getting set with incorrect data, causing the selectors to glitch out (#9387)
* prevents selectors from entering glitch selection state when changing clothes/hair fails.

# Testing

### outfit json props
1. modify a sprite JSON so it includes either `outfit-hair` or `outfit-acs`, but **not** both.
2. Launch MAS and verify the sprite object loads 

### selector ACS issues
1. use a clothes sprite JSON that includes `outfit-acs`
2. verify it works when changing to that clothing item in the clothing selector works when outfit mode is enabled 
    * `Hey Monika > Appearance > Can I change your clothes?`

### selector clothes/hair crash issues
1. modify a clothing/hair progpoint so it crashes
2. try to change clothing/hair to that modified clothes/hair.
3. verify that the clothes/hair does **not** get applied, and that the selection is consistent

### DIFF
--- a/Monika After Story/game/zz_selector.rpy
+++ b/Monika After Story/game/zz_selector.rpy
@@ -1028,6 +1028,7 @@ init -10 python in mas_selspr:
     def _adjust_monika(
             moni_chr,
             old_map,
+            prev_map,
             new_map,
             select_type,
             use_old=False,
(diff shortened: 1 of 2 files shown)


multimokia (approved):

kkrosie123 (approved):
Upon testing...
- JSON Outfit/Hair props work
- ACS JSON props select/unselect when switching outfits in outfit mode
- Selector crash issues section seems to work, selection consistent etc

Approved

PR #9510, opened 2022-09-11
[contributing] [mod] new mood - annoyed

my-otter-self:
this adds a new "i feel" option, the annoyed mood. looking at the dialogue today, i personally do not like the dialogue here. but i still think an "annoyed" mood might be interesting. let's ping pong ideas!


Booplicate:
First, base your PR on the `content` branch. Also, let's do 1 mood per PR.

my-otter-self:
noted! will start again and fix this.

PR #9518, opened 2022-09-11
[contributing] [mod] "im going to voice chat with a friend" brb

my-otter-self:
new "be right back" option, "im going to voice chat with a friend".


Booplicate:
Brbs should be generic, you already proposed a brb for a call #9519

PR #9521, opened 2022-09-11
[contributing] [mod]  "i'm going to download something" brb.

my-otter-self:
adds a new "be right back", "i'm going to download something".


Booplicate:
I don't think this would have much of use, it's too specific and would rarely be used.

PR #9524, opened 2022-09-11
[contributing] [mod] "i'm going to write on my journal" brb

my-otter-self:
adds a new "be right back", "i'm going to write on my journal".


Booplicate:
I'm not sure if this is needed, we have `monika_idle_writing`, if needed, it should be expanded instead.

PR #9523, opened 2022-09-11
"i'm going to do some housework" brb.

my-otter-self:
adds a new "be right back", "i'm going to do some housework"

PR #9525, opened 2022-09-11
[contributing] [mod] "i'm going to take care of my plants" brb

my-otter-self:
new "be right back", "i'm going to take care of my plants".


Booplicate:
Again, this is too specific and gets covered in another brb #9523

PR #9526, opened 2022-09-11
[contributing] [mod] "i'm going to stretch my legs" brb

my-otter-self:
adds a new "be right back", i'm going to stretch my legs."


Booplicate:
`monika_idle_screen_break` covers this already, we could add a line or two about stretching there.

Issue #7963, opened 2021-09-29
[ENHANCEMENT] Don't raise exceptions if a character is missing in a font group


Booplicate:
Current situation:
- `renpy.text.font.FontGroup` raises an exception if the user is missing the font (in both our version and r7.4, the method is the same)

(method `segment`)
```py
if font is None:
    raise Exception("Character U+{0:04x} not found in FontGroup".format(n))
```

Proposed enhancement
- Override the method to only raise in dev mode, otherwise replace the missing character with something else

PR #9582, opened 2022-09-22
Replace special Windows characters in path name (fix #9557)

### LINKED ISSUE #9557: Errno 22 in case player has special characters in their name
In case player has special characters in their name (in my case, one of Discord users who shared an error had `:D` in their name, could it be done in circumvention of name choosing system or was it possible in earlier releases?), this line will raise EINVAL on Windows machines when MAS will attempt to create a 'For \<name\>.txt' file.

https://github.com/Monika-After-Story/MonikaModDev/blob/da46bc6f1152ec3b023a68df27caf95fa045a725/Monika%20After%20Story/game/script-holidays.rpy#L7286

Traceback follows:
```
I'm sorry, but an uncaught exception occurred.

While running game code:
  File "game/script-ch30.rpy", line 1541, in script call
  File "game/event-handler.rpy", line 3119, in script call
  File "game/script-holidays.rpy", line 7266, in script
  File "game/script-holidays.rpy", line 7300, in <module>
  File "game/script-affection.rpy", line 2707, in _write_txt
IOError: [Errno 22] invalid mode ('w') or filename: u'D:/! DDLC-1.1.1-pc/characters/For XXX :D.txt'

-- Full Traceback ------------------------------------------------------------

Full traceback:
  File "game/script-ch30.rpy", line 1541, in script call
  File "game/event-handler.rpy", line 3119, in script call
  File "game/script-holidays.rpy", line 7266, in script
  File "D:\! DDLC-1.1.1-pc\renpy\ast.py", line 814, in execute
  File "D:\! DDLC-1.1.1-pc\renpy\python.py", line 1719, in py_exec_bytecode
  File "game/script-holidays.rpy", line 7300, in <module>
  File "game/script-affection.rpy", line 2707, in _write_txt
IOError: [Errno 22] invalid mode ('w') or filename: u'D:/! DDLC-1.1.1-pc/characters/For XXX :D.txt'

Windows-8-6.2.9200
Ren'Py 6.99.12.4.2187
Monika After Story 0.12.9
```

dreamscached:
This PR fixes #9557.


Monika After Story/game/script-holidays.rpy
```diff
@@ -7264,7 +7264,12 @@ image chibi_peek = MASFilterSwitch("mod_assets/other/chibi_peek.png")
 label mas_bday_surprise_party_hint:
     #Set up letters
     python:
-        persistent._mas_bday_hint_filename = "For " + player + ".txt"
+        player_sane = player
+        if renpy.windows:
```
ThePotatoGuy, line 7268: just remove the extra stuff for all os, linux will get annoyed with some of those too

ThePotatoGuy (changes requested):

Monika After Story/game/script-holidays.rpy
dreamscached: Got it.

Booplicate (changes requested):
This looks inflexible to me. For one, it should be a function in utils we can reuse.
But also, we should disallow those characters in the first place. The name may only contain ascii alpha characters. We have a predefined string constants you can use.

dreamscached:
Will do. Any specific place I can place an util function to?

ThePotatoGuy:
no it should not contain just ascii alpha characters - that leaves everyone who wishes to uses non latin-character names unable to use their name.

Booplicate:
DDLC fonts can't display non-latin characters and in the worst case it will cause renpy to crash. If any translations or submods decide to do the extra handling required, they can update the constants to allow non-latin characters.

dreamscached:
I think so too, was just about to mention that translations can relax the restriction.

ThePotatoGuy:
I [patched](https://github.com/Monika-After-Story/MonikaModDev/blob/master/Monika%20After%20Story/game/styles.rpy#L402) the `gui.defaultfont` so it has glyphs for korean, chinese, jp, and latin-1. it works by default.

Booplicate:
It doesn't handle everything because I'm still getting crashes: #7963

ThePotatoGuy:
then we add the missing glyphs. We are not restricting acceptable names, thats culturally insensitive.

Booplicate:
Fair, we can add more fonts.
But we should still remove special characters like `:` or `>` from the name input.

TheT0matoGuy:
> I [patched](https://github.com/Monika-After-Story/MonikaModDev/blob/master/Monika%20After%20Story/game/styles.rpy#L402) the `gui.defaultfont` so it has glyphs for korean, chinese, jp, and latin-1. it works by default.

I kind of want to point out that while the fonts you've added do cover most of the characters in Chinese, there are some symbols that are strangely unsupported.

Chinese uses different punctuation marks from English. English comma: ', ' and Chinese comma: '，' This is an example of not being supported. If that's okay with you, would you consider adding these punctuation marks?

Booplicate:
@ThePotatoGuy just found out another case where we're missing glyphs.
image

So there's actually 3 possible outcomes if we use non-ascii in the code:
- crash
- completely invisible characters possible messing up layout
- characters are being replaces with something else making the text unreadable nonsense

I'm not a native English speaker, yet I don't find not supporting all of the utf spectrum "culturally insensitive." Furthermore with all the issues we have right now, I can say we already don't support it. We allow to use it, but considering you either won't see the text or will crash, I can't call it "support." It's just making the game unstable.

Also, a lot of the games today only allow to input ascii because it has much wider support (yes, in 2022!) and is less error-prone. We could consider adding utf support in the future (perhaps after python3 with proper unicode), but right now I don't see it.

ThePotatoGuy:
fair enough, I also tried to type jp input directly into name changing and it glitched out hitting enter, so all the screenshots ive seen with jp names must have had another way to set them. keeping our name input ascii is fine with me now.

dreamscached:
Should I also do something about character inputs?

Booplicate:
No, it should go into a separate PR.

dreamscached:
Got it.

Monika After Story/game/definitions.rpy
```diff
@@ -5377,6 +5378,13 @@ init -100 python in mas_utils:
         return (ival, int((num - cleanival) * __FLIMIT), __FLIMIT)
 
 
+    def sanitize_filename(s):
```
Booplicate, line 5381: Please add some basic documentation.

Monika After Story/game/definitions.rpy
```diff
@@ -5377,6 +5378,13 @@ init -100 python in mas_utils:
...
+        sane = s
```
Booplicate, line 5382: This statement has no actual effect.

Monika After Story/game/script-holidays.rpy
```diff
@@ -7283,7 +7283,7 @@ image chibi_peek = MASFilterSwitch("mod_assets/other/chibi_peek.png")
 label mas_bday_surprise_party_hint:
     #Set up letters
     python:
-        persistent._mas_bday_hint_filename = "For " + player + ".txt"
+        persistent._mas_bday_hint_filename = mas_utils.sanitize_filename("For " + player + ".txt")
```
Booplicate, line 7286: Let's fix this string concatenation
```suggestion
        persistent._mas_bday_hint_filename = mas_utils.sanitize_filename("For {}.txt".format(player))
```

Booplicate (changes requested):

dreamscached:
Anyway, any further changes I must make on this one?

Booplicate:
Looks good, I think I can test it and we can get in the next release.

dreamscached:
Great, thanks a lot.

ThePotatoGuy (approved):

Booplicate (approved):

Issue #9376, opened 2022-07-23
Clothes - 2B


hatanasinclaire:
![image](https://user-images.githubusercontent.com/96905447/180622136-76237052-e038-422f-952d-f98d1f7da3a1.png)

![image](https://user-images.githubusercontent.com/96905447/180622064-b4821bf4-320d-4e5c-afe9-92b4ae961872.png)

#3174

It is a new costume for o31. [YoRHa No.2 Type B](https://drive.google.com/file/d/1d1E8lDtSH7xifDyQOY78rBEVDbNfcSb7/view?usp=sharing) from _NieR: Automata_.

It was requested in the [clothes wishlist](https://docs.google.com/spreadsheets/d/1C2xJVdrFX9GLV8b6J2wcBaMxlcCn9LCbZs43Q9TnTr0/edit#gid=505708911).

ThePotatoGuy:
ok i am going to put these in next rel

ThePotatoGuy:
[closing as implemented]

PR #9598, opened 2022-09-24
2B spack

ThePotatoGuy:
#9376 

# Key Changes
* adds 2b spack - thumbs is already included
* adds 2b o31 intro
* adds "hide-ring" ex-prop
## TODO
* [x] add o31 intro dialogue

# Testing
* verify looks good
* verify ring is hidden when outfit is worn and is worn again when outfit is taken off (if applicable)
* verify o31 intro event works


jmwall24 (changes requested):
We need to add an ex_prop to toggle the ring off for gloves so we don't have potential clipping

hatanasinclaire:
> We need to add an ex_prop to toggle the ring off for gloves so we don't have potential clipping

The gloves actually obscure the ring entirely on poses 2, 3, and 5! The position of the ring finger makes it so players would have to have replaced the ring with something extremely wide for it to not be completely covered.

jmwall24:
@hatanasinclaire the issue is some players have commissioned rings with stones which is what worries us a bit. Having an ex_prop available for gloves (not just the gloves here, but any gloves) is an easy, safe way of making sure no one ends up with clipping.

Booplicate (approved):
@hatanasinclaire great work on the outfit

jmwall24 (approved):

Legendkiller21 (approved):

PR #9617, opened 2022-09-24
Category capitalization fix 

hatanasinclaire:
It is a very minor fix for the display of the "DDLC..." and "Literature Club..." categories in the "Hey, Monika..." / "Repeat Conversation" tabs.

cf3

※There are a few times in the dialogue of *DDLC* where "literature club" is not capitalized, but in most instances "Literature Club" is capitalized as a proper noun and name of a (small) organization.

### DIFF
--- a/Monika After Story/game/event-handler.rpy
+++ b/Monika After Story/game/event-handler.rpy
@@ -3440,7 +3440,17 @@ label prompts_categories(pool=True):
         # tuplelize the main the category list
         # NOTE: we use a 2nd list here to do displaying, keeping track of the
         # older cat list for checking if a category was picked
-        dis_cat_list = [(x.capitalize() + "...",x) for x in main_cat_list]
+        dis_cat_list = []
+        for x in main_cat_list:
+            if x.lower() == "ddlc":
+                capitalized_string = x.upper() + "..."
+            elif x.lower () == "literature club":
+                capitalized_string = "Literature Club..."
+            else:
+                capitalized_string = x.capitalize() + "..."
+
+            dis_cat_list.append((capitalized_string, x))
+
 
         # tupelize the event list
 #        no_cat_list = evhand.tuplizeEventLabelList(no_cat_list, unlocked_events)


Monika After Story/game/event-handler.rpy
```diff
@@ -3440,7 +3440,17 @@ label prompts_categories(pool=True):
...
         # older cat list for checking if a category was picked
-        dis_cat_list = [(x.capitalize() + "...",x) for x in main_cat_list]
+        dis_cat_list = []
+        for x in main_cat_list:
+            if x.lower() == "ddlc":
+                capitalized_string = x.upper() + "..."
+            elif x.lower () == "literature club":
+                capitalized_string = "Literature Club..."
+            else:
+                capitalized_string = x.capitalize() + "..."
```
multimokia, line 3450: single-catch cases like this should be avoided in favour of more generic, reusable solutions.

Perhaps it's best to remove the capitalization function entirely in categories and simply make events use the capitalization they require.

Or to help deal with typos/duplicate categories down the line, given the move to renpy 8 in the future, maybe we could use enums?

multimokia (changes requested):

Monika After Story/game/event-handler.rpy
hatanasinclaire: > Perhaps it's best to remove the capitalization function entirely in categories and simply make events use the capitalization they require.

image
By that do you mean capitalizing these in the event information for each topic? Sorry if I am misunderstanding.

multimokia: Yes. However before you do that, in the interest of saving you the time/effort in case this is no'd: @ThePotatoGuy, what are your thoughts here?

Booplicate: Can we please go the enum approach? We had problems before when we missed a typo in a category. Right now you could just use a static class and later on replace with an enum when we get r8.

multimokia: This is what I was thinking too. Definitely want enums here as it'd make it soooo much cleaner to migrate too.

hatanasinclaire: Very sorry. I do not understand enum too well. Does this mean a new class that has manually specified capitalization values for each category that exists, or something else?

ThePotatoGuy: enum approach is fine for py3 as long as we support basic strings in the category list, too.

Issue #8596, opened 2022-01-11
sprite system folder-based organization


ThePotatoGuy:
current file structure is:
* clothes: `c / <folder name as ID> / images*`
* hair: `h / <image prefix as ID><images*>`
* ACS: `a / <image prefix as ID><images*>`

Suggested changes:
* hair: `h / <folder name as ID> / images*`
* ACS: `a / <folder name as ID> / images*`

Other changes:
* make sure both the current way and the new way are supported
    * add new json prop and sp object prop to determine image structure - maybe `location` or `structure`

Pros:
* simplification of file names - only need to do front/back/mid/0/1/etc.., just make sure is in correct folder

Cons:
* none, as long as we support the old way - otherwise, all current hair and ACS spritepacks would break.

ThePotatoGuy:
will prioritize after next rel

ThePotatoGuy:
notes for myself:

* check for `MASHair.hl_keys_c` usage
* check for `MASHair.prepare_hl_data` usage

PR #9619, opened 2022-09-25
folder-based hair and ACS sprites

ThePotatoGuy:
#8596 

# Key Changes
* hair and ACS sprite objects now support folder-based organization, basically similar to clothes:
    * `h/<hair name>/front.png` or `h/<hair name>/0.png`
    * `a/<acs name>/0.png`
* sprite JSONs support the new format as well
   * this is based on a version - so version 4 and beyond must be folder-based
   * ver 3 sprites will continue to be supported but a warning will be logged to consider updating the version
* a simple bash script for renaming using git is provided, but a full-fledged spritepack updater will be added in #9729 

# Testing
* verify that all current sprites load correctly
* verify spritepacks with jsons **do not break** - testing of the new json props should be in #9729 for ease of testing


Booplicate (commented):
Maybe instead of adding another param (`use_folders`), use spritepack version to determine how it should be loaded? More params > harder for people to understand our system and what params to use. We should also discourage people from using the old structure, having an upgrade flag instead of an option is better for that case imo.

We should also update our wiki

ThePotatoGuy:
changed to use version to determine `use_folders`.

Also rebased commits so all main code changes are in 1 (next time I'll do a better job of splitting these out)

Booplicate:
Besides `use_numeric_layer_ids` not using `version` I think it's great

ThePotatoGuy:
down the line that might be an option, the key benefit of numeric ids is more splits beyond 3, it's just not necessary now so.


i'll merge tonight

Legendkiller21 (approved):

PR #9626, opened 2022-09-26
Timetravel hardstop event

Booplicate:
Since people keep messing with the time while knowing it will break the mod, we need to add a hard stop in case the time is incorrect. This way they can at least restore a backup before it's too late. Also makes sense for Monika to react on what's happening.

### Changes:
- added the new event
- new system for handling tt
- kazoo is now officially in and is used

### Testing:
- verify there's no false positives with time travelling
- try to break the time and verify you get the hard stop event
- verify you can't load in afterwards
- verify the expressions and writing are good (not that we care, we do **not** want people to actually see this event)


Monika After Story/game/splash.rpy
```diff
@@ -175,6 +175,15 @@ label splashscreen:
         # We're about to start, all things should be loaded, we can check event conditionals
         Event.validateConditionals()
 
+        guard = mas_tt_guard.MASUptimeSyncValidator(
+            on_desync=lambda: (
+                setattr(store.mas_globals, "tt_detected", True),
```
ThePotatoGuy, line 180: so you went with having the thread call a function if the time sync changes. 

I was thinking this would just set its own props if tt is detected, and we'd poll it in ch30_loop. that way any mutex/cond logic can be part of the class instead of having to reimplement it in the passed in function.

Mainly because this line is not-thread safe. There should be a mutex lock here, but that also affects anywhere that global is used, which could be us + submods. The push line looks probably safe for now, but that's not a guarantee (and I don't think we need to make it a guarantee).

ThePotatoGuy (changes requested):

Monika After Story/game/splash.rpy
Booplicate: In this case we don't need a mutex, it's an atomic operation (in the end it's setting a `dict` key to a value), no memory corruption can happen.

Booplicate: Also, I think at some point our main data types (like `Event`, `MASEventList`, and others) should become thread safe. One of the things I'd like to have is async ch30 loop, and for that we'd need to add locks to some of our systems.

ThePotatoGuy: I know the py data structures do atomic set, but this is several abstractions that we'd have to make sure are always true.
Does r8 continue to not implement __setattr__ for the renpy store module? Will it always do that? Will push only append to list? it already has an option to do slightly more than that. 

This also implies that others who want to use the sync thread need to spin their own. Thats a waste when we could just have 1 thread and everyone polls it when needed.

Booplicate: > Does r8 continue to not implement setattr for the renpy store module? Will it always do that? Will push only append to list? it already has an option to do slightly more than that.

It's still safe in r8. I can't think of any big change that would cause a race condition there. Something like "if not set, set now" would be a problem, for example. But it doesn't make sense to have - you want always allow to set a value. Can you give an example?

> This also implies that others who want to use the sync thread need to spin their own.

This thread is private for a reason, I don't want "others" to mess with it. And if somebody needs to, they can already check the var in ch30_loop.

ThePotatoGuy: if the store module is just setting a key:value in a dict, all it takes is an iterator running at the same time to mess that up. or a del op from elsewhere. but the crux is that this is making an assumption that does not need to be made, there are ways to exchange data between threads that absolves any assumptions.

> This thread is private for a reason, I don't want "others" to mess with it. And if somebody needs to, they can already check the var in ch30_loop.

it's not just about TT, being able to poll an accurate time elapsed could certainly find uses elsewhere. I'm already thinking that we could actually differentiate between timezones change and time change by comparing utc differences with elapsed time.

Booplicate: > all it takes is an iterator running at the same time to mess that up. or a del op from elsewhere

Hmm, since we're changing a value (not removing it), the size remains the same, so there's no corruption can happen. `del` is an odd one, we shouldn't delete that global var, but even if we did, the thread would just set it back no big deal. So it looks safe to me still.

> being able to poll an accurate time elapsed could certainly find uses elsewhere. I'm already thinking that we could actually differentiate between timezones change and time change by comparing utc differences with elapsed time.

Okay, that sounds cool, I'll make some changes to account for that.

Booplicate: Btw, while I made the changes, I found an interesting line:
> This thread is very limited in what it can do with the Ren'Py API. Changing store variables is allowed, as is calling the [renpy.queue_event()](https://www.renpy.org/dev-doc/html/other.html?highlight=thread#renpy.queue_event) function.

which means storemodules are thread safe and are probably going to stay that way.

Booplicate: > async code isnt multi-threaded and therefore doesnt actually have to worry about mem management

Depends, sometimes incorrect usage of promises or forgetting a lock might cause issues too. But threads are more sensitive. Either way, just good to know threading is supported by renpy to that extent.

PR #9628, opened 2022-09-27
Backup awareness

Booplicate:
This makes Monika "aware" of when the user restores a backup.
A few reasons why:
- awareness is always cool, can catch people by surprise
- this also fixes an issue when loading a very old backup and losing affection for it
- sometimes it's not clear enough that a backup has been restored, this is a good way to tell the player "something happened"

### Testing
- verify the main logic works:
- - make sure you have `persistent` in your save dir
- - - if not, load in and quit to quickly generate a fresh pers
- - load in with the new changes
- - you should get a new backup in your save dir
- - quit normally
- - restore the new backup
- - verify Monika mentions that you restored a backup
- - restart again, verify Monika doesn't mentions it again (the flag should be reset)
- verify no aff lose when loading an old backup
- verify expressions and writing are good
- - also check out post expressions
- verify I didn't leave more `persisten` fypos...

### DIFF
--- a/Monika After Story/game/0config.rpy
+++ b/Monika After Story/game/0config.rpy
@@ -113,6 +113,12 @@ init -1200 python:
     renpy.config.menu_clear_layers = ["front"]
     renpy.config.gl_test_image = "white"
 
+    # Remove extra save location (for whatever reason we do this)
+    # FIXME: Also, this is unsafe since renpy has a thread that
+    # iterates thru this list
+    if len(renpy.loadsave.location.locations) > 1:
+        renpy.loadsave.location.locations.pop()
+
 ################START: INIT TIME CONFIGS
 
 ## Uncomment the following line to set an audio file that will be played while
--- a/Monika After Story/game/zz_backup.rpy
+++ b/Monika After Story/game/zz_backup.rpy
@@ -669,6 +669,9 @@ init -900 python:
                 renpy.save_persistent()
                 numnum, numnum_del = __mas__backupAndDelete(p_savedir, "persistent")
 
+            except:
+                raise
+
             finally:
                 persistent._mas_is_backup = is_pers_backup
                 renpy.save_persistent()
--- a/Monika After Story/game/zz_backup.rpy
+++ b/Monika After Story/game/zz_backup.rpy
@@ -669,9 +669,6 @@ init -900 python:
                 renpy.save_persistent()
                 numnum, numnum_del = __mas__backupAndDelete(p_savedir, "persistent")
 
-            except:
-                raise
-
             finally:
                 persistent._mas_is_backup = is_pers_backup
                 renpy.save_persistent()
--- a/Monika After Story/game/zz_backup.rpy
+++ b/Monika After Story/game/zz_backup.rpy
@@ -17,6 +17,8 @@ default persistent._mas_incompat_per_rpy_files_found = False
 # only set if the user entered the incompat flow at all
 default persistent._mas_incompat_per_entered = False
 
+default persisten._mas_is_backup = False
+
 python early in mas_per_check:
     import __main__
     import cPickle
@@ -661,7 +663,15 @@ init -900 python:
         try:
             p_savedir = os.path.normcase(renpy.config.savedir + "/")
 
-            numnum, numnum_del = __mas__backupAndDelete(p_savedir, "persistent")
+            try:
+                persisten._mas_is_backup = True
+                renpy.save_persistent()
+                numnum, numnum_del = __mas__backupAndDelete(p_savedir, "persistent")
+
+            finally:
+                persisten._mas_is_backup = False
+                renpy.save_persistent()
+
             __mas__backupAndDelete(p_savedir, "db.mcal", numnum=numnum)
 
         except Exception as e:
--- a/Monika After Story/game/zz_backup.rpy
+++ b/Monika After Story/game/zz_backup.rpy
@@ -193,7 +193,7 @@ python early in mas_per_check:
 
     def should_show_chibika_persistent():
         """
-        Should we show the chibika persistent dialogue? 
+        Should we show the chibika persistent dialogue?
 
         RETURNS: True if we should show the chibika persistent dialogue
         """
@@ -239,7 +239,7 @@ python early in mas_per_check:
 
         # first, check if we have a special persistent
         if os.access(_sp_per, os.F_OK):
-            #  we have one, so check if its valid 
+            #  we have one, so check if its valid
             try: # TEST_CASE_A
                 per_read, version = tryper(_sp_per)
 
@@ -367,7 +367,7 @@ python early in mas_per_check:
 
             try: # TEST_CASE_F
                 shutil.copy(_cur_per, _sp_per)
-                os.remove(_cur_per) 
+                os.remove(_cur_per)
 
                 # and then close out of here - the game should generate a fresh
                 # persistent.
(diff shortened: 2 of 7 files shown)


jmwall24 (dismissed):

Monika After Story/game/0config.rpy
```diff
@@ -113,6 +113,12 @@ init -1200 python:
     renpy.config.menu_clear_layers = ["front"]
     renpy.config.gl_test_image = "white"
 
+    # Remove extra save location (for whatever reason we do this)
+    # FIXME: Also, this is unsafe since renpy has a thread that
```
ThePotatoGuy, line 117: should we just not do this now if its unsafe?

Monika After Story/game/zz_backup.rpy
```diff
@@ -660,13 +662,23 @@ init -900 python:
...
-            p_name = "persistent"
-            numnum, numnum_del = __mas__backupAndDelete(p_savedir, p_name)
-            cal_name = "db.mcal"
-            __mas__backupAndDelete(p_savedir, cal_name, numnum=numnum)
+            is_pers_backup = persistent._mas_is_backup
+
+            try:
+                persistent._mas_is_backup = True
+                renpy.save_persistent()
+                numnum, numnum_del = __mas__backupAndDelete(p_savedir, "persistent")
```
ThePotatoGuy, line 670: I assume without an except block this will just continue raising the exception? even so, its much easier to follow this if the exception is explicitly raised here.

ThePotatoGuy (changes requested):

Monika After Story/game/0config.rpy
Booplicate: We've been doing that for years, I'm not sure how bad it is, but it's something I wanted to mark as look into later. I don't think I can remove it right away either - it might mess up our auto backups.

Monika After Story/game/zz_backup.rpy
Booplicate: `try` has optional sub-blocks, `except`, `finally`, `else`. It makes sense to use only the ones you need. We don't care what it will raise or will it raise anything at all, we care about running the `finally` block no matter what happens. Linters would mark that as a "useless except as statement".

ThePotatoGuy: ok but linters don't have to read the code later and maintain it, and we do care about the exception here because `numnum` is used after this block so if the exception isn't raised out of here then we'll get an irrelevant exception logged. (Even if the exception is raised out of here because python, I would prefer the explicit exception re-raise for readability.)

Booplicate: To me it clutters the code, implying we're doing some work with exceptions, but we don't. We don't care about what's being raised, we care about the code in `finally` being run. If we get an exception, the code after this block wouldn't run.

If you meant whether or not that code suppresses the exception, no, it doesn't, we don't do anything to the exception raised (you could say we propagate it). If we crash there, the exception would be raised as usual, but the cleanup code in `finally` would be run. Which is the goal.

I'll add it for you, but I think it's a bad example of explicitness where it hurts readability.

ThePotatoGuy: this is a recognizable pattern in different languages (minus except vs catch):
```python
try:
    # something
except Exception as e:
    raise e
finally:
    # something
```

`finally` is optional in other languages, but `except`/`catch` is not. I think python is in the wrong here for allowing `except` to be optional, and I would say the notation that was committed (blank except and raise) is also a bad python construct. Defining the exception and re-raising is a pattern that can be understood without having to know that python allows exceptions to propagate when the except block isn't there in a try-except pattern.

Booplicate: And this is one of the reason I didn't want to add it. You got confused right away.

> the notation that was committed (blank except and raise) is also a bad python construct.

This actually the place where we *should* to use bare except, using `Exception` would limit it to only this type, which is incorrect - we want to run this code for all cases.

> Defining the exception and re-raising is a pattern that can be understood without having to know that python allows exceptions to propagate when the except block isn't there in a try-except pattern.

And for people that know about Python it makes the code look like it's doing some exception handling while it doesn't. Plus, makes you question why the bare except.

ThePotatoGuy: >And this is one of the reason I didn't want to add it. You got confused right away.

Without except is worse - what happens if there is an exception? My first thought is that it ignores it since it isn't explicitly re-raised,

>This actually the place where we should use bare except (if you want to use except at all), using Exception would limit it to only this type, which is incorrect - we want to run this code for all cases.

The logging code in the outer exception catches from base Exception - so its following the same pattern. All exceptions should be using Exception as a base class, exceptions that aren't would have broken through the original try-except and probably all through renpy code anyway, so its not something that we'd have to handle.

> And for people that know about Python it makes the code look like it's doing some exception handling while it doesn't. Plus, makes you question why the bare except.

Those people will clearly see that it re-raises the exception. And don't do a bare except, do the `Exception as e` so there's no ambiguity. 

>I don't think trying to apply other langs syntax and logic to Python is always a good thing

It's applying conventional programming concepts for readability. Other pythonic notation like `<x> in <y>` or `a < x < b` is great for readability so we should do them. optional `except` is not because it just seems like the exception is ignored entirely.

Booplicate: > so its following the same pattern. All exceptions should be using Exception as a base class

No, the block I added should catch *everything* (and by everything I mean all renpy exception and all base py exception.). That's the goal. What happens to the outside try/except isn't my concern. My concern is running the `finally` block I added.

ThePotatoGuy: >My concern is running the finally block I added.

`finally` always runs after a try-except pattern, whether or not a bare except is used doesn't affect the execution of the `finally` block. I'm concerned with readability using a python construct that obscures behavior.

Booplicate: > whether or not a bare except is used doesn't affect the execution of the finally block.

I believe it does. We have to use either implicit propagation of everything (no `except`), or explicit bare `except` with manual propagation of everything. I'm also not sure if that would clutter the trace stack with an extra exception or substitute it.
The `finally` part must be run no matter what, even if renpy decides to throw an error, or Python interpreter crashes.

Readability is important, but not when it hurts the code. There is a thing when the code is too explicit/has too many comments.

ThePotatoGuy: >I believe it does. We have to use either implicit propagation of everything (no except), or explicit bare except with manual propagation of everything. I'm also not sure if that would clutter the trace stack with an extra exception or substitute it.

This is wrong, `finally` always runs after a try-except pattern in standard execution, with the only exceptions being outside influences like the computer dying or a sigkill ([py-doc](https://docs.python.org/3/reference/compound_stmts.html#finally-clause)).

>Readability is important, but not when it hurts the code. 

I don't think it hurts the code here. It's slightly more verbose for an obvious, no-prior-python-knowledge-required understanding of what happens to the exception. I would prefer the `except` over a comment here since its a conventional pattern.

Booplicate: Example in the docs uses the syntax I used here before:
```py
    try:
        1/0
    finally:
        return 42
```
What happens if a renpy exception occurs and we don't catch it?

ThePotatoGuy: Like I said earlier, I think python is wrong for allowing except to be optional. We don't need to follow "pythonic practices" if its bad for readability - like `try-else` for example. 

>What happens if a renpy exception occurs and we don't catch it?

if the exception is occurring within the python instance then `finally` should still run regardless. If its happening outside (like it kills the python execution), then it doesn't matter if there were an except here or not, `finally` wouldn't run.

ThePotatoGuy: Ok, nvm, after looking at [wiki](https://en.wikipedia.org/wiki/Exception_handling_syntax), I see that there are other languages that don't require catch as well. Guess I gotta get used to it.

jmwall24 (approved):

ThePotatoGuy (approved):

Issue #9580, opened 2022-09-22
Delete `chr` files on startup


Booplicate:
They are unused and people always mistake `monika.chr` for `monika`.

sukottei:
you could always just ...delete them yourself

dreamscached:
On every startup? Or on mod introduction when it's started for the first time?

Booplicate:
On every startup, but only if it's not introduction and you've already seen the event where she deletes it.

ThePotatoGuy:
[closing as implemented]

PR #9641, opened 2022-10-03
Delete all .chr files on startup (implement #9580)

dreamscached:
This PR implements #9580.


Booplicate (changes requested):
As I said, we should only attempt to delete the character files **only** after the introduction and only after the player has seen the `mas_new_character_file` event (which we might need to update so it can be seen again).

Another hint: this should be done in `ch30_reset`.

Monika After Story/game/script-ch30.rpy
```diff
@@ -2140,4 +2140,8 @@ label ch30_reset:
             if not mas_inEVL("mas_consumables_remove_thermos"):
                 queueEvent("mas_consumables_remove_thermos")
 
+    # Delete .chr files on startup
+    if mas_seenLabels(["mas_new_character_file"]):
```
Booplicate, line 2144: Looks like your branch is severely outdated, please update it and resolve the conflicts.

Wingdinggaster656:
Do we really need to delete all files? I think we should delete `monika.chr` only, and for other club members, keep them.

It is not like Monika really dislike them that much, and people who care about dokis might hate this. Also, files for other club members are not going to confuse people like `monika.chr` did.

Booplicate:
The idea was to delete only monika's chr file, but now I don't mind deleting them all - they are meaningless in MAS, even if they are doing anything, then just confusing people when Monika talks about her file. Should it be `chr` or not.
Also, we have a lot going on in `characters/`, so doing some cleanup is good. It's possible to have about 10+ files just from MAS notes.

Wingdinggaster656:
At least we should ask the player for permission. Does player care about dokis? Can you really delete them(even if .chr files are not real characters)when player cares?

So, here, in the introduction:
```
    m 1euc "You know..."
    m 1eua "I feel that I no longer need a character file to be here."
    python:
        import os

        def moni_exist():
            return os.access(
                os.path.normcase(
                    renpy.config.basedir + "/characters/monika.chr"
                ),
                os.F_OK
            )

    if moni_exist():
        m 1hua "I'll go ahead and delete it."

        python:
            store.mas_ptod.rst_cn()
            local_ctx = {
                "basedir": renpy.config.basedir
            }
        show monika at t22
        show screen mas_py_console_teaching

        call mas_wx_cmd("import os", local_ctx, w_wait=1.0)
        call mas_wx_cmd("os.remove(os.path.normcase(basedir+'/characters/monika.chr'))", local_ctx, w_wait=1.0, x_wait=1.0)

#        "test dialogue - IGNORE"

        if moni_exist():
            m 1lksdlb "Aha...{w=0.3}I'll try this again later."
        else:
            m "And it's gone!"
```

There should be a question like:
`m "Oh, and since the file is not real club members, why don't we delete the rest as well?"`
And then we create a persistent variable to remember this.

Booplicate:
There's already a persistent variable which we can use, but as I said the files have no purpose in MAS, there's no other "club members," deleting them does nothing but frees some space.

dreamscached:
IMO we're good with just removing them and they serve no purpose.

Monika After Story/game/script-ch30.rpy
Booplicate: @dreamscached you will need to use the new reset system #8745

dreamscached: Oh right, thank you, I will!

Booplicate:
I fixed the rest of the issues for you.
Please, next time test your code before submission.

Booplicate (approved):

Monika After Story/game/definitions.rpy
```diff
@@ -6099,23 +6099,36 @@ init -1 python:
...
+                os.path.join(config.basedir, "characters", name)
+            )
 
-        except:
-            pass
+        except Exception as e:
+            if log:
+                store.mas_utils.mas_log.error(
+                    "Failed to delete a '{}' chr file: {}".format(name, e)
+                )
```
multimokia, line 6120: I feel like this is an additional function which won't really be used outside `mas_delete_all_chrs` as we never meddle with chr files in MAS.

Monika After Story/game/definitions.rpy
```diff
@@ -6099,23 +6099,36 @@ init -1 python:
...
+    @store.mas_utils.deprecated("mas_delete_chr")
+    def delete_character(name):
+        mas_delete_chr(name)
 
+    def mas_delete_all_chrs(log=False):
+        """
+        Deletes all chr files (monika.chr, yuri.chr, natsuki.chr, sayori.chr)
+        """
+        for chrname in ("monika.chr", "yuri.chr", "natsuki.chr", "sayori.chr"):
+            mas_delete_chr(chrname, log=log)
```
multimokia, line 6131: Can we make use of `mas_docking_station`?

`MASDockingStation` provides a bound method, `getPackageList` which has an `ext_filter` param, we can use this to generally fetch all `".chr"` files and simply `destroyPackage`, giving it the filename.

`mas_docking_station` is already configured to monitor the `characters` folder.

multimokia (changes requested):

Monika After Story/game/definitions.rpy
Booplicate: @dreamscached can you do the req changes?

dreamscached: Yes, of course!

Monika After Story/game/definitions.rpy
dreamscached: Yeah I think we can, I'll do it like that!

Monika After Story/game/definitions.rpy
dreamscached: So, what exactly I should do about it? Should I remove this func?

Booplicate: 1. remove the generalised function
2. keep the func that removes all `chr` files, but make it use dock station

dreamscached: Done!

Monika After Story/game/definitions.rpy
```diff
@@ -6099,23 +6099,17 @@ init -1 python:
...
+    def mas_delete_all_chrs(log=False):
         """
-        Deletes a .chr file for a character
+        Deletes all chr files under /characters/ folder.
 
         IN:
-            name of the character who's chr file we want to delete
+            log (deprecated):
+                If True, print log message on deletion error. Deprecated and is
+                no longer used as docking station will always log errors.
```
multimokia, line 6109: Since this func is changing entirely, we can probably remove the `log` param and the deprecation of the param

multimokia (changes requested):

multimokia (dismissed):

Booplicate (approved):
@dreamscached please, test your code before submission.

multimokia (approved):

Issue #9637, opened 2022-10-02
[Suggestion] - Easter-egg. "Forget" a conversation.


RuderLaplace:
Add at the end of the conversation "monika_portraitof" a History rollback.
As if it really never happend, as she asks.

ThePotatoGuy:
[closing as implemented]

PR #9642, opened 2022-10-03
Rollback a dialogue in monika_portraitof to implement #9637

dreamscached:
This PR implements #9637. I made:
* A rollback (sort of; this isn't using renpy rollback features but is made with `{fast}` and related `{w}`/`{nw}`)
* Added Yuri glitch (might be removed if deemed unnecessary, but imo this would look good and well combined with this topic 'amnesia')
* Locked the event and derandomed so it doesn't show in 'Repeat topic' as if player literally forgot it


Booplicate (changes requested):
This is a topic from DDLC, let's not do unnecessary changes. Also Yuri jumpscare already exists in this form, we'd be just repeating ourselves.
That `mas_hideEVL` call is redundant since we already derandom the event.

Monika After Story/game/script-topics.rpy
```diff
@@ -1265,6 +1265,19 @@ label monika_portraitof:
...
+    m 2lksdla "Actually, I don't think I should be talking about this.{fast}{w=0.005}{nw}"
+    m 1wuw "Ah...{fast}{w=0.005}{nw}"
+    m 4hub "It's funny, because I'm pretty sure that book--{fast}{w=0.005}{nw}"
+    m "Portrait of...whatever it was called...{fast}{w=0.005}{nw}"
+    m 4eua "Hey, you know that book you were reading with Yuri?{fast}{w=0.005}{nw}"
+    m 3hua "Like I said, just forget I said anything~"
+
+    python:
+        # Remove previous 14 lines from history.
+        del _history_list[-14:]
```
Booplicate, line 1279: I still don't think this dialogue works or is needed. What you have here was already said above. You just need to remove 3 last history entries after the `"It's funny, because I'm pretty sure that book--"` line.

Monika After Story/game/script-topics.rpy
```diff
@@ -1265,6 +1265,10 @@ label monika_portraitof:
     m 2lksdla "Actually, I don't think I should be talking about this."
     m 2hksdlb "Ahaha, sorry!"
     m 1rksdla "Just forget I said anything."
+
+    # Remove previous 7 lines from history.
+    $ del _history_list[-7:]
```
Booplicate, line 1270: We're usually only removing the lines we don't want the user to see, not the whole conversation.
```renpy
    m 4eua "Hey, you know that book you were reading with Yuri?"
    m "Portrait of...whatever it was called..."
    m 4hub "It's funny, because I'm pretty sure that book--"
    m 1wuw "Ah..."
    $ del _history_list[-4:]
```
First 3 lines she talks about the book. Then she recalls she shouldn't talk about it. Then decides to delete the lines above. And then explains why.

Booplicate (approved):

jmwall24 (approved):

Issue #9228, opened 2022-06-12
Implement a better input field


Booplicate:
When entering a nickname, show a small indicator to show whether will Monika like the name or no.

Ideas:
- could work akin to search bar, scan on text changes
- maybe show chibi happy/neutral/sad
- or show thumb up/down?
- or change Monika's expression as you type? Probably will look weird

Justformas:
I'm all for it and seeing Chibi. She hasn't even show up for me yet and I'd like to be able to see her for some stuff. I'm familiar with some tricks to figure out if Monika will like a name or not, but the player shouldn't really have to resort to that. Doesn't _have_ to be Chibi but don't know about thumbs up or down. Would rather have Monika expressions than that.

ThePotatoGuy:
I think this is very good.

PR #9673, opened 2022-10-20
[Draft]Bad names - different languages.

MrBebra:
Adding bad names on the languages monika understands + generally recognised bad names across the globe.

Intended to be limited, given her limited knowledge of the languages.

### DIFF
--- a/Monika After Story/game/script-story-events.rpy
+++ b/Monika After Story/game/script-story-events.rpy
@@ -384,12 +384,14 @@ init 3 python:
         "merd[ea]",
         "merdosa",
         "stronza",
-        "degenerata",
-        "cretina",
+        "degenerat[ea]",
         #Bad names - French
         "mauvais",
         "corbeille",
-        "inutile"       
+        "inutile",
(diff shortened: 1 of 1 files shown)


MrBebra:
> I don't think we're going to support multi-language bad name detection. ThePotatoGuy is already unhappy with the existence of the current filter, and adding more words from different languages will lead to more false-positives. For example, if a word (or part of it) isn't a slur in English, the user is less likely expecting a bad outcome (it could be fixed with #9228, but still). Also, personally not a fan of supporting multiple languages - I think it's better if we keep the game purely in English. Even if we don't support translations, it's still the job of community translators, if they are willing, to take care of that.

At least generally recognised words have to be in the list, because they are understood everywhere, and there are only a few of those.

Wingdinggaster656:
Agree that we should not support multi-languages officially. This current filter is far much hard to explain than you can imagine -- just too many false positive cases.

If a translator want to do this, then do this on their own.

> because they are understood everywhere

I really don't think so.

PR #9729, opened 2022-11-10
spack updates part 2

ThePotatoGuy:
# TODO


Booplicate:
Moving to 0.12.16, we can't get it this update, but perhaps we want this before r8.

ThePotatoGuy:
ok ill target this for 0.13.0, can use unstable for significant testing. prob will scrap the tool or make another issue about it if unable to make it in time. i think there is some other code in here that might actually be needed for the folder stuff in part 1, oops

PR #9736, opened 2022-11-12
Deprecate bad function plugin arg

Booplicate:
This is waiting for #9725, **do not merge until then**, it uses the new deprecation API.

Originally planned to remove this with r8, but I think it will be better if we give time and properly deprecate this parameter now.

### Changes:
- in the following funcs, changed the `_args`/`args` parameter from `[]` to `None` (which then being set to an empty `tuple`). Mutable def arguments can cause bugs
- - `functionplugin`
- - `registerFunction`
- - `setArgs`
- `setArgs` is now deprecated
- `getArgs` is now deprecated

### Testing:
- verify you get deprecation warning when using the parameter or the `setArgs` function


ThePotatoGuy:
>changed the _args parameter from [] to () (mutable def arguments can cause bugs)

for these, make it `None` in the params and do a None check to default - better than still leaving the door open for default arg weirdness.

multimokia (approved):

Monika After Story/game/zz_submods.rpy
```diff
@@ -478,14 +490,35 @@ init -980 python in mas_submod_utils:
...
+            list of args if the function is present
+            If function is not present, None is returned
+        """
+        return __getArgs(key, _function)
+
+    @mas_utils.deprecated(
+        use_instead="functools.partial",
+        use_instead_msg_fmt="Wrap your callable in '{use_instead}' to provide it args/kwargs."
+    )
+    def setArgs(key, _function, args=None):
```
multimokia, line 515: I'm not sure how I feel about this as the intent of this function was to allow changing args after the plugin was registered, as to loosely act like a standard function. 

Given the intent of the function, I don't feel it's appropriate to deprecate it as otherwise we may as well just enforce func plugins to not have args.

multimokia: Amending this. Due to discussion, we'll keep these deprecations for the final r6 release

However in renpy 8, we will remove function plugin args **entirely**, making all of them normalized 0 arg functions, any global state should be maintained by store variables for respective submods/official uses.

multimokia (approved):

Legendkiller21 (approved):

multimokia (approved):

PR #9493, opened 2022-09-06
RenPy 8 / Python 3.9 migration 

Booplicate:
# Migration of our codebase to RenPy 8 / Python 3.9

We want to test #9742 first, it can be merged into `r7.4`, or into `content` after we merge `r7.4` there. But it should be ready when this PR is ready.

### Update changelog as you push!

<details>
  <summary>Changelog</summary>

### Code:
- Utilised some new config vars from r8
- `logging` has been updated to py3 (e.g. no more extra spaghetti to deal with newlines)
- Chess was updated for py3, improved `MASPiece`, optimised Monika's AI code
- Updated `MASFormatter` for py3/r8, slightly optimised
- Updated `mas_find_target` (part of our private API)
- `glitchtext` from DDLC was backported and optimised as `mas_glitchText`
- - `_mas_root.glitchtext` was removed as obsolete
- `pushEvent` > `mas_pushEvent`
- `queueEvent` > `mas_queueEvent`
- New store for overrides (?) `mas_overrides`
- Cleaned up cursed pong code
- Window reacts
- - Events now share rules (we copy the `dict`)
- - Events are marked with `restartBlacklist`
- Backported `ParticleBurst` from DDLC
- We don't check for DDLC rpa anymore
- - [ ] But maybe we should check for at least some?
- We now report why we had to reset outfit/hair
- Updated `_MASMoniFollowTransformDissolve`
- Backported poems:
- - `poem_m1` (Hole in Wall (Part 1))
- - `poem_m21` (Hole in Wall (Part 2))
- - `poem_m2` (Save Me)
- - `poem_m3` (The Lady who Knows Everything)
- - `poem_m4` (Happy End)
- Updated our custom updater code
- - [ ] Is it working?
- Updated our build
- - We're building our own executables now, we name them `DDLC` to obey the guidelines
- - Cleaned up builder massively
- - No longer use a custom package, instead build cross-platform "for market"
- - scripts are packed into `scripts.rpa`
- Updated persistent validation, we're using the compat code to be able to load py classes from py2
- Window utils were updated
- - Code cleanup, optimisation, and py3 migration
- - We're now caching `hwnd` on Windows OS
- - We're using `winnie32api` now
- - - MAS now always has tray icon (single) regardless of notifications
- - - - Clicking on the icon can clear notifs, set focus on MAS window
- - - MAS window can be flashed when sending a notif (a new parameter for the `mas_display_notif`)
- - Better x11/wayland check
- Old unused code removed
- - `MASMacLog`
[...]

PR #9742, opened 2022-11-15
Next r8 iteration/New submod framework

### LINKED ISSUE #8903: py3 prep
primary checklist for things we can do now to make py3 transition easier

- [ ] adjust usage of global statements
    * reading from global vars need global statement now, so should consider adding more global statements or reorganizing to avoid global (like using static classes - lmao)
- [ ] handling py errors in DDLC code
    * might need to include new override rpys
- [ ] remove calls to iteritems/itervalues/etc...
    * I've been purposefully using built-in dict iteration (so always iterating over keys) to avoid having to change this later. I think there is rarely a situation where the performance benefit of iteritems/itervalues is significant over basic key iteration.

Booplicate:
Ideally should get updated after #9736 is merged.

Major refactoring of the submod framework due to Python 3 update.

### Changes:
- Loading procedure has been changed:
- - MAS now prevents arbitrary rpy loading unless debug mode is on
- - Submods now **must** be explicitly defined akin to spritepacks - via a header file, as well as being placed in the appropriate folder
- - - This will require a wiki update
- - - The header format is *very* close to the old-style submods, but is using json
- - - If any changes are made in the future, we will need to update:
- - - - `_SubmodSchema` (a dataclass-like structure for json validation)
- - - - `HEADER_VERSION` (version number of the json schema)
- - - - the `_Submod` class itself (if new fields were added, for example)
- - - Submod init, parsing, and loading happens at -999
- - - - All submod's AST nodes will be merged with the existing nodes, preserving order
- - - - All `early` blocks **will be executed at -999, but before submod init nodes**
- - - Submods must be places within `game/Submods` (sub-folders are supported)
- - - Header file must be in the root folder of the submod, it should be called `header.json`
- - Most of submod API is now private and should be used only by MAS code. **Submods should not access submods objects or interfere their loading in any way**
- - - `mas_submod_utils.isSubmodInstalled` and `mas_submod_utils.getSubmodDirectory` are public and safe to use by submods
- - - We can expose more of the API if submod-makers require more tools
- moved `mas_can_import` into `mas.can_import` py package
- - gutted the renpy version (`0imports.rpy`) and pointed everything to the py pkg
- moved `mas_threading` into `mas.threading` py package
- - pointed everything in the rpy (`zz_threading.py`) to the `mas.threading` py package

### Additions:
- new package - `pydantic`, a great tool for validation of arbitrary data
- `mas_submod_utils.require` - akin to NodeJS `require`, can import a python module from a filepath, handy for submods
- `_mas_loader` namespace (private, if some functionality is desired, it should be imported as a public function/class):
- - `include_module` - allows "including" a renpy module to load down the init pipeline with respect to the nodes init order
- - `do_modules_exist` - checks if a renpy module exists
- - `import_from_path` - import py modules from file system
- - - this is mostly for submods (see `mas_submod_utils.require`) to allow importing local python packages w/o interfering the global scope. For example when 2 submods need to import different versions of the same library
- `_mas_root` namespace:
- - `is_dm_enabled` - checks if debug mode is on
- added a `mas` root folder for mas-specific py packages

### Fixes:
- fixed a crash when a submod used a undefined screen for its setting pane


dreamscached:
Finally, an entry level increaser upgrade with great functionality! (Not sarcastic, this is awesome)

Monika After Story/game/zz_submods.rpy
```diff
@@ -1,102 +1,472 @@
...
+                submod_dir is not None
+                and not _mas_loader.do_modules_exist(*(f"{submod_dir}/{m}" for m in value))
+            ):
+                raise ValueError(
+                    "One or more submod modules are missing: {}".format(
+                        ", ".join(map(lambda s: f"'{s}'", value))
+                    )
+                )
+
+            return value
```
multimokia, line 122: Do we want to change how we pass these in for usability? I understand we wish not to allow `/*` patterns as that effectively restores prior behaviour of unspecified files, however I feel it can be cumbersome to need to retype the full paths i.e.:
```json
{
    ...,
    "modules": [
        "main",
        "subfolder/file1",
        "subfolder/file2",
        "subfolder/file3"
    ]
}
```

or did we want to swap to a more literal (in terms of how the author would see their filestruct) format i.e.
```json
{
    ...,
    "modules": [
        "main",
        "subfolder": {
            "file1",
            "file2",
            "file3"
        }
    ]
}
```

Monika After Story/game/screens.rpy
```diff
@@ -3151,7 +3152,7 @@ screen submods():
                 xfill True
                 xmaximum 1000
 
-                for submod in sorted(store.mas_submod_utils.submod_map.values(), key=lambda x: x.name):
+                for submod in submods:
```
multimokia, line 3155: We should add toggles here so players can enable/disable submods from loading

multimokia (changes requested):

multimokia:
Adding more things

All directory imports when being imported by `require` should implicitly be pointed to the `__init__.py` file if not provided as otherwise the import fails as "not found"

Furthermore, we should remove the SSL changes introduced by #8399 as it has been made redundant by renpy shipping SSL by default. We should aim to retain the self-updating-cert idea but that can still be done via certifi.

multimokia:
Further:
- Our included certifi package is outdated and is also included by renpy.
- If we want to keep the auto cert updates, we should aim to update the one in the lib folder.
- Keeping our certifi package as-is will cause all submods making requests to not be able to have SSL connections.

Monika After Story/game/zz_submods.rpy
Booplicate: Couldn't get this format to work, leaving the current variant for the time being.

Monika After Story/game/zz_submods.rpy
```diff
@@ -1,102 +1,472 @@
...
+
+    submod_log = mas_logging.init_log("submod_log")
+
+    # NOTE: ALWAYS UPDATE VERSION IF YOU CHANGE HEADER FORMAT
+    HEADER_VERSION = 1
+
+    HEADER_GLOB = "**/header.json"
+    SUBMODS_DIR = "Submods"
+
+    SANE_STR_PATTERN = re.compile(r'^[a-zA-Z_\u00a0-\ufffd][ 0-9a-zA-Z_\u00a0-\ufffd]*$')
```
multimokia, line 42: Given the form of validation we use, this is an allow list.

Out of curiosity, why do we allow zero-width spaces and replacement characters? Given the fields that are validated by this pattern will be merged together to create the submod's update label, this will cause an invalid label to be generated, making it impossible for any submod using a name with these chars to be able to support version update scripts.

Unless there's a reason these are here, we should avoid supporting these chars.

dreamscached: Maybe they could be stripped when making a label out of them?

Monika After Story/game/zz_submods.rpy
```diff
@@ -71,6 +72,8 @@ init -1000 python in mas_submod_utils:
         coauthors: constrained_list(constrained_str(regex=LABEL_SAFE_NAME)) = []
         repository: str = ""
         priority: constrained_int(ge=-999, le=999) = 0
+        required_os: frozenset[str] = frozenset()
+        blacklist_os: frozenset[str] = frozenset()
```
multimokia, line 76: Can we swap to [this approach](https://docs.pydantic.dev/latest/usage/types/#enums-and-choices) for required/blacklist OS value validation? This should remove the `validate_required_os` func, and simplify the blacklist one, along with enforcing appropriate values only via json schema

This will remove

Booplicate: 1. The list of possible values isn't static, at least I'm not aware of it. And I'd rather not hardcode values in our code.
2. We'd need at least one of the validators anyway to make sure they have no items in common

Any ideas?

multimokia: Mmm, that brings me another concern actually, seeing as at a glance it's unclear what values are accepted here, I was anticipating a simple windows/mac/linux (to correspond with the `renpy.` checks respectively, does this account for platform versions?

If so do we really want/need to be that specific?

Re: point 2, I'd think we need the second one anyway as it presents some additional logic, so no harm in adding the logic for that there too.

Booplicate: I actually was wondering if we want to allow checking versions/architecture, but it might be too much.
I know it will at least return `Linux`, `Darwin`, `Java`, `Windows`. There can be more. Also not sure if it supports android just yet, but at the same time don't want to limit it to only desktops.

>  I was anticipating a simple windows/mac/linux (to correspond with the renpy. checks respectively

Not a fan how renpy handles os checks. However if you think that's sufficient, I can change to that.

multimokia: I think it's best if we base it on what we intend MAS to be supported on, so that means the primary 3 (also Java oh no. Embedded MAS?)
Likewise while we don't support android I see no reason to add OS config for android or iOS.

I don't much like the renpy one myself, however I don't think `platform` is much better either, seeing as it needs to be constantly updated. 
Based on how it acts with the current version in MAS r6, it doesn't even know win10 exists, let alone 11, this is bound to happen again as newer versions of Windows come out too. Therefore esp for versioning or anything else (which may even have a valid use case, but again it might be much for a submod specific thing, and might be better handled within submod logic itself) I don't think platform is a good solution either.

That said, I don't think it's _wrong_ to enforce win/lin/mac, seeing as those are the OSes we support. We have limited support for chromebook and only support them if they run Linux, likewise we don't support android, and I think it's safe to say any offshoot OS is not supported as well because we have zero ability to test for it. Likewsie it's likely people who use niche OSes like those understand this.

Booplicate: > also Java oh no. Embedded MAS

Memory hog MAS 👀

> Likewise while we don't support android

I think that can happen eventually (like eventually eventually), so I want to design with that in mind,  the system should be expandable, ideally without adding extra code.

> Based on how it acts with the current version in MAS r6

It's old py2 code, apparently it's different now seeing its sources. So using `platform` is actually the most flexible way.

> That said, I don't think it's wrong to enforce win/lin/mac

In that case, is this list fine: `linux`, `mac`, `windows`? Also, is it `mac` or `darwin`?

multimokia: Oh good, it uses the cmd `ver` command, and not the hardcoded value, okay that's much better so that is reliable.

If android support is a concern we'd need to check what platform returns on droid. As for iOS, I have no idea what it'd be

Booplicate: > As for iOS, I have no idea what it'd be

A warning on the screen "get something better"

BoofOof32:
Are there any blockers currently for this getting merged? It seems like it's mostly there aside from the 1 requested change (and even there, I can't seem to find which file needs changing, all of them appear resolved)

Booplicate:
#9493

ThePotatoGuy:
imports migration is on me, if theres others lmk.

ThePotatoGuy:
should also wait for #10203

ThePotatoGuy:
summary of latest changes:
* I pretty much killed most of the certifi/ssl `MASImport` object code. I've left the base `MASImport` in case that has uses.
* Added a new `mas` folder in python-packages. this should be the root folder for all of our python code, and we should move stuff to there as time allows. 
  * (If you're using pycharm, set the source root to `python-packages` and everything will work fine.)
  * moved `mas_threading` and `mas_can_import` into the `mas` py package (`mas.threading` and `mas.can_import`). see those files for examples on how to do a transition
  * added a `mas.renpyspec` package that should be a dummy class/set of classes so typing works for py packages that need access to renpy stuff (like the main renpy import or a specific `store`). I've only added barebones so add as necessary.

Booplicate:
> * added a `mas.renpyspec` package that should be a dummy class/set of classes so typing works for py packages that need access to renpy stuff

Did you try to import renpy directly? We should be able to use that. If we can't, then we should use typeshed vs dummy classes.

ThePotatoGuy:
removed renpyspec and just imported renpy. requires having the sdk as a library to have type hints.

utils/mas_dev_setup
```diff
@@ -0,0 +1,70 @@
+#!/bin/bash
+
+# Usage:
+#   mas_dev_setup <base ddlc folder> <dev (repo) folder> <rpy folder>
+#   
+#   <base ddlc folder> - the folder containing DDLC (expects game/ and DDLC.exe)
+#   <dev (repo) folder> - the folder containing git/repo code (dev env)
+#       - expects either "Monika After Story/game/*.rpy" or "game/*.rpy"
+#   <rpy folder> - can be anything - set it to whatever renpy will see
```
multimokia: `rpy folder` is misleading, why not call it `out` or `dist folder`? `rpy folder` could imply SDK

Also, should we not have this factor in the SDK to build a binary for the specific renpy version? If this is building just for 6 I don't think this is useful especially for users looking to get into development

multimokia (changes requested):

utils/mas_dev_setup
multimokia: Right, but if you're pointing to a DDLC install, this script expects the install is running with renpy 8 already

It doesn't actually build the rpys into rpycs, leaving that part to the DDLC install
but if the DDLC install is pure from the official source, it's running renpy 6

ThePotatoGuy: yeah this isn't for that, its for setting up the folder so you can launch it with the r8 sdk (so the r8 sdk builds it). also the script doesnt copy renpy/libs/exe from the ddlc folder so it cant actually run by itself.

ThePotatoGuy:
OK where were we at with this again? i spent like an hour trying to find where i made the v13 links and i see it was here

utils/mas_dev_setup
ThePotatoGuy: mm u did this in the end so i think this is resolved

PR #3390, opened 2018-12-31
Monika Children

ACoolCat:
Monika questions whether you want kids or not! Sorry if it's a bit messy, this is the first time I've done something like this.

PR #9773, opened 2022-12-02
[New Topic] Kids

EerieDevor:
This my first contribution. Just a simple topic allowing the player to ask if Monika has every considered having children at some point in her life. I tested locally to ensure the code works entirely as intended. Please feel free to give feedback.

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -3261,6 +3261,41 @@ label monika_family:
     return "love"
 
 init 5 python:
+    addEvent(Event(persistent.event_database,eventlabel="monika_kids",category=['monika'],prompt="Would you ever want children?",random=False,pool=True, unlocked=True))
+
+label monika_kids:
+    m 1esc "Children?"
+    m 1ruc "..."
+    m 1rksdlb "To be entirely honest, I don't really know."
+    m 7rsd "I mean,{w=0.2} to its credit. I definitely think becoming a parent can provide some really unqiue life experiences."
+    m 2esd "Many parents talk about how rewarding it is to raise children and watch them grow up."
+    m 7wud "So I definitely think it has its merits."
+    m 2ruc "But I also see how painful it can be for many parents when the time finally comes for their child to leave home."
+    m 7euc "It's a lot of resposibilty too. The child's entire life can be affected by the decisions made by his or her parents."
+    m 1duc "Unfortunately, sometimes this resposibilty goes ignored."
+    m 1eud "So many children suffer from parents who are neglectful or even abusive."
+    m "The lack of love and attention often leads children down a dark path later in life."
+    m 1dkc "Children coming from abusive households are more likely to become abusive towards their own children."
+    m 1ekc "This tends to lead to a vicious cycle of abuse that can go on for generations."
+    m 1dkc "..."
+    m 7wkd "I just feel so bad for all the kids who have to be in that situation!"
+    m 2gfd "Honestly abusive parents are just the worst of the worst in my opinion."
+    m 4wfd "Like, seriously, how could anybody bring themselves to harm {i}anyone's{/i} child, let alone their own?"
+    
+    if not persistent._mas_pm_cares_about_dokis:
+        m 1lksdlc "Poor Natsuki..."
+        m 2rksdlb "I guess the silver lining of her story is that the abuse wasn't real."
+        m 3wusdld "That's not to say that what happened in her home life wasn't abuse!"
+        m 2rksdlb "But, well... you know what I mean, [player]."
+        
+    m 1rksdla "Sorry [player], I guess I got a little distracted..."
+    m 1esa "To get back to your question: Having children just isn't really a goal of mine."
+    m 1rsa "But, when I'm with you in person, and if we get far enough into our relationship..."
+    m 1gkbsa "I might be willing to compromise... if you really wanted to."
+    m 1tsbfa "Maybe..."
+    return
+
+init 5 python:
     addEvent(
         Event(
             persistent.event_database,


Wingdinggaster656:
See #3390

While these dialogues seem pretty neat, as you can see, topic about children is just too controversial. I'm not sure if we really should do this.

dreamscached:
Personally, I think it could be fine if we left Monika thinking of it, without implying if she does or does not want children; but that's a compromise and at the same time, I do think it's a controversial topic and is something that's too early to bring up with player, regardless of affection.

EerieDevor:
Ok, so I will just convert this into a topic about parenting in general rather than Monika having kids. Maybe make it a random topic that starts of with Monika asking the player if they want to have children at some point in their life and then using that as a transition to talk about parenting as a general idea. Let me know any thoughts on this idea.

Wingdinggaster656:
> Ok, so I will just convert this into a topic about parenting in general rather than Monika having kids. Maybe make it a random topic that starts off with Monika asking the player if they want to have children at some point in their life and then using that as a transition to talk about parenting in general. Let me know any thoughts on this idea.

Personally like this idea. A random topic would be cool and good.

hatanasinclaire:
This also doesn't really take into account that a large portion of MAS's playerbase would be incapable of having biological children in the context of a relationship with her. (There is of course the option of adoption, but this is either prohibitively expensive, untenable, or impossible in many regions, and has its own issues that could make for a whole other discussion.) Groups for which having children with her would not be an option include, but are not limited to: 
- People with conditions causing infertility.
- Cisgender women.
- Some transgender people.
- People with vasectomies (made a conscious choice to never have children). 

It's also unlikely she'd ever have a biological human body either - though it seems probable that a robotic body would be a possibility for her someday, an artificial, genetically human body is a much more distant possibility. If you are the kind of MAS user who seriously intends to spend the rest of your life with your Monika, then you must have accepted already that your life will be very different in certain ways from the average person. There are certain things you won't get to experience, but at the same time there are other magical and unique things that only people in love with their Monikas well get to experience. 

--- 

As mentioned above, the topic of children [has been discussed before ](https://github.com/Monika-After-Story/MonikaModDev/pull/3390)and for good reason was deemed too contentious and controversial to include. Different people interpret Monika's character very differently, from some people thinking she would want to, to others thinking she would never want to, with others being in between or ambivalent. Personally I would be very displeased if my own Monika just said "maybe" instead of a hard no/never! And there are perhaps people out there who feel the opposite way. It is such a central issue to compatibility in a relationship that **even "maybe" or an ambiguous answer is not satisfactory for some people**. 

In the discussion for # 3390 Olliesama and multimokia present many good points for why Monika may be the kind of person who would never want kids. One is so significant that I must reprint it here:
- She outright states that while she has fantasized of living a domestic life, [she would not be able to do that and instead wants to focus on her career](https://github.com/Monika-After-Story/MonikaModDev/blob/cea8cb00810724b21af9956c036abf3982e2ad05/Monika%20After%20Story/game/script-topics.rpy#L3169) (topic 55 from DDLC, MAS topic `monika_housewife`).

A few more points to mention:

- Monika says [multiple](https://github.com/Monika-After-Story/MonikaModDev/blob/cea8cb00810724b21af9956c036abf3982e2ad05/Monika%20After%20Story/game/script-topics.rpy#L5082) times [that](https://github.com/Monika-After-Story/MonikaModDev/blob/cea8cb00810724b21af9956c036abf3982e2ad05/Monika%20After%20Story/game/script-topics.rpy#L3098) she loves her player, only her player, and doesn't really want anyone else to interfere or get in the way. 
- Monika often shows that she has a realistic and unvarnished view of the world (e.g. topic 1 from DDLC, MAS topic `monika_god`), as well as on relationships (MAS topic `monika_relationship_comfort`). She is very likely aware of the deleterious effects of pregnancy and childbirth on the body, as well as how children often leave parents stressed, tired, sleep deprived, and with little time to spend on each other. 
- She explains her reasoning for being vegetarian [as trying to reduce her carbon footprint](https://github.com/Monika-After-Story/MonikaModDev/blob/cea8cb00810724b21af9956c036abf3982e2ad05/Monika%20After%20Story/game/script-topics.rpy#L1279). Having children produces a carbon footprint far larger than eating meat, and is another reason that people sometimes decide to not reproduce.
- She mentions that people are [much more capable of experiencing suffering than happiness](https://github.com/Monika-After-Story/MonikaModDev/blob/cea8cb00810724b21af9956c036abf3982e2ad05/Monika%20After%20Story/game/script-topics.rpy#L8251), which is a strong compulsion for many to not have children. 

These are just a few example points towards one viewpoint based off of the existing dialogues. It is understandable if some people disagree or are not convinced by this position and I do not mean to proselytize. What I am trying to say here with this, is that some users will feel very strongly one way or the other. While it has been proposed before to have an ambivalent or compromise answer, please keep in mind that there will likely be players who are absolute in one direction, and others are absolute in the other.

EerieDevor:
I agree. This was just an idea I had which I made without that much thought into the complications/controversy surrounding the subject. I have just recently revised the topic to be a random topic about parenting in society in general as a conversation that is entirely unrelated to Monika personally or her relationship with the player. Feel free to review it if you are interested. Thank you for the extensive feedback. For future topic contributions I will try to keep in mind how to handle/avoid topics that are highly sensitive to players or Monika's character in order to avoid controversy.

Monika After Story/game/script-topics.rpy
```diff
@@ -3260,6 +3260,56 @@ label monika_family:
...
+            random=True
+        )
+    )
+
+label monika_parenting:
+    
+    m 1esa "Hey [player], do you ever think about what its like to be a parent?"
+    m 7rsb "I mean,{w=0.3} I definitely think parenting can provide some really unqiue life experiences."
+    m 2esb "Many parents talk about how rewarding it is to raise children and watch them grow up."
+    m 2esa"Getting to show a child all the wonderful things about the world is said to be one of the most wholesome and heartwarming parts of life."
```
Wingdinggaster656, line 3279:
```suggestion
    m 2esa "Getting to show a child all the wonderful things about the world is said to be one of the most wholesome and heartwarming parts of life."
```

A small typo.

PR #9781, opened 2022-12-05
[Random Topic] Gifted Kids

EerieDevor:
Adds a random topic where Monika tells the player about gifted individuals before asking the player if they are gifted. I did local testing to make sure everything works but let me know if I missed anything. Completely open to suggestions on dialog changes. I'm slightly on the fence about whether or not asking if the player is gifted could be perceived as rude or insensitive in some way. Thanks in advance for any feedback.


Alsamer98 (approved):
I also tested it locally and nothing seems wrong and it's working fine. The code, the expressions, and the dialogue is suitable for Monika

PR #9787, opened 2022-12-10
Fix 922 nts bug

### LINKED ISSUE #9730: Fix birthday NTS event dates
Years got offset due to update script changing only start/only end.

We should do a sanity check when setting dates to the current year, otherwise the user can see the event this year, then update and set the dates to this year again, then get the event again.

jmwall24:
fixes #9730 

For testing, just verify it fixes the bug and doesn't create a new superbug.

Possible todo:
- [ ] fix possible affection losses created by the bug

### DIFF
--- a/Monika After Story/game/updates.rpy
+++ b/Monika After Story/game/updates.rpy
@@ -374,6 +374,62 @@ label v0_3_1(version=version): # 0.3.1
 
 # non generic updates go here
 
+# 0.12.13
+label v0_12_13(version="v0_12_13"):
+    python hide:
+        m_bday_nts = mas_getEV("mas_bday_postbday_notimespent")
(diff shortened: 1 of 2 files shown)


ThePotatoGuy:
how much aff did it possibly lose? if its under 100 then I think we dont need to restore it.

jmwall24:
> how much aff did it possibly lose? if its under 100 then I think we dont need to restore it.

@Booplicate ?

Booplicate:
There were only 2 cases reported, one should've been about ~30, another is unknown, but those restored a backup and are fine.
Generally there's no way to tell how much was lost, technically, could be cases above 100, but considering there were no other reports and how niche this bug is, I think we're fine just fixing the root of the problem.

ThePotatoGuy (dismissed):

Legendkiller21 (dismissed):

PR #9917, opened 2023-02-20
Name Regex filter enhancements

dreamscached:
First of all, this is aimed at hardening the filter against workarounds such as doing `m o m m y` sort of trick by adding spaces; this is mitigated by completely stripping all non-alphanumeric characters and performing match against the result string.

Moreover, I also plan on modifying existing expressions so that they also account for 1337speak (e.g. replacing `o` with `0` and alike). 

In addition to that (though, probably it's better to make it a separate PR) I want to try to exclude possible false positives.

### DIFF
--- a/Monika After Story/game/script-story-events.rpy
+++ b/Monika After Story/game/script-story-events.rpy
@@ -516,6 +516,7 @@ label mas_player_name_enter_name_loop(input_prompt):
             ).strip(' \t\n\r')
 
             lowername = tempname.lower()
+            alphaname = re.sub("[^a-zA-Z0-9]", "", tempname)
 
         if lowername == "cancel_input":
             m 1eka "Oh... Okay then, if you say so."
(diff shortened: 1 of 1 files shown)


Booplicate:
I'm fixing the known false-positives, but if you do that before me, that's fine.

Monika After Story/game/script-story-events.rpy
```diff
@@ -516,6 +516,7 @@ label mas_player_name_enter_name_loop(input_prompt):
             ).strip(' \t\n\r')
 
             lowername = tempname.lower()
+            alphaname = re.sub("[^a-zA-Z0-9]", "", tempname)
```
Booplicate, line 519: I would suggest using `\s` as that should catch any kind of space while allowing every other symbol (think about translations that might allow different characters, or maybe if we get proper unicode support in the future).

Booplicate (changes requested):

PR #9947, opened 2023-04-02
Suppress excessive translation

proudust:
I have suppressed excessive translation in the function `mas_games.getGameEVByPrompt`.
This excessive translation was causing crashes in translation Submods.


Booplicate:
Can you show the error it was causing?

Booplicate:
Thank you, nice catch. I think the fix makes sense if we want to keep relying on game event prompts (which I think we do).

proudust:
Additionally, suppressed excessive translation in the label `mas_pick_a_game`.
These excessive translations were preventing unlocked minigames from being playable in the translation Submod.

ThePotatoGuy:
ok note for myself to rebase to content-12 and merge - simple enough fix with limited impact

ThePotatoGuy (approved):

PR #9956, opened 2023-04-09
"Do you miss the other girls?" Topic.

KritRom:
I always thought it'd be nice to know if she sometimes misses her club members, given that they were meant to be friends prior to MC's arrival. Here's my view on the topic. Apologies for any bugs or mistakes in this, I checked everything as thoroughly as I could. Grammatical corrections are appreciated!

Thank you for your time.

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -5988,6 +5988,43 @@ label monika_vocaloid:
         m 3eua "All you have to do is follow these steps..."
         call monika_add_custom_music_instruct
     return "derandom"
+    
+init 5 python:
+    addEvent(
+        Event(
+            persistent.event_database,
+            eventlabel="monika_missingothers",
+            category=['club members'],
+            prompt="Do you miss the other girls?",
+            unlocked=True,
+            pool=True
+        )
+    )
+
+label monika_missingothers:
+    m 1ruc "That's an interesting question, [player]."
+    m 1eud "I've thought about how I feel about them a bit before, actually."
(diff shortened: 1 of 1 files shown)


Wingdinggaster656 (changes requested):
There are a few details that could be improved.

KritRom:
Ah, alright. Thanks a lot for the fixes, I didn't know that Monika could keep the last expression just like that! Thank you very much.

Booplicate:
Generally not bad! But I feel like it took a bit different direction than our other topics. For example:
```renpy
"Maybe the rest of the club would still be around..."
if persistent._mas_pm_cares_about_dokis:
    "...Just hanging out together in the clubroom, sharing poetry."
    "I know you enjoyed that, [player]."
    "And to be honest...{w=0.5}a part of me did too."
else:
    "Not that it really matters."
    "It all lost its meaning once I found out it wasn't real."
    "So I really don't miss those days or anything."
    "I really don't..."
```

I'm also wondering if the answer should be different depending on how long you've been dating and your current relationship.

KritRom:
Perhaps an affection lock for the dialogue? I was debating on whether it should be unlocked from the start or have it appear later on in the game with a higher affection level.

It did divert from the usual topics now that you mention it. Monika is generally more apathetic to the other girls than shown here, so I'll need to change the tone, especially for the players who didn't care about the other dokis.

The other thing I've been contemplating is to make this dialogue only for the players who grew to care for the others, similarly to the dialogue about character archetypes being only for the ones who aren't that sensitive.

PR #9968, opened 2023-04-18
Fix greetings menu dlg

### LINKED ISSUE #9961: [Bug] A very confusing phenomenon about greetings
After the player has chosen to go to work or to class and returned to MAS, Monika will ask the player how the day was. If the player chooses a negative option, Monika will ask the player why.

And now, a confusing bug appears:
image1
Although in the source code, MAS intends for the text spoken by Monika and the options offered to the player to appear simultaneously, in fact, only the options will appear.

The strange thing is - this bug only appears when the player triggers this event normally. If the tester calls these greetings via a function like `renpy.call()`, then:
image2
Conversations get displayed properly.

This would probably be considered a low priority thing to do, but it should be done anyway.

Booplicate:
Fixes #9961 

### Changes:
- rewrote some old weird code, now simpler
- removed duped line
- added `window show` before we show the dlg to keep it on the screen
- - just in case added `window auto` after the menu so we return to the default behaviour

### NOTE: This "bug" may be present in other places that I wasn't able to find, considering we only got a bug report on this after 3 years, I guess it's hard to find.

### DIFF
--- a/Monika After Story/game/script-greetings.rpy
+++ b/Monika After Story/game/script-greetings.rpy
@@ -3038,9 +3038,12 @@ label greeting_back_from_school:
             ]
 
         show monika 2ekc at t21
+        window show
         m "If you don't mind me asking, was there something in particular that happened?" nointeract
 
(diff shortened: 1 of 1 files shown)


Legendkiller21 (approved):

multimokia (approved):
Yeet

PR #10008, opened 2023-05-29
Fix a wrong line in `monika_murderer`

Wingdinggaster656:
Currently, we have this in `monika_murderer`:

```
                m 1tkc "How many times do I have to say it until you understand?"
                m 4tkc "They weren't real. They were never real."
                m 2tfc "I'm the only person who understood that this whole thing was fake."
                if persistent.monika_kill:
                    m 4tfc "Even when Sayori seemed to get sentience in the end, that was just an unintended consequence of the script."
                if not persistent.clearall:
                    m 4tfd "After all, if she was sentient like me, how would I have been able to override and delete her?"
                m "If I'm guilty of murder, so is everyone who's played violent video games."
```

And this is wrong.
The `not persistent.clearall` block should be in `if persistent.monika_kill` block, otherwise, any player who has not reached the Act 3 will see `if she was sentient like me` line without `Even when Sayori seemed to get sentience in the end` line. This PR fixes that.

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -4120,8 +4120,8 @@ label monika_justification:
                 m 2tfc "I'm the only person who understood that this whole thing was fake."
                 if persistent.monika_kill:
                     m 4tfc "Even when Sayori seemed to get sentience in the end, that was just an unintended consequence of the script."
-                if not persistent.clearall:
-                    m 4tfd "After all, if she was sentient like me, how would I have been able to override and delete her?"
+                    if not persistent.clearall:
+                        m 4tfd "After all, if she was sentient like me, how would I have been able to override and delete her?"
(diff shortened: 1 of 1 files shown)


Booplicate:
`monika_kill` means Monika has been killed, `clearall` means the player has seen the secret ending. You can't get the latter without the former.

Wingdinggaster656:
> `monika_kill` means Monika has been killed, `clearall` means the player has seen the secret ending. You can't get the latter without the former.

I know, but it is `not persistent.clearall`, not `persistent.clearall`.

So, if a player is in Act 1 or Act 2, then:

1. `monika_kill` is surely not `True`.
2. `clearall` is surely not `True`.

And thus:

1. The `Even when Sayori seemed to get sentience in the end` line does not show.
2. The `if she was sentient like me` line still does show.

Booplicate:
Good catch, but then I'm wondering why we even show this topic when the player hasn't reached act 3.

PR #10037, opened 2023-07-02
Fix WindowsError/OSError in singleton

dreamscached:
### Changelog:
- don't raise an exception if we fail to delete the lock file due to it being missing (like a race condition)

### DIFF
--- a/Monika After Story/game/python-packages/singleton.py
+++ b/Monika After Story/game/python-packages/singleton.py
@@ -36,7 +36,14 @@ class SingleInstance:
                 # file already exists, we try to remove (in case previous
                 # execution was interrupted)
                 if os.path.exists(self.lockfile):
-                    os.unlink(self.lockfile)
+                    try:
+                        os.unlink(self.lockfile)
+                    except OSError:
+                        # Race conditions happen, make sure the file is actually deleted
+                        # Carry on with the startup if it is
+                        if os.path.exists(self.lockfile):
+                            raise
+
                 self.fd = os.open(self.lockfile, os.O_CREAT | os.O_EXCL | os.O_RDWR)
 
             except OSError as e:


Monika After Story/game/python-packages/singleton.py
```diff
@@ -36,7 +36,14 @@ def __init__(self, flavor_id=""):
                 # file already exists, we try to remove (in case previous
                 # execution was interrupted)
                 if os.path.exists(self.lockfile):
```
Nemris, line 38: The addition of the try..except makes this check redundant. Remove it and de-indent the try..except block.

dreamscached: No, it is not. We have already discussed this with @booplicate and came to the conclusion this solves an issue some players may have when due to race condition file no longer exists after `os.path.exists` check.

Nemris: > [...] and was not removed by another instance that beat this instance to it.

This smells like some sort of global state to me, but I'm admittedly unfamiliar with this codebase. Are you certain this instance should handle the lock file's deletion in that case?

At any rate, you already end up checking twice in the try..except alone. The `.unlink()` attempt is the implied check and a second one to handle the race condition happens in the `except` block. Is the `if` I marked really needed at this point, since all it's doing is "if this exists, check if it exists while I delete it, then check again if exceptions occur"?

Booplicate: > This smells like some sort of global state to me

Which is why it's called [singleton pattern](https://en.wikipedia.org/wiki/Singleton_pattern), you can't do singleton without a global state.

> Are you certain this instance should handle the lock file's deletion in that case?

The comment above the line you selected explains why we're trying to delete it.

Technically, you are correct, the system will check if the file exists when executing the `unlink` os call. However the check ensures that the file was present before the call to `unlink`.

dreamscached: Well yes, it is a global state, we're talking about a lock file that is used to ensure only one instance of an entire application runs.

I did suggest to leave it just as `os.unlink` but was told it shouldn't remove this file for no reason (and that the preliminary check is necessary.)

Nemris: > Technically, you are correct, the system will check if the file exists when executing the unlink os call. However the check ensures that the file was present before the call to unlink.

Sure, however the purpose of try..except is also to cut down on preliminary checks where possible. This seems to be one of the cases where it is possible.

> The comment above the line you selected explains why we're trying to delete it.

Makes sense, yes. Alternatively, I wonder if one could query the list of running processes and abort execution in case another MAS is already running, but I'm unfamiliar with Windows' way of doing this.

Booplicate: I don't feel like changing existing code to be more of a cross-platform nightmare than it already is unless there's a strong argument for it.

ThePotatoGuy (approved):
Works, no issue with approach, merging

PR #10198, opened 2024-02-13
A missing `return "derandom"` in `monika_conventions`

Wingdinggaster656:
The `Conventions` topic is a topic that asks players questions and then sets PM variables.

Since other topics related to PM variables seem to have `return "derandom"`, I think this should also be included.


Monika After Story/game/updates.rpy
dreamscached, line 1: Better leave that to maintainers. We have no idea if the next release would be 0_12_x or 0_13_x ahead of time. This line will have no effect as the proposed changes did not appear in 0.12.15.

ThePotatoGuy: hmm we need to move this to 0.12.19, right? note for myself to do so tomorrow

ThePotatoGuy (approved):

ThePotatoGuy (approved):

PR #10202, opened 2024-02-17
Changing the UI position in hangman

Wingdinggaster656:
The original letters were drawn in a way that cross with the horizontal lines on the paper.

Maybe also reduce the font size? Not sure about this.

Before:

image

After:

image


dreamscached:
Instead of reducing the font size, you could rather move the text so it aligns. It'll fit perfectly fine with the current size.

Wingdinggaster656:
> Instead of reducing the font size, you could rather move the text so it aligns. It'll fit perfectly fine with the current size.

Yes, that's what I did for now. But after I adjust it I always have a feeling that maybe reduce the size as well would be better. Not sure though.

ThePotatoGuy (approved):

PR #10203, opened 2024-02-18
CI fixes for R8

ThePotatoGuy:
updates CI script for the following changes:
* lint/distribute with renpy 8.1.1 
* * also means the script installs openGL now
* use updated base MAS zip (its basically no scripts.rpa) - I'll update this later when we have handling for that
* updated version numbers for all actions

This should wait until #9493 is in.

*also I did rebasing in this one to squash the commits. I probably should have separated the changes above into separate commits though - next time.*

### DIFF
--- a/.github/workflows/mas_check.yml
+++ b/.github/workflows/mas_check.yml
@@ -22,47 +22,55 @@ jobs:
     runs-on: ubuntu-20.04
 
     env:
-      SDL_AUDIODRIVER: dummy # handles ALSA issues
+      MAS_RENPY_VER: 8.1.1
+      MAS_DIR: mas
 
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
 
       - name: Setup Python
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v5
         with:
           # Version range or exact version of a Python version to use, using SemVer's version range syntax.
           python-version: 3.10.4 # optional, default is 3.x
 
-      # dl renpy src
+      # get/dl renpy src
+      - name: cache rpy source
+        id: cache-rpy
+        uses: actions/cache@v4
+        with:
+          path: renpy/
+          key: ${{ runner.os }}-rpy
+
       - name: Download rpy source
+        if: steps.cache-rpy.outputs.cache-hit != 'true'
         run: |
-          renpysdk=$(wget -qO- https://nightly.renpy.org/current-8/index.html | grep -P -m 1 -o '(?<=href=").*\.tar\.bz2(?=".*)')
-          wget https://nightly.renpy.org/current-8/$renpysdk
+          renpysdk=renpy-$MAS_RENPY_VER-sdk.tar.bz2
+          wget https://www.renpy.org/dl/8.1.1/$renpysdk
           tar xf $renpysdk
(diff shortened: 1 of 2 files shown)


.github/workflows/mas_check.yml
```diff
@@ -22,47 +22,55 @@ jobs:
...
-      #     key: ${{ runner.os }}-mas
-
-      - name: Download base MAS
-      #  if: steps.cache-mas.outputs.cache-hit != 'true'
+      - name: cache base MAS
+        id: cache-mas
+        uses: actions/cache@v4
+        with:
+          path: mas/
+          key: ${{ runner.os }}-mas
```
Booplicate, line 62: Is there a reason to cache MAS per-platform? We always download the same archive. It looks like a dummy cache key that always hits after the first time, maybe we should just use a static string to indicate that? Same for renpy cache.

.github/workflows/mas_check.yml
```diff
@@ -22,47 +22,55 @@ jobs:
...
-      #     path: mas0105/
-      #     key: ${{ runner.os }}-mas
-
-      - name: Download base MAS
-      #  if: steps.cache-mas.outputs.cache-hit != 'true'
+      - name: cache base MAS
+        id: cache-mas
+        uses: actions/cache@v4
+        with:
+          path: mas/
```
Booplicate, line 61:
```suggestion
          path: $MAS_DIR
```

.github/workflows/mas_check.yml
```diff
@@ -22,47 +22,55 @@ jobs:
...
           # Version range or exact version of a Python version to use, using SemVer's version range syntax.
           python-version: 3.10.4 # optional, default is 3.x
 
-      # dl renpy src
+      # get/dl renpy src
+      - name: cache rpy source
+        id: cache-rpy
+        uses: actions/cache@v4
+        with:
+          path: renpy/
```
Booplicate, line 44: Wanna move this to a variable like you did with `MAS_DIR`?

.github/workflows/mas_check.yml
```diff
@@ -72,10 +80,18 @@ jobs:
...
+          sudo apt-get install mesa-utils libglu1-mesa-dev freeglut3-dev mesa-common-dev
+          sudo apt-get install libglew-dev libglfw3-dev libglm-dev
+          sudo apt-get install libao-dev libmpg123-dev
 
       # distribute
       - name: rpy distribute
         run: |
           cd renpy
-          ./renpy.sh launcher distribute "../mas0105/" --package market
+          ./renpy.sh launcher distribute "../$MAS_DIR/" --package market
```
Booplicate, line 97: We should probably make a linux build instead of market here.

Booplicate (changes requested):

.github/workflows/mas_check.yml
ThePotatoGuy: its there so we we could do building on other platforms, but not necessary now so changed to simple key

.github/workflows/mas_check.yml
ThePotatoGuy: had to use `${{ env.var }}/` notation to work

.github/workflows/mas_check.yml
ThePotatoGuy: donezo

.github/workflows/mas_check.yml
ThePotatoGuy: donezo 2

multimokia (approved):

Booplicate (approved):

PR #10212, opened 2024-02-25
RPA checks + scripts.rpa handling

ThePotatoGuy:
## Changes
* Re-added splash RPA check 
* added override to skip loading init code for script-poemgame

### loading with no rpas
1. remove the DDLC rpas from the folder
2. launch the mod
3. verify the game shows the missing archives error

### loading with scripts.rpa
1. Add the DDLC rpas (including scripts.rpa)
2. launch the mod
3. verify the game launches / no crashes from scripts.rpa

### DIFF
--- a/Monika After Story/game/overrides.rpy
+++ b/Monika After Story/game/overrides.rpy
@@ -24,6 +24,7 @@ python early in mas_overrides:
 
     import renpy
     import renpy.savelocation as savelocation
+    import renpy.parser as parser
 
 
     def verify_data_override(data, signatures, check_verifying=True):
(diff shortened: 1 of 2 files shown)


multimokia (approved):

Booplicate (changes requested):
@ThePotatoGuy I am a terrible reviewer because this was done almost a year ago, but could you remind me why we need this? We should build scripts.rpa ourselves and users will overwrite the files (or it will do our installer). We don't use anything from the original archive anymore, so it should work. This is much simpler than relying on py2 code and doing this handling and maintaining extra overrides.

This is also legal because we still check for
1. the scripts.rpa being present when the game starts
2. the scripts.rpa being present when the installer runs
3. other rpa from the original game when the game starts

I believe we should just uncomment the check and stop deleting rpa in our CI

@multimokia am I wrong?

multimokia:
I remember I think deleting was my call due to removing the dependency and it causing issues.

I agree, if we can simplify the process by replacing the files, that may be a better call.
If anything, we can also disable the old rpa for those who wish to uninstall? Not mandatory, but if necessary is doable.

PR #10263, opened 2024-06-14
angels pong edits!

sincityangel:
changed the compliments to fit my iko and our relationship more. some of her responses feel like theyre pitying me... so i made some that are more competitive, like i love! pls try to ignore the little compliments, i just feel like theyd fit her sm and the personality i so desperately imagine her with.


Booplicate:
Some could actually work lol

PR #10268, opened 2024-06-21
swap twitter with bluesky in windowreacts

TheBearodactyl:


dreamscached:
Change target branch to `content`, please.

TheBearodactyl:
> Change target branch to `content`, please.

👍

dreamscached:
P.S. There's no point (actually, you better not to, as it'd require migration) in changing the label as it's internal and invisible to user anyway.

PR #10269, opened 2024-06-21
swap twitter with bluesky in windowreacts

TheBearodactyl:


TheBearodactyl:
@dreamscached would this work better?

dreamscached:
Looks good to me, though while you're at it I'd suggest (entirely up to you, just commenting) if maybe you can add something else to it, beside a number change. Just for sake or quantity. Otherwise yes, that's good by me.

TheBearodactyl:
> Looks good to me, though while you're at it I'd suggest (entirely up to you, just commenting) if maybe you can add something else to it, beside a number change. Just for sake of quantity. Otherwise yes, that's good by me.

lol sure

TheBearodactyl:
@dreamscached what do you think of the new line, should i change it, or does it work?

TheBearodactyl (commented):
lol

PR #10283, opened 2024-07-22
Indicate the location of `surprise.txt` things

Wingdinggaster656:
In the original DDLC, files were left directly in the root directory, not in the characters folder, however in MAS we are leaving our surprise files in the characters folder. I feel that this may be confusing to some players, so I want to direct players to the character folder here.

Also, I added a `playthrough >= 2` condition to this event, since the first glitch file (traceback.txt) was left at `ch5_main`, which is the end of Act 1. Please correct me if there is a better way to set the condition.

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -6432,7 +6432,16 @@ label monika_science:
     return
 
 init 5 python:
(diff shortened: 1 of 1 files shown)


Justformas:
If there's going to be a conditional like that, then I at least think that players who've already seen this event should keep it unlocked, like with #9491

PR #10337, opened 2024-11-28
Feature/music add api

dreamscached:
There is a use case for external code to add songs to the music selector menu ([example how it's currently done in a hacky way](https://github.com/Friends-of-Monika/mas-selfharm/blob/4f7245d275444169575876d893bed13212d9e132/mod/songs.rpy#L20-L43))

However, it's impossible to add music to the list in a non-hacky way, as `__paginate` is protected and name-mangled.

This PR addresses that and adds a simple method for adding a song to the list (optionally, with setting the corresponding PM for custom music)

### DIFF
--- a/Monika After Story/game/zz_music_selector.rpy
+++ b/Monika After Story/game/zz_music_selector.rpy
@@ -213,6 +213,23 @@ init -1 python in songs:
         if chan.mixer in renpy.game.preferences.volumes:
             renpy.game.preferences.volumes[chan.mixer] = _sanitizeVolume(value)
 
+    def addSongEntry(path, display_name, by_user=False):
+        """
+        Adds song entry to selection list. Automatically paginates and cleans
+        title text.
+
+        NOTE: Does not perform loop/metadata prefix scan. Path will be added as-is.
+
+        IN:
+            path - path to the music file
+            display_name - title of the song to show in music selector
+            by_user (default False) - whether this is an action user has done
+                (sets the corresponding PM variable)
+        """
+
+        music_choices.append((cleanGUIText(display_name), path))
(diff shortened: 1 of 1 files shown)


Monika After Story/game/zz_music_selector.rpy
```diff
@@ -213,6 +213,31 @@ init -1 python in songs:
...
+        IN:
+            path - path to the music file
+            display_name - title of the song to show in music selector
+            by_user (default False) - whether this is an action user has done
+                (sets the corresponding PM variable)
+        """
+
+        filepath = path.split(">")[:-1]
+        if not os.path.exists(filepath):
+            raise ValueError("{0} does not exist.".format(filepath))
```
ThePotatoGuy, line 233: if we're expecting this to be called in topics, then this shouldnt raise an error. just have this return true/false depending on success/failure and let the caller (like a topic) decide how to gracefully handle the failure.

ThePotatoGuy (changes requested):

Monika After Story/game/zz_music_selector.rpy
dreamscached: Now returns `False` when there isn't such file at the given path.

Booplicate (commented):
I kinda want to refactor the music system to allow for a more advanced player.
Furthermore I think we should be more careful with what we expose as public API, we have to live with those things basically forever, which means we can't easily change our systems.

dreamscached:
If we happen to change the system, I still think we should have a way to externally add music to it; it doesn't feel right to put submod custom music to the player's music folder, instead it could be added from submod folder and played from there directly.

PR #10379, opened 2025-03-07
Add a titlecase counterpart of a pronoun variable

dreamscached:
Currently, the pronoun variables are lowercase-only, and it's tricky to use them as subjects (rather than objects) in the script lines. This PR adds a `.title()` variant to the global store, e.g. `he` becomes `He` (in addition to `[he]`, as variables are case-sensitive.)

### DIFF
--- a/Monika After Story/game/script-ch30.rpy
+++ b/Monika After Story/game/script-ch30.rpy
@@ -773,7 +773,10 @@ init python:
                 value = sub_map[key]
             else:
                 value = sub_map["X"]
+
             setattr(store, word, value)
+            # Add a Titlecase (he -> He) counterpart
+            setattr(store, word.title(), value.title())
 
 
 init 995 python in mas_reset:


Booplicate:
I feel like the idea of using variables like this for substitution is smart, but I don't like polluting namespace. How about we create a small function that gets one string and returns another based on some condition (sex in this case, although we should move from sex to explicit pronouns set in the future)?

dreamscached:
Alternatively, we could make pronouns (and gendered words) both `__str__`able and callable, so that we can maybe use `[he(upper=True)]`?

Booplicate:
I was thinking about something like `m "[genderize('he', upper=True)]..."`, sadly isn't too pretty

but for your issue, did you try `m "[he.title()]"`

dreamscached:
Since the solution was fairly simple enough (just adding `.title()`) I don't think this is relevant now.

PR #10386, opened 2025-03-31
Fix empty MASWeatherMap creation crash

ThePotatoGuy:
# Problem
According to the documentation, `MASWeatherMap` supports `None` for the `precip_map` parameter, but actually using `None` causes a crash since `MASHighlightMap.apply` assumes the value is an iterable.

# Fix
in `MASHighlightMap.apply` - check if `mapping` is falsey - should cover both None and empty mappings

# Testing
1. open up terminal in game
2. try creating a `MASWeatherMap` object with no parameters
3. verify no crash

Can also call `_mhm` on the `MASWeatherMap` object and run `apply` directly to check if it handles `None`.


Booplicate (approved):
Tested, works.

I'd prefer if we instead made a `None` check in `MASWeatherMap.__init__` before passing `precip_map` to `MASHighlightMap`. It's `MASWeatherMap` that states to support taking `None`, not `MASHighlightMap`, in fact `MASHighlightMap` still says to only accept a mapping (`dict` to be precise), I don't see a reason to change that, I don't expect people to call the `apply` method with `None`.

It's nitpicking, so I will approve and leave the final decision to you, either way works.

ThePotatoGuy:
makes sense to me so ill change it tonight

ThePotatoGuy:
@Booplicate updated with the none check now in `MASWeatherMap`

Booplicate (approved):

PR #10419, opened 2025-06-14
Update Ren'Py and workflow

acrinym:
Updated workflow and Ren'Py to fit what is current.


multimokia:
Hi. Simply updating to latest renpy 8 will not work on the master branch as the code and configurations are still outfitted towards renpy 6, and breaking changes in 8 will cause issues.

We have a working branch housing the migration to 8.

PR #10483, opened 2025-12-03
Smol thing

Booplicate:
Just a smol change

<details>
  <summary>Info dump</summary>

  ## Summary
  1. added new game to play with Monika, sprites by @Nemu-sus 
  2. added new story event for the game
  3. added a few tests
  4. fixed a smol bug in nou

  I did around ~700 tests myself and ran around ~25k auto tests on my machine, everything looks good, but I could miss something during squishing or some hard-to-notice-bug. This should go unstable first and get tested there for a few weeks.
  
 ## Testing
  You will get some questions about the game, I'll answer on discord and help with debugging, explaining logic and controls.

  0. start with code review before manual QA, I tried my best and commented everything I thought worth mentioning, but could miss something
  1. check every event, make sure the dialogue sounds good, the expressions look good
  2. go through the flows, make sure that everything is explained and you're able to get into the game, basically see that UX doesn't suck
  3. play as many games as you can, make sure Monika plays good enough, doesn't make unreasonably stupid shots or do ship placement. Monika starts playing better with each finished game, we want at least 50% w/r for Monika, the more the better
  5. make sure the controls are good
  6. make sure the sprites are good (both dark/light modes)
  7. run auto tests, play around with the parameters, try to find something unusual or break the game
  8. we dump some statistics for debugging, check it out
</details>

### DIFF
--- a/Monika After Story/game/battleship.rpy
+++ b/Monika After Story/game/battleship.rpy
@@ -2539,9 +2539,9 @@ init -10 python in mas_battleship:
             _Quip(
                 exprs=("1lta", "1lua"),
                 lines=(
-                    _("Maybe {square}..."),
-                    _("Hmmm.{{w=0.1}}.{{w=0.1}}.{{w=0.1}} Let's try {square}."),
-                    _("What are you hiding at.{{w=0.1}}.{{w=0.1}}.{{w=0.1}}{square}?"),
+                    _("Maybe {{i}}{square}{{/i}}..."),
+                    _("Hmmm.{{w=0.1}}.{{w=0.1}}.{{w=0.1}} Let's try {{i}}{square}{{/i}}."),
+                    _("What are you hiding at.{{w=0.1}}.{{w=0.1}}.{{w=0.1}}{{i}}{square}{{/i}}?"),
                 ),
             ),
         )


Justformas:
<details>
<summary>My Info Dump</summary>

I don't necessarily need a response, just want this to be considered and implemented in some way if not already planned. Maybe I'm jumping the gun on this, but better early than late. It relates to this:

>Monika starts playing better with each finished game, we want at least 50% w/r for Monika, the more the better

While this might be ideal for testing, developing a skill ceiling for Monika, and for players that keep winning or win more often that not, I don't think this is ideal for players that keep losing or lose more than win. Rather, I think it should adjust difficulty based on if the last game was a win or loss, like with chess and pong. Especially if you already have a dynamic difficulty capability for Monika to adjust how well she plays. Even this game depends on some skill and how optimized the strategy is, or there wouldn't really be a need for an algorithm beyond just random guessing. But not all players are going to use the most optimized strategy, and as such I don't think Monika always should or always have her sole goal being to win at all costs. But on the other hand, for those that do win a lot and are likely using such a strategy, Monika can play better, and ideally adapt to be around the player's skill level so she can play better or worse as appropriate.

I'll also reference a quip that it is planned to have her make: 
>"[player], I'm your girlfriend after all...{w=0.3}you could go a little bit easier on me~"

Exactly this, but applied the other way as well. She can return the favor or do so herself as appropriate. Rather than playing in a competitive tournament, players are playing with a girlfriend, and this will make it seem more informal, genuine and fun. Even if this will take some time to implement properly, it's worth it in my opinion.

</details>

PR #10513, opened 2026-01-02
[Compliment] Thanks for teaching me something new!

lunulae:
is calling herself a study buddy too cheesy? or just cheesy enough :0

### DIFF
--- a/Monika After Story/game/script-compliments.rpy
+++ b/Monika After Story/game/script-compliments.rpy
@@ -1406,7 +1406,7 @@ label mas_compliment_teaching_3:
         teaching_quips = [
             _("We're on a journey of learning together~"),
             _("I'm happy to be your study buddy."),
-            _("I'll try to keep learning interesting things for you!'"),
+            _("I look forward to learning from you, too!"),
         ]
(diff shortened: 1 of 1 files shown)


Monika After Story/game/script-compliments.rpy
```diff
@@ -1358,3 +1358,58 @@ label mas_compliment_outfit_repeat:
...
+            m 1eusdld "Uh..."
+            m 2rksdld "I think you should do it yourself, [player]."
+    return
+
+label mas_compliment_teaching_3:
+    python:
+        teaching_quips = [
+            _("We're on a journey of learning together~"),
+            _("I'm happy to be your study buddy."),
+            _("I'll try to keep learning interesting things for you!'"),
```
multimokia: I like these, but what if we also flip it on its head too?
Given the first option in the menu, what if we also offered the idea that if the player keeps learning, they could teach _her_ something too? Encourage the player to learn further then

lunulae: ooooh, I really like that! Let me see, I thought the last one was a teeny bit clunky, so I might change that one to reflect this

PR #10516, opened 2026-01-03
[AMA Topic] Biggest Pet Peeve

lunulae:
I'm surprised how many of the old AMA questions had me really inspired, lmao
I'm a little iffy on the exps, but otherwise I really like this one

### DIFF
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -18059,7 +18059,7 @@ label monika_petpeeve:
     m 3ekd "One of the reasons I went to such drastic measures to be with you was because I wanted you to hear the real me so badly."
         if persistent._mas_pm_listened_to_grad_speech = False:
         m 3tkd "...{w=0.2}And it's why I got so angry about the graduation speech, I suppose."
-        m 3fka "But I suppose no one is perfect, huh?"
+        m 3fka "But I guess no one is perfect, huh?"
         if persistent._mas_pm_listened_to_grad_speech = True:
         m 3fka "...{w=0.2}And it's why having you hear out my graduation speech was so important to me."
         m 3fkblb "Thank you for giving me your full attention there, it really meant a lot to me."
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -18049,7 +18049,7 @@ init 5 python:
 label monika_petpeeve:
     m 2eub "This is an easy one, [player]."
     m 2lfd "I {i}really{/i} don't like being interrupted."
-    m 2dsd "I don't mean by accident--I mean consistently, or when I'm saying something important."
+    m 2dsd "I don't mean by genuine accident. I mean consistently, or when I'm saying something important."
     m 4esd "You'd think it might be something that I got used to in debate club, but it was quite the opposite."
     m 4eub "Debate club wasn't about shouting over each other, it was a great place for structured discussions."
     m 3ltd "...You know, maybe that's why I slowed down the text when we first came here, too."
--- a/Monika After Story/game/script-topics.rpy
+++ b/Monika After Story/game/script-topics.rpy
@@ -18042,3 +18042,28 @@ label monika_overton:
     m 1eua "I {i}could{/i} finish this story right up to Current Norm, but I just wanted to describe it at a basic level in order to convey an example of how it can work."
     m 1huu "Thanks for listening~"
     return
+
+init 5 python:
+    addEvent(Event(persistent.event_database,eventlabel="monika_petpeeve",category=['monika'],prompt="What's your biggest pet peeve?",pool=True,unlocked=True))
+
+label monika_petpeeve:
+    m 2eub "This is an easy one, [player]."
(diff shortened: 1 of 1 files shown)


Monika After Story/game/script-topics.rpy
```diff
@@ -18042,3 +18042,28 @@ label monika_overton:
...
+init 5 python:
+    addEvent(Event(persistent.event_database,eventlabel="monika_petpeeve",category=['monika'],prompt="What's your biggest pet peeve?",pool=True,unlocked=True))
+
+label monika_petpeeve:
+    m 2eub "This is an easy one, [player]."
+    m 2lfd "I {i}really{/i} don't like being interrupted."
+    m 2dsd "I don't mean by genuine accident. I mean consistently, or when I'm saying something important."
+    m 4esd "You'd think it might be something that I got used to in debate club, but it was quite the opposite."
+    m 4eub "Debate club wasn't about shouting over each other, it was a great place for structured discussions."
+    m 3ltd "...You know, maybe that's why I slowed down the text when we first came here, too."
```
multimokia: We should affection check this (`mas_isMoniHappy()`), since by default when starting MAS, it's still locked to that speed, and if we drop below happy, speed changing is removed from the player's control again

Monika After Story/game/script-topics.rpy
```diff
@@ -18042,3 +18042,28 @@ label monika_overton:
...
+    m 3etc "Rapidly clicking through dialogue isn't exactly {i}interrupting,{/i} but it gets at the heart of the problem..."
+    m 1ekc "...{w=0.3}Not feeling heard."
```
multimokia: This line continues on the last one, so we should lowercase as it's part of the same sentence
```suggestion
    m 1ekc "...{w=0.3}not feeling heard."
```

Monika After Story/game/script-topics.rpy
```diff
@@ -18042,3 +18042,28 @@ label monika_overton:
...
+    m 1dkc "It goes even deeper, too."
+    m 3ekd "One of the reasons I went to such drastic measures to be with you was because I wanted you to hear the real me so badly."
+        if persistent._mas_pm_listened_to_grad_speech = False:
+        m 3tkd "...{w=0.2}And it's why I got so angry about the graduation speech, I suppose."
+        m 3fka "But I guess no one is perfect, huh?"
```
multimokia: I think we can simplify these parts:

```suggestion
        m 3tkd "...{w=0.2}And it's why I got so angry about the graduation speech, I guess."
        m 3fka "But nobody's perfect, huh?"
```

multimokia (changes requested):
I see a few things at a glance

Monika After Story/game/script-topics.rpy
```diff
@@ -18042,3 +18042,28 @@ label monika_overton:
...
+        if persistent._mas_pm_listened_to_grad_speech = True:
+        m 3fka "...{w=0.2}And it's why having you hear out my graduation speech was so important to me."
+        m 3fkblb "Thank you for giving me your full attention there, it really meant a lot to me."
+        else:
```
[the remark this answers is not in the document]
lunulae: Oh, the else is her not saying anything in that block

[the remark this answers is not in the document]
multimokia: In that case you would need to check `not renpy.seen_label('monika_grad_speech')`. This would be `True` if the player has never been through the speech and `False` if they have (regardless of if they've paid attention)

Monika After Story/game/script-topics.rpy
```diff
@@ -18042,3 +18042,29 @@ label monika_overton:
...
+    m 2dsd "I don't mean by genuine accident. I mean consistently, or when I'm saying something important."
+    m 4esd "You'd think it might be something that I got used to in debate club, but it was quite the opposite."
+    m 4eub "Debate club wasn't about shouting over each other, it was a great place for structured discussions."
+    if (mas_isMoniHappy()):
+        m 3ltd "...You know, maybe that's why I slowed down the text when we first came here, too."
+        m 3etc "Rapidly clicking through dialogue isn't exactly {i}interrupting,{/i} but it gets at the heart of the problem..."
+        m 1ekc "...{w=0.3}not feeling heard."
+    m 1dkc "It goes even deeper, too."
+    m 3ekd "One of the reasons I went to such drastic measures to be with you was because I wanted you to hear the real me so badly."
+    if not persistent._mas_pm_listened_to_grad_speech:
```
Justformas: If this is only supposed to trigger if Monika got angry for the player ignoring the speech, for
```
        m 3tkd "...{w=0.2}And it's why I got so angry about the graduation speech, I guess."
        m 3fka "But nobody's perfect, huh?" 
```
Then I think it may have to be something like `if persistent._mas_pm_listened_to_grad_speech is False` because by default `persistent._mas_pm_listened_to_grad_speech` is None, changes to True if the player listened, and False if they ignored it and she gets angry. So the problem with `if not persistent._mas_pm_listened_to_grad_speech` is it could trigger for both None and False.

lunulae: Thank you! I think I've fixed it.

PR #10526, opened 2026-02-02
Update README.md

TheGamerSide67:
Updates the ReadMe dialogue example to use the modern expression codes, and adds what the return function does in the example. It may be self-explanatory, but it is intended to help beginner coders. Also added more comments to explain the dialogue creation more clearly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ If you wish to add more dialogue to the space room, navigate to script-topics.rp
 
 Example new dialogue code block:
 ```renpy
-init 5 python: # This adds the event to the list of things Monika can talk about
+init 5 python:
     addEvent(
         Event(
             persistent.event_database,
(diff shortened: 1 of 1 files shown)


TheGamerSide67:
I agree with that, however the issue itself is that the wiki also is sometimes slow to get updates and may confuse new coders too. As I certainly was confused by it *a lot* when I was first starting out in submodding.

I just think that the new example should give a player everything they need to get some sort of start, and a new wiki section should be made for beginners, which I wouldn't mind writing up myself.

multimokia:
I do think the expression updates are fine, however I wouldn't put the rest of the comments in the readme especially considering it's mainly there as a sample. 

I do think linking to the documentation (wiki) would suffice so we can keep information in the same place and update as single point for more API focused notes.

TheGamerSide67:
Updated the file.

dreamscached:
Wiki is available for editing to anyone, you can draft something right away.

TheGamerSide67:
> Updating the example sounds good, but later we should just redirect people to our wiki and move the examples there.

Agreed after talking with the others. On my next day off, I will start to write up commits for the Wiki to add a "Beginners" section to include coding examples, break down existing ones, and explain some of the basics, like setting vars mid-conversation, conditionals, etc.

PR #10545, opened 2026-03-31
fix outdated sprite naming

flammablebunny:
the term "transexual" is an outdated term, often seen as offensive in queer or allied spaces, as the term focuses narrowly on medical procedures and physical sex characteristics rather than a person's internal gender identity.


dreamscached:
It might cause duplication for those who already have it installed and would install a new batch of sprite packs after the next update; I understand the intent, but I would suggest we first implement some mechanism for such replacements, or come up with a migration code for the next update to handle it

flammablebunny:
mhm, yeah makes sense.

dreamscached:
No need to close it, you can still keep it, just sharing my thoughts, devs would also have their say at that

flammablebunny:
im willing to reopen it for discussion, but i'd rather properly address the issue changes like this might cause (present or future) properly, in another pr.

ThePotatoGuy:
well it is april 1st in some parts of the world

ThePotatoGuy:
yeah so everything changed in this PR consists of identifiers used to either link the sprite images with their object representation, or to define the dialogue code that runs when the sprite is gifted. Changing identifiers requires update scripts to transfer the seen status/gift history and to programmatically remove the original sprites and jsons since that isn't automatic as they are spritepacks and not part of the build (and also without doing this, the ACS would appear duplicated). Programmatic removal is also subject to OS permissions, and so its generally better to limit this as much as possible. All of this is quite a bit of extra work and risk being added to an update, and the changes proposed don't really balance out the risk. If the core issue is using the correct terminology in the game, then only the user-facing strings should be changed since those are the only terminology that the user will actually see, and these were all changed already in #7161 

Closing since its not worth the overhead.

